PR #2, opened 2020-03-15
Ori/swagger with generation

ori-amizur:


filanov:
usually we split the auto generated commits from actual changes. it's hard to do a review when you have 100 files that most of them are auto generated.

regardless you changed the operation id so the interface have changed as well, did you checked that make all works?

ori-amizur:
It is not working for me yet.  Still in progress.
Anyway, I can separate commits.  My only changes were in the swagger.

filanov:
but swagger change the interface that the service need to implement.

PR #4, opened 2020-03-16
Ori/connectivity swagger changes

ori-amizur:


swagger.yaml
```diff
@@ -405,27 +409,40 @@ definitions:
...
       step-type:
         $ref: '#/definitions/step-type'
-      return-code:
-        type: integer
-      data:
+      succcessful-completion:
+        type: boolean
+      output:
+        type: string
+      error:
```
filanov: Why do you have error if you have output?
if the operation fails i expect to see the error in the output no?

swagger.yaml
```diff
@@ -405,27 +409,40 @@ definitions:
     properties:
       step-type:
         $ref: '#/definitions/step-type'
-      return-code:
-        type: integer
-      data:
+      succcessful-completion:
```
filanov: maybe it should be an error code? 0 for success anything else error?

ori-amizur: It can.  But I will probably have to invent errors.  So these definitions need to be somewhere (swagger)?

swagger.yaml
ori-amizur: I distinguished the output of the command and the error I got from golang. I can unify them if you think it is better.

swagger.yaml
filanov: not if we are working like bash output.

swagger.yaml
```diff
@@ -192,18 +193,21 @@ paths:
...
-      responses:
-        200:
-          description: Steps reply
+        - name: reply
+          in: body
           schema:
-            $ref: '#/definitions/steps-reply'
+            $ref: '#/definitions/step-reply'
+      responses:
+        201:
```
avishayt: We don't need to return anything, so return 204

swagger.yaml
avishayt: Agree, let's leave it as integer and call it "exit code" - this seems to be the standard bash term

swagger.yaml
avishayt: Not sure.  Maybe Ori is correct and we should have stderr and stdout?

swagger.yaml
```diff
@@ -405,27 +409,40 @@ definitions:
...
+    properties:
+      node-id:
+        type: string
+        format: uuid
+      nics:
+        type: array
+        items:
+          $ref: '#/definitions/connectivity-check-nic'
+
   connectivity-check-params:
```
avishayt: Not used in the swagger?

swagger.yaml
```diff
@@ -483,3 +500,153 @@ definitions:
...
+      cpus:
+        type: integer
+      threads-per-core:
+        type: integer
+      sockets:
+        type: integer
+      cpu-mhz:
+        type: number
+
+  block-device:
```
avishayt: I would add some more information - model, capacity, SSD/HDD, rotation rate if HDD

swagger.yaml
```diff
@@ -405,27 +409,40 @@ definitions:
...
-        type: integer
-      data:
+      succcessful-completion:
+        type: boolean
+      output:
+        type: string
+      error:
         type: string
 
+  connectivity-check-nic:
```
avishayt: What is this used for?

swagger.yaml
```diff
@@ -483,3 +500,153 @@ definitions:
...
+        type: string
+      model-name:
+        type: string
+      cpus:
+        type: integer
+      threads-per-core:
+        type: integer
+      sockets:
+        type: integer
+      cpu-mhz:
```
avishayt: More user-friendly to use GHz with float?

swagger.yaml
```diff
@@ -483,3 +500,153 @@ definitions:
...
+
+  cidr:
+    type: object
+    properties:
+      ip-address:
+        type: string
+      mask:
+        type: integer
+
+  nic:
```
avishayt: Add vendor/model, bandwidth

swagger.yaml
```diff
@@ -483,3 +500,153 @@ definitions:
...
+      outgoing-ip-address:
+        type: string
+      remote-ip-address:
+        type: string
+      remote-mac:
+        type: string
+      successful:
+        type: boolean
+
+
```
avishayt: Remove enters

swagger.yaml
ori-amizur: Currently lsblk is used so we can use the available columns there. Some of the columns  are available:capacity == size, device-type.  The model can be added.
We are going to talk with ironic guys, so maybe we will use the ironic inspector implementation.

swagger.yaml
ori-amizur: OK

swagger.yaml
avishayt: Please find a way to make it clearer where things are used

swagger.yaml
avishayt: OK, I guess the top priority right now is getting to a demo, so use what's available now and we'll refine it later.

PR #11, opened 2020-03-23
MGMT-543 Nodes discovery framework [WIP]

oshercc:


internal/bminventory/inventory.go
```diff
@@ -367,7 +367,17 @@ func (b *bareMetalInventory) ListHosts(ctx context.Context, params inventory.Lis
 
 func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params inventory.GetNextStepsParams) middleware.Responder {
 	steps := models.Steps{}
+	var host models.Host
+
 	b.debugCmdMux.Lock()
+
+	if err := b.db.First(&host, "id = ?", params.HostID).Error; err != nil {
```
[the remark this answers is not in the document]
oshercc: So to make changes here or not ?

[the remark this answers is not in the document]
filanov: you probably should it will teach you something new :)

internal/bminventory/inventory.go
```diff
@@ -367,7 +367,17 @@ func (b *bareMetalInventory) ListHosts(ctx context.Context, params inventory.Lis
...
+		return inventory.NewGetNextStepsNotFound()
+	}
+	if *host.Status == "disabled"{
```
filanov: probably cleaner to use `swag.StringValue`

internal/bminventory/inventory.go
```diff
@@ -369,15 +375,14 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params inventory.
 	steps := models.Steps{}
 	var host models.Host
 
-	b.debugCmdMux.Lock()
-
 	if err := b.db.First(&host, "id = ?", params.HostID).Error; err != nil {
 		return inventory.NewGetNextStepsNotFound()
 	}
-	if *host.Status == "disabled"{
+	if host.Status == swag.String(HostStatusDisabled){
```
filanov: `host.Status` is the one that can be null so it safer to use the `swag.StringValue` on the pointer and not the other way around.

filanov: relevant comment for all the functions.

internal/bminventory/inventory.go
```diff
@@ -375,6 +381,17 @@ func (b *bareMetalInventory) ListHosts(ctx context.Context, params inventory.Lis
...
 func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params inventory.GetNextStepsParams) middleware.Responder {
 	steps := models.Steps{}
+	var host models.Host
+
+	if err := b.db.First(&host, "id = ?", params.HostID).Error; err != nil {
+		logrus.WithError(err).Error("failed to find host")
+		return inventory.NewGetNextStepsNotFound()
+	}
+
+	if *host.Status == *swag.String(HostStatusDisabled){
```
filanov: what is this? 
```suggestion
	if swag.StringValue(host.Status) == HostStatusDisabled {
```

filanov: same comment for other places as well

subsystem/host_test.go
```diff
@@ -58,6 +58,31 @@ var _ = Describe("Host tests", func() {
...
+		})
+		Expect(err).NotTo(HaveOccurred())
+		_, err = bmclient.Inventory.DisableHost(ctx, &inventory.DisableHostParams{HostID: *host.GetPayload().ID})
+		reply, err := bmclient.Inventory.GetHost(ctx, &inventory.GetHostParams{HostID: *host.GetPayload().ID})
+		Expect(err).NotTo(HaveOccurred())
+		replyHost := reply.GetPayload()
+		Expect(*replyHost.Status).Should(Equal("disabled"))
+
+		nsteps, err := bmclient.Inventory.GetNextSteps(ctx, &inventory.GetNextStepsParams{HostID: *host.GetPayload().ID})
+		_, ok := getStepInList(nsteps.GetPayload(), models.StepTypeHardawareInfo)
```
filanov: check the list length. don't look for a specific step.

internal/bminventory/inventory.go
```diff
@@ -404,9 +421,36 @@ func (b *bareMetalInventory) SetDebugStep(ctx context.Context, params inventory.
...
+
+	if err := b.db.First(&host, "id = ?", params.HostID).Error; err != nil {
+		return inventory.NewEnableHostNotFound()
+	}
+
+	if *host.Status != *swag.String(HostStatusDisabled){
+		return inventory.NewEnableHostConflict()
+	}
+
+	if err := b.db.Model(&host).Where("host_id = ?", params.HostID.String()).Update("status", *swag.String(HostStatusDiscovering)).Error; err != nil {
```
filanov: i think that the HW info should be cleared as well because the node will be in discovery state it will need to send the info again. 
if you not reset the HW info then the host monitor will think that everything is ok.

internal/bminventory/inventory.go
```diff
@@ -404,9 +421,36 @@ func (b *bareMetalInventory) SetDebugStep(ctx context.Context, params inventory.
...
+		return inventory.NewEnableHostNotFound()
```
filanov: it's not possible that the node is not found. you already checked it. it probably an internal error.

subsystem/host_test.go
```diff
@@ -58,6 +58,31 @@ var _ = Describe("Host tests", func() {
...
+		Expect(ok).Should(Equal(false))
+
+		_, err = bmclient.Inventory.EnableHost(ctx, &inventory.EnableHostParams{HostID: *host.GetPayload().ID})
+		reply, err = bmclient.Inventory.GetHost(ctx, &inventory.GetHostParams{HostID: *host.GetPayload().ID})
+		Expect(err).NotTo(HaveOccurred())
+		replyHost = reply.GetPayload()
+		Expect(*replyHost.Status).Should(Equal("discovering"))
```
filanov: get next step again to make sure it contain HW info requests

filanov:
general comment about the commits, usually we split between auto generated code commits and actual changes, sometimes the amount of generated files can be huge and it's hard to do the review.

PR #13, opened 2020-03-24
Michaelf/cluster

filanov:


filanov:
@avishayt ignore the implementation just take a look on the swagger commit.

PR #15, opened 2020-03-26
1. Add support for execute step-type

ori-amizur:
2. Add command and arguments for execute
3. Add step-id
Note that in the future, step-type will be removed, and the agent will only execute commands.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -469,16 +469,21 @@ definitions:
     enum:
       - hardaware-info
       - connectivity-check
-      - debug
+      - execute
 
   step:
     type: object
     properties:
       step-type:
         $ref: '#/definitions/step-type'
-      data:
+      step-id:
         type: string
-        format: json
+      command:
+        type: string
+      args:
+        type: array
+        items:
+          type: string
 
   steps-reply:
     type: array
@@ -488,8 +493,8 @@ definitions:
   step-reply:
     type: object
     properties:
-      step-type:
-        $ref: '#/definitions/step-type'
+      step-id:
+        type: string
       exit-code:
         type: integer
       output:


avishayt:
Swagger looks OK

swagger.yaml
```diff
@@ -469,16 +469,22 @@ definitions:
...
-      - debug
+      - execute
 
   step:
     type: object
     properties:
       step-type:
         $ref: '#/definitions/step-type'
-      data:
+      step-id:
```
ronniel1: Aren't these fields relevant only for the 'execute' command type?

swagger.yaml
```diff
@@ -469,16 +469,22 @@ definitions:
...
         type: string
-        format: json
+        format: uuid
```
ronniel1: I think step-id should be a string and not a uuid
possible values will be (inventory, connectivity)

ori-amizur: This is the step-type which will be removed (eventually).
Since the agent is not aware of what it is executing, it has unique id for the step.
I think we would like to differentiate between steps, even if the same command was executed.

ronniel1: But this forces the server to keep track of these.
Otherwise it will not know that results with UUID1 is the result of the connectivity check.

@filanov what do you think?

swagger.yaml
ori-amizur: Eventually all will be execute.
The parameters for command and arguments should not change when done through execute

ronniel1: that id right, but step-type of hardware-info will not have these fields, will it?

swagger.yaml
ori-amizur: We might want to keep history of results for debug purposes - then you need to distinguish between them.
Maybe a node was added to the connectivity check request.  So, the request and reply are coupled.

swagger.yaml
ori-amizur: Every command has these fields, but hardware-info ignores them.

swagger.yaml
filanov: i tend to agree with @ronniel1 it will be easy to see the steps in the logs.
anyway we can combine: `connectivity-<UUID>` is that ok?

ori-amizur: Yes

PR #16, opened 2020-03-29
Pass public ssh key to ignition file

oshercc:


internal/bminventory/inventory.go
```diff
@@ -162,7 +163,7 @@ func (b *bareMetalInventory) createImageJob(ctx context.Context, id string) erro
 								},
 								{
 									Name:  "IGNITION_CONFIG",
-									Value: fmt.Sprintf(ignitionConfigFormat, b.InventoryURL, b.InventoryPort),
+									Value: fmt.Sprintf(ignitionConfigFormat, b.GetUserSshKey(id), b.InventoryURL, b.InventoryPort),
```
filanov: please make it a function.

internal/bminventory/inventory.go
```diff
@@ -185,6 +186,21 @@ func (b *bareMetalInventory) createImageJob(ctx context.Context, id string) erro
 	return nil
 }
 
+func (b *bareMetalInventory) GetUserSshKey(id string)  string{
```
filanov: i think that the parameter name should be `clusterId` it's not clear what is the id from the name.

deploy/bm-inventory.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
         - name: bm-inventory
-          image: REPLACE_IMAGE
+          image: quay.io/oscohen/bm-inventory:latest
```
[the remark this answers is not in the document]
filanov: there i a script that replace it. don't do it manually

internal/bminventory/inventory.go
```diff
@@ -185,6 +186,25 @@ func (b *bareMetalInventory) createImageJob(ctx context.Context, id string) erro
 	return nil
 }
 
+func (b *bareMetalInventory) FormatIgnitionFile(ClusterId string)  string{
+	return fmt.Sprintf(ignitionConfigFormat, b.GetUserSshKey(ClusterId), b.InventoryURL, b.InventoryPort)
+}
+
+func (b *bareMetalInventory) GetUserSshKey(ClusterId string)  string{
+	var cluster models.Cluster
+	b.db.First(&cluster, "id = ?", ClusterId)
```
filanov: where do you check for errors?

internal/bminventory/inventory.go
```diff
@@ -185,6 +187,28 @@ func (b *bareMetalInventory) createImageJob(ctx context.Context, id string) erro
...
 }
 
+func (b *bareMetalInventory) FormatIgnitionFile(ClusterId string)  string{
+	return fmt.Sprintf(ignitionConfigFormat, b.GetUserSshKey(ClusterId), b.InventoryURL, b.InventoryPort)
+}
+
+func (b *bareMetalInventory) GetUserSshKey(clusterId string)  string{
+	var cluster models.Cluster
+	if err := b.db.First(&cluster, "id = ?", clusterId); err != nil {
+		errors.Wrap(err, "cluster not found")
```
filanov: you are not doing anything with the error
the ignition is wrong but you continue anyway?

internal/bminventory/inventory.go
```diff
@@ -185,6 +187,28 @@ func (b *bareMetalInventory) createImageJob(ctx context.Context, id string) erro
 	return nil
 }
 
+func (b *bareMetalInventory) FormatIgnitionFile(ClusterId string) string {
```
filanov: change the function to return string and error
then use this function above the creation of the job and check if it failed. if it is then return an error

internal/bminventory/inventory.go
```diff
@@ -185,6 +187,28 @@ func (b *bareMetalInventory) createImageJob(ctx context.Context, id string) erro
...
+	var cluster models.Cluster
+	if err := b.db.First(&cluster, "id = ?", clusterId); err != nil {
+		return errors.Wrap(err, "cluster not found")
+	}
+
+	if cluster.SSHPublicKey == "" {
+		return ""
+	}
+
+	User := fmt.Sprintf(`,{
```
filanov:
```suggestion
	return fmt.Sprintf(`,{
```

filanov:
i think it will be a better option if you get the cluster in the beginning of DownloadClusterISO it's missing the validation there anyway.
then pass the cluster to createImageJob then you don't have to validate any get there. and you will use this cluster struct anyway for other parameters in the future

internal/bminventory/inventory.go
```diff
@@ -185,6 +187,21 @@ func (b *bareMetalInventory) createImageJob(ctx context.Context, id string) erro
 	return nil
 }
 
+func (b *bareMetalInventory) FormatIgnitionFile(cluster models.Cluster) string {
```
filanov: user a pointer.

filanov: make it a private function `formatIgnitionFile`

internal/bminventory/inventory.go
```diff
@@ -185,6 +187,21 @@ func (b *bareMetalInventory) createImageJob(ctx context.Context, id string) erro
...
+	return fmt.Sprintf(ignitionConfigFormat, b.GetUserSshKey(cluster), b.InventoryURL, b.InventoryPort)
+}
+
+func (b *bareMetalInventory) GetUserSshKey(cluster models.Cluster) string {
```
filanov: make it a private function `getUserSshKey`

internal/bminventory/inventory.go
oshercc: user vat is removed

internal/bminventory/inventory.go
```diff
@@ -249,7 +266,12 @@ func (b *bareMetalInventory) DeregisterCluster(ctx context.Context, params inven
 }
 
 func (b *bareMetalInventory) DownloadClusterISO(ctx context.Context, params inventory.DownloadClusterISOParams) middleware.Responder {
-	if err := b.createImageJob(ctx, params.ClusterID); err != nil {
+	var cluster *models.Cluster
```
filanov: here it was find. you have an object and you use its pointer. 
```suggestion
	var cluster models.Cluster
	if err := b.db.First(&cluster, "id = ?", params.ClusterID); err != nil {
```

internal/bminventory/inventory.go
```diff
@@ -249,7 +266,12 @@ func (b *bareMetalInventory) DeregisterCluster(ctx context.Context, params inven
...
+	if err := b.db.First(cluster, "id = ?", params.ClusterID); err != nil {
+		return inventory.NewDownloadClusterISONotFound()
+	}
+
+	if err := b.createImageJob(ctx, cluster); err != nil {
```
filanov: and then here 
```suggestion
	if err := b.createImageJob(ctx, &cluster); err != nil {
```

PR #21, opened 2020-03-31
swagger: add kubeconfig download

avishayt:


filanov:
@eranco74 can you please remind me where does this file used?

swagger.yaml
```diff
@@ -174,6 +174,31 @@ paths:
...
+      produces:
+        - text/x-yaml
+      parameters:
+        - in: path
+          name: clusterId
+          description: The ID of the cluster whose kubeconfig to download
+          type: string
+          format: uuid
+          required: true
+      responses:
```
ronniel1: What will be the response if the kubeconfig file was not created yet?

filanov: i think that it's a synchronous API.

avishayt: If no kubeconfig it should be 409

PR #22, opened 2020-03-31
Avishay/swagger add params

avishayt:


swagger.yaml
```diff
@@ -612,20 +639,36 @@ definitions:
...
+        description: OpenShift cluster name
       openshiftVersion:
         type: string
         pattern: '^4\.\d$'
+        description: OpenShift cluster version
       baseDnsDomain:
         type: string
+      clusterNetworkCIDR:
+        type: string
+        description: IP address block from which Pod IPs are allocated
```
eranco74: Consider adding that it should be in CIDR notation and have an exampe: e.g.  10.128.0.0/14

avishayt:
@eranco74 Please review again

swagger.yaml
```diff
@@ -174,6 +174,31 @@ paths:
...
           description: Internal server error
 
+  /clusters/{clusterId}/downloads/kubeconfig:
+    get:
+      tags:
+        - inventory
+      summary: Download the kubeconfig file for the specified cluster
+      operationId: DownloadClusterKubeconfig
+      produces:
+        - text/x-yaml
```
filanov, line 184: i hope it supported - not all the formats supported by go - i'll let you know when i try to implement it

avishayt: Yep no problem

PR #23, opened 2020-04-01
State machine updates

avishayt:
- Removed "insufficient" state. This is now represented by a host
  property called "minimumRequirements". If minimum requirements are not
  met for the node's current role, minimumRequirements will be set to
  false, and the reason will be in statusInfo.
- Removed "disabled" state. This is now a host property called
  "enabled".
- Removed the transition from "known" to "discovering". The service
  should periodically refresh the info without changing state. If hosts
  are always transitioning between "discovering" and "known", it will
  be hard for the user to initiate installation (all hosts must be in
  "known" state).


filanov:
@avishayt you need to change the swagger enum as well no?
the enabled/disabled status will break current logic, i'm not sure we need to change it. because it will be clear to see in the ui that the node is in disabled state no? maybe we can call it maintenance or something like that?

avishayt:
> @avishayt you need to change the swagger enum as well no?

I alreayd removed the states from the enum, unless I didn't understand what you meant?

> the enabled/disabled status will break current logic, i'm not sure we need to change it. because it will be clear to see in the ui that the node is in disabled state no? maybe we can call it maintenance or something like that?

Adding it back.

ronniel1:
Why did the disabled state become a property?
It does affect what we do with the host

PR #27, opened 2020-04-02
Run introspector in the host

ori-amizur:


internal/bminventory/inventory.go
```diff
@@ -70,7 +70,7 @@ const ignitionConfigFormat = `{
 "units": [{
 "name": "introspector.service",
 "enabled": true,
-"contents": "[Service]\nType=oneshot\nExecStart=docker run --rm --privileged --net=host quay.io/oamizur/introspector:latest /usr/bin/introspector --host %s --port %s --cluster-id %s\n\n[Install]\nWantedBy=multi-user.target"
+"contents": "[Service]\nType=simple\nExecStartPre=docker run --privileged --rm -v /usr/local/bin:/hostbin quay.io/oamizur/introspector:latest cp /usr/bin/introspector /usr/sbin/dmidecode /hostbin\nExecStart=/usr/local/bin/introspector --host %s --port %s --cluster-id %s\n\n[Install]\nWantedBy=multi-user.target"
```
filanov, line 73: you don't actually need `--privileged` right?

ori-amizur: Yes.  Seems do.

filanov: then please remove

ori-amizur: @filanov Doesn't work without --privileged

filanov: why not?

PR #31, opened 2020-04-05
Liat/mgmt 597 - get next step operation logic

gamli75:


internal/bminventory/inventory.go
```diff
@@ -31,10 +31,13 @@ const (
 )
 
 const (
-	HostStatusDisabled    = "disabled"
-	HostStatusInstalling  = "installing"
-	HostStatusInstalled   = "installed"
-	HostStatusDiscovering = "discovering"
+	HostStatusDisabled     = "disabled"
```
filanov, line 34: you should install file watcher (goimports) to your golang. 
or run `skipper make format` but the file watcher is automatic so it's better.

internal/bminventory/inventory.go
```diff
@@ -547,14 +573,8 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params inventory.
...
-	steps = append(steps, &models.Step{
-		StepType: models.StepTypeHardawareInfo,
-		StepID:   createStepID(models.StepTypeHardawareInfo),
-	})
-	for _, step := range steps {
-		logrus.Infof("Submitting step <%s> to cluster <%s> host <%s> Command: <%s> Arguments: <%+v>", step.StepID, params.ClusterID, params.HostID,
-			step.Command, step.Args)
-	}
+	logrus.Infof("GetNextSteps host: <%s> returns <%d> steps:", params.HostID, len(steps))
+	printSteps(steps)
```
filanov, line 577: the previous log had more information. So i think it's better no? 
```
logrus.Infof("Submitting step <%s> to cluster <%s> host <%s> Command: <%s> Arguments: <%+v>", step.StepID, params.ClusterID, params.HostID,
			step.Command, step.Args)
```

internal/bminventory/inventory.go
```diff
@@ -521,6 +524,12 @@ func createStepID(stepType models.StepType) string {
 	return fmt.Sprintf("%s-%s", stepType, uuid.New().String()[:8])
 }
 
+func printSteps(steps models.Steps)  {
```
filanov, line 527: suggesting to change the name to logSteps (print is confusing)

internal/bminventory/inventory.go
```diff
@@ -531,8 +540,25 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params inventory.
...
 	}
 
-	if swag.StringValue(host.Status) == HostStatusDisabled {
-		return inventory.NewGetNextStepsOK().WithPayload(steps)
+	logrus.Infof("GetNextSteps host: <%s>, host status: <%s>", params.HostID, swag.StringValue(host.Status))
+
+	switch swag.StringValue(host.Status) {
+	case HostStatusKnown, HostStatusInsufficient, HostStatusDisconnected:
+		step := &models.Step{}
+		step.StepType = models.StepTypeConnectivityCheck
```
filanov, line 548: if you have the same logic twice then it should be a function.

subsystem/host_test.go
```diff
@@ -66,6 +66,28 @@ var _ = Describe("Host tests", func() {
 		steps := getNextSteps(clusterID, *host.ID)
 		_, ok := getStepInList(steps, models.StepTypeHardawareInfo)
 		Expect(ok).Should(Equal(true))
+		_, ok = getStepInList(steps, models.StepTypeConnectivityCheck)
```
filanov, line 69: maybe it should be a unit test? @ronniel1 what do you think? one we will have a host monitor this test may become flaky

subsystem/host_test.go
```diff
@@ -198,3 +220,4 @@ func getNextSteps(clusterID, hostID strfmt.UUID) models.Steps {
 	Expect(err).NotTo(HaveOccurred())
 	return steps.GetPayload()
 }
+
```
filanov, line 223: why do you need this line?

internal/bminventory/inventory.go
gamli75: done

internal/bminventory/inventory.go
gamli75: fixed

subsystem/host_test.go
gamli75: it was left from code that I added for testing :) I removed it.

internal/bminventory/inventory.go
gamli75: I changed it back to previous log

internal/bminventory/inventory.go
gamli75: fixed

internal/bminventory/inventory.go
```diff
@@ -521,6 +524,26 @@ func createStepID(stepType models.StepType) string {
...
+	if len(steps) == 0 {
+		logrus.Infof("No steps required for cluster <%s> host <%s>", clusterId, hostId)
+	}
+	for i := 0; i < len(steps); i++ {
+		logrus.Infof("Submitting step <%s> id <%s> to cluster <%s> host <%s> Command: <%s> Arguments: <%+v>", steps[i].StepType, steps[i].StepID, clusterId, hostId,
+			steps[i].Command, steps[i].Args)
+	}
+}
+
+func addStep(steps models.Steps, stepType models.StepType, stepCmd string, stepArg []string) models.Steps {
```
gamli75, line 537: @filanov I'm not sure if I should use a pointer for steps in this method

internal/bminventory/inventory.go
```diff
@@ -531,30 +554,26 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params inventory.
 		return inventory.NewGetNextStepsNotFound()
 	}
 
-	if swag.StringValue(host.Status) == HostStatusDisabled {
-		return inventory.NewGetNextStepsOK().WithPayload(steps)
+	logrus.Infof("GetNextSteps host: <%s>, host status: <%s>", params.HostID, swag.StringValue(host.Status))
+
+	switch swag.StringValue(host.Status) {
+	case HostStatusKnown, HostStatusInsufficient, HostStatusDisconnected:
+		steps = addStep(steps, models.StepTypeConnectivityCheck, "", []string{})
```
gamli75, line 561: @filanov is it the correct way to send empty string/ array for non mandatory parameters?

ori-amizur:
In general it is OK, but it touches same places that I added for this PR https://github.com/filanov/bm-inventory/pull/32.  We will need to be in sync for that

internal/bminventory/inventory.go
filanov, line 537: i think that the right way to do it will be define a function on models.Steps
then you will have something like this:
```
step.addStep(...)
```

filanov: Pointer on what?

internal/bminventory/inventory.go
filanov: some times in easy to work with structs.
then when you don't set a parameter it will get default value. for string it's empty string.

gamli75:
Closing this PR - not relevant anymore

PR #32, opened 2020-04-05
MGMT-636 Add container running support in the agent

ori-amizur:
Add YAML file to describe available commands

### DIFF
--- a/Dockerfile.bm-inventory
+++ b/Dockerfile.bm-inventory
@@ -1,3 +1,4 @@
 FROM scratch
 ADD build/bm-inventory /bm-inventory
+COPY deploy/agent-commands.yaml  /etc/ocp-metal/agent-commands.yaml
 CMD ["/bm-inventory"]
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -64,7 +64,10 @@ func main() {
 		logrus.Fatal("failed to auto migrate, ", err)
 	}
 
-	bm := bminventory.NewBareMetalInventory(db, kclient, Options.BMConfig)
+	bm, err := bminventory.NewBareMetalInventory(db, kclient, Options.BMConfig)
+	if err != nil {
+		logrus.Fatalf("Error creating baremetal inventory: %s", err.Error())
+	}
 	h, err := restapi.Handler(restapi.Config{
 		InventoryAPI: bm,
 		Logger:       logrus.Printf,
--- a/deploy/agent-commands.yaml
+++ b/deploy/agent-commands.yaml
@@ -1,0 +1,9 @@
+commands:
+  - name: hardware-info
+    command: podman
+    arguments: [run,--rm,--privileged,--quiet,--net=host,-v,/var/log:/var/log,quay.io/oamizur/hardware_info,/usr/bin/hardware_info]
+    expected-argument-num: 0
+  - name: connectivity-check
+    command: podman
+    arguments: [run,--rm,--privileged,--quiet,--net=host,-v,/var/log:/var/log,quay.io/oamizur/connectivity_check,/usr/bin/connectivity_check]
+    expected-argument-num: 1
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,8 @@ require (
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
 	github.com/sirupsen/logrus v1.4.2
+	golang.org/x/net v0.0.0-20200226121028-0de0cce0169b // indirect
+	golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 // indirect
 	gopkg.in/yaml.v2 v2.2.8
 	k8s.io/api v0.17.3
 	k8s.io/apimachinery v0.17.3
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -3,9 +3,12 @@ package bminventory
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"strings"
 	"sync"
+
+	"gopkg.in/yaml.v2"
 
 	"github.com/filanov/bm-inventory/internal/installcfg"
 	"github.com/filanov/bm-inventory/models"
@@ -42,10 +45,17 @@ const (
 	ResourceKindCluster = "cluster"
 )
 
+const AgentsCommandsFile = "/etc/ocp-metal/agent-commands.yaml"
+
+const (
+	HardwareInfo      = "hardware-info"
+	ConnectivityCheck = "connectivity-check"
+)
+
 type Config struct {
 	ImageBuilder    string `envconfig:"IMAGE_BUILDER" default:"quay.io/oscohen/installer-image-build"`
 	ImageBuilderCmd string `envconfig:"IMAGE_BUILDER_CMD" default:"echo hello"`
-	AgentDockerImg  string `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/oamizur/introspector:latest"`
+	AgentDockerImg  string `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/oamizur/agent:latest"`
 	InventoryURL    string `envconfig:"INVENTORY_URL" default:"10.35.59.36"`
 	InventoryPort   string `envconfig:"INVENTORY_PORT" default:"30485"`
 	S3EndpointURL   string `envconfig:"S3_ENDPOINT_URL" default:"http://10.35.59.36:30925"`
@@ -69,9 +79,9 @@ const ignitionConfigFormat = `{
   },
 "systemd": {
 "units": [{
-"name": "introspector.service",
+"name": "agent.service",
 "enabled": true,
-"contents": "[Service]\nType=simple\nExecStartPre=docker run --privileged --rm -v /usr/local/bin:/hostbin %s cp /usr/bin/introspector /usr/sbin/dmidecode /hostbin\nExecStart=/usr/local/bin/introspector --host %s --port %s --cluster-id %s\n\n[Install]\nWantedBy=multi-user.target"
+"contents": "[Service]\nType=simple\nExecStartPre=docker run --privileged --rm -v /usr/local/bin:/hostbin %s cp /usr/bin/agent /hostbin\nExecStart=/usr/local/bin/agent --host %s --port %s --cluster-id %s\n\n[Install]\nWantedBy=multi-user.target"
 }]
 }
 }`
@@ -81,6 +91,17 @@ type debugCmd struct {
 	stepID string
 }
 
+type Command struct {
+	Name                string
+	Command             string
+	Arguments           []string
+	ExpectedArgumentNum int
+}
+
+type AgentCommands struct {
+	Commands []Command
+}
+
 type bareMetalInventory struct {
 	Config
 	imageBuildCmd []string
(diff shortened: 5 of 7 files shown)


Dockerfile.bm-inventory
```diff
@@ -1,3 +1,5 @@
-FROM scratch
+FROM alpine:3.7
```
filanov, line 1: it's not change the size drastically but why?

deploy/agent-commands.yaml
```diff
@@ -0,0 +1,9 @@
+commands:
+  - name: hardware-info
+    command: podman
+    arguments: [run,--rm,--privileged,--quiet,--net=host,-v,/var/log:/var/log,quay.io/oamizur/hardware_info,/usr/bin/hardware_info]
```
filanov, line 4: what about parameters?

go.mod
```diff
@@ -16,6 +16,8 @@ require (
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
 	github.com/sirupsen/logrus v1.4.2
+	golang.org/x/net v0.0.0-20200226121028-0de0cce0169b // indirect
```
filanov, line 19: please split commits for go.mod and go.sum

internal/bminventory/inventory.go
```diff
@@ -42,10 +45,12 @@ const (
 	ResourceKindCluster = "cluster"
 )
 
+const AgentsCommadsFile = "/etc/ocp-metal/agent-commands.yaml"
+
 type Config struct {
 	ImageBuilder    string `envconfig:"IMAGE_BUILDER" default:"quay.io/oscohen/installer-image-build"`
 	ImageBuilderCmd string `envconfig:"IMAGE_BUILDER_CMD" default:"echo hello"`
-	AgentDockerImg  string `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/oamizur/introspector:latest"`
+	AgentDockerImg  string `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/oamizur/agent:latest"`
```
filanov, line 53: we are changing the agent permanently?

internal/bminventory/inventory.go
```diff
@@ -81,20 +86,76 @@ type debugCmd struct {
...
 }
 
 func NewBareMetalInventory(db *gorm.DB, kclient client.Client, cfg Config) *bareMetalInventory {
 	b := &bareMetalInventory{db: db, kube: kclient, Config: cfg, debugCmdMap: make(map[strfmt.UUID]debugCmd)}
 	if cfg.ImageBuilderCmd != "" {
 		b.imageBuildCmd = strings.Split(cfg.ImageBuilderCmd, " ")
 	}
+
+	// TODO: Maybe panic on error
+	_ = b.readAgentCommands()
```
filanov, line 158: no panic here. only in the main.
please change `NewBareMetalInventory` to return an error as well and use `Fatal` log if you have an error

internal/bminventory/inventory.go
```diff
@@ -81,20 +86,76 @@ type debugCmd struct {
...
+
+func (b *bareMetalInventory) readAgentCommands() error {
+	contents, err := ioutil.ReadFile(AgentsCommadsFile)
+	if err != nil {
+		logrus.WithError(err).Errorf("Could not read file %s", AgentsCommadsFile)
+		return err
+	}
+	err = yaml.Unmarshal(contents, &b.agentCommands)
+	if err != nil {
+		logrus.WithError(err).Errorf("Error unmarshaling %s", contents)
```
filanov, line 118: return error?

internal/bminventory/inventory.go
```diff
@@ -81,20 +86,76 @@ type debugCmd struct {
...
+	return nil
+}
+
+func (b *bareMetalInventory) getCommandByName(name string) *Command {
+	for _, cmd := range b.agentCommands.Commands {
+		if cmd.Name == name {
+			return &cmd
+		}
+	}
+	logrus.Warnf("Command %s was not found", name)
```
filanov, line 129: is that really just a warning?

internal/bminventory/inventory.go
```diff
@@ -81,20 +86,76 @@ type debugCmd struct {
...
+	return nil
+}
+
+func (b *bareMetalInventory) createAgentExecuteStep(cmdName string, args ...string) *models.Step {
+	cmd := b.getCommandByName(cmdName)
+	if cmd == nil {
+		return nil
+	}
+	if len(args) != cmd.ExpectedArgumentNum {
+		logrus.Warnf("Argument number mismatch: expected %d received %d", cmd.ExpectedArgumentNum, len(args))
```
filanov, line 139: why just a warning?

internal/bminventory/inventory.go
```diff
@@ -553,10 +614,10 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params inventory.
 	}
 	b.debugCmdMux.Unlock()
 
-	steps = append(steps, &models.Step{
-		StepType: models.StepTypeHardawareInfo,
-		StepID:   createStepID(models.StepTypeHardawareInfo),
-	})
+	hardwareInfoStep := b.createAgentExecuteStep("hardware-info")
```
filanov, line 617: please sync with @gamli75 you are changing the same code.

subsystem/host_test.go
```diff
@@ -103,16 +104,16 @@ var _ = Describe("Host tests", func() {
 		var step *models.Step
 		var ok bool
 		// debug should be only for host1
-		_, ok = getStepInList(getNextSteps(clusterID, *host2.ID), models.StepTypeExecute)
+		_, ok = getStepByIdPrefix(getNextSteps(clusterID, *host2.ID), "execute")
```
filanov, line 107: maybe should a const on `execute`

internal/bminventory/inventory.go
filanov, line 617: maybe should be const `hardware-info`

filanov:
In general i'm not sure how i feel about this change. 
if we add a new command and the code needs to change as well so what is the added value here?
we can just store the commands in a go struct and do the same no?

Dockerfile.bm-inventory
ori-amizur: For mkdir

filanov: ok

deploy/agent-commands.yaml
ori-amizur: For now it is only positional parameters.  Verification is done on the number of parameters that there is a complete match.

subsystem/host_test.go
ori-amizur: The value is received from the YAML.  We are not going to "const" them all

filanov: it's used more then once in this file.

internal/bminventory/inventory.go
```diff
@@ -81,20 +86,76 @@ type debugCmd struct {
...
+}
+
 type bareMetalInventory struct {
 	Config
 	imageBuildCmd []string
 	db            *gorm.DB
 	kube          client.Client
 	debugCmdMap   map[strfmt.UUID]debugCmd
 	debugCmdMux   sync.Mutex
+	agentCommands AgentCommands
```
ronniel1, line 107: Shouldn't this be a map of commands, with the name being the key?
Why search each time for the command?

internal/bminventory/inventory.go
```diff
@@ -81,20 +86,76 @@ type debugCmd struct {
...
+		}
+	}
+	logrus.Warnf("Command %s was not found", name)
+	return nil
+}
+
+func (b *bareMetalInventory) createAgentExecuteStep(cmdName string, args ...string) *models.Step {
+	cmd := b.getCommandByName(cmdName)
+	if cmd == nil {
+		return nil
```
ronniel1, line 136: return error

filanov:
> In general i'm not sure how i feel about this change.
> if we add a new command and the code needs to change as well so what is the added value here?
> we can just store the commands in a go struct and do the same no?
> 
> @ori-amizur @ronniel1 what do you think?

internal/bminventory/inventory.go
filanov: +1

filanov:
> In general i'm not sure how i feel about this change.
> if we add a new command and the code needs to change as well so what is the added value here?
> we can just store the commands in a go struct and do the same no?
> 
> @ori-amizur @ronniel1 what do you think?

maybe we can combine it with the state machine. and each command will have the states states that this command should be executed on? 

but we still will need to change the code for new parameters.

Dockerfile.bm-inventory
```diff
@@ -1,3 +1,5 @@
...
 ADD build/bm-inventory /bm-inventory
+RUN mkdir -p /etc/ocp-metal
+ADD deploy/agent-commands.yaml  /etc/ocp-metal/agent-commands.yaml
```
filanov, line 4: if you are using add or copy maybe you will no need to use `alpine:3.7`

ori-amizur: Yes.  You are right

subsystem/host_test.go
ori-amizur: OK.  But we will need to const all names in the YAML file.

internal/bminventory/inventory.go
ori-amizur: agent seems more suitable name.  Still it can be changed by environment variable.

internal/bminventory/inventory.go
ori-amizur: As a standard, we can decide that all names that appear in the YAML file should be const.

internal/bminventory/inventory.go
ori-amizur: Error might be more suitable

internal/bminventory/inventory.go
filanov: i asked to know that the target build in the agent changed as well?

internal/bminventory/inventory.go
ori-amizur: Error is useless here since the error was already logged

internal/bminventory/inventory.go
ori-amizur: I think so.  I think the name fits better.

internal/bminventory/inventory.go
ori-amizur: OK

internal/bminventory/inventory.go
filanov: sure

internal/bminventory/inventory.go
ori-amizur: Will be changed to error

subsystem/host_test.go
ori-amizur: OK

ori-amizur:
> > In general i'm not sure how i feel about this change.
> > if we add a new command and the code needs to change as well so what is the added value here?
> > we can just store the commands in a go struct and do the same no?
> > @ori-amizur @ronniel1 what do you think?
> 
> maybe we can combine it with the state machine. and each command will have the states that this command should be executed on?
> 
> but we still will need to change the code for new parameters.

Assuming that we need some kind of definition of what can be executed, then I think that YAML file is better, since there is separation between the logic (which is the GO code), and the definition of what needs to be executed.
Adding new commands should be easier, although it is also hard-coded since it is bundled with the image.
I am not so aware what you plan for the state machine, but it does not seem to me a good idea that execution should be bundled with state machine.  A state machine should be separate from its operations, whether they are queries or just unconditional action.

filanov:
> > > In general i'm not sure how i feel about this change.
> > > if we add a new command and the code needs to change as well so what is the added value here?
> > > we can just store the commands in a go struct and do the same no?
> > > @ori-amizur @ronniel1 what do you think?
> > 
> > 
> > maybe we can combine it with the state machine. and each command will have the states that this command should be executed on?
> > but we still will need to change the code for new parameters.
> 
> Assuming that we need some kind of definition of what can be executed, then I think that YAML file is better, since there is separation between the logic (which is the GO code), and the definition of what needs to be executed.
> Adding new commands should be easier, although it is also hard-coded since it is bundled with the image.
> I am not so aware what you plan for the state machine, but it does not seem to me a good idea that execution should be bundled with state machine. A state machine should be separate from its operations, whether they are queries or just unconditional action.

The state machine that i'm taking about is to map between the operation and the state of the node.
for example hw info need to executed when the host in discovering state. 
This way we need less logic in the code. the yaml will define the behavior

cmd/main.go
```diff
@@ -64,7 +64,10 @@ func main() {
 		logrus.Fatal("failed to auto migrate, ", err)
 	}
 
-	bm := bminventory.NewBareMetalInventory(db, kclient, Options.BMConfig)
+	bm, err := bminventory.NewBareMetalInventory(db, kclient, Options.BMConfig)
+	if err != nil {
+		logrus.Fatalf("Error creating baremetal inventory: %s", err.Error())
```
filanov, line 69:
```suggestion
		logrus.WithError(err).Fatal("Failed to create baremetal inventory")
```

internal/bminventory/inventory.go
```diff
@@ -42,10 +45,17 @@ const (
 	ResourceKindCluster = "cluster"
 )
 
+const AgentsCommandsFile = "/etc/ocp-metal/agent-commands.yaml"
+
+const (
+	HardwareInfo      = "hardware-info"
```
filanov, line 51:
```suggestion
	StepTypeHardwareInfo      = "hardware-info"
	StepTypeConnectivityCheck = "connectivity-check"
```

internal/bminventory/inventory.go
```diff
@@ -81,21 +91,84 @@ type debugCmd struct {
...
+		b.agentCommands[cmd.Name] = &cmd
+	}
+	return nil
+}
+
+func (b *bareMetalInventory) getCommandByName(name string) *Command {
+	cmd, ok := b.agentCommands[name]
+	if !ok {
+		logrus.Errorf("Command %s was not found", name)
+		return nil
```
filanov, line 138: why not return an error?

internal/bminventory/inventory.go
filanov: i think that @ronniel1  is not talking about the log. We tried to execute a command that doesn't exists. the api should return an error.

internal/bminventory/inventory.go
```diff
@@ -81,21 +91,84 @@ type debugCmd struct {
...
+}
+
+func NewBareMetalInventory(db *gorm.DB, kclient client.Client, cfg Config) (*bareMetalInventory, error) {
 	b := &bareMetalInventory{db: db, kube: kclient, Config: cfg, debugCmdMap: make(map[strfmt.UUID]debugCmd)}
 	if cfg.ImageBuilderCmd != "" {
 		b.imageBuildCmd = strings.Split(cfg.ImageBuilderCmd, " ")
 	}
-	return b
+
+	err := b.readAgentCommands()
```
filanov, line 167:
```suggestion
	if err := b.readAgentCommands(); err !=nil {
	...
```

internal/bminventory/inventory.go
```diff
@@ -553,10 +626,10 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params inventory.
 	}
 	b.debugCmdMux.Unlock()
 
-	steps = append(steps, &models.Step{
-		StepType: models.StepTypeHardawareInfo,
-		StepID:   createStepID(models.StepTypeHardawareInfo),
-	})
+	hardwareInfoStep := b.createAgentExecuteStep(HardwareInfo)
+	if hardwareInfoStep != nil {
```
filanov, line 630: if hardwareInfoStep is nil then it's an error.

internal/bminventory/inventory.go
ori-amizur: The StepTypeHardwareInfo will be removed

internal/bminventory/inventory.go
ori-amizur: The error was already logged.  If nil is returned then the step will mot be executed.  What else should be done?

internal/bminventory/inventory.go
ori-amizur: Correct

internal/bminventory/inventory.go
filanov: i'm taking about a prefix to the const. it will be clear what is the use of it

internal/bminventory/inventory.go
filanov: i'm not talking about logging, we wanted to add a command and failed, something is wrong.

internal/bminventory/inventory.go
filanov: it's more golike to do it this way.

internal/bminventory/inventory.go
filanov: then it should return error no?

PR #35, opened 2020-04-06
Added request-ID logging to inventory

ronniel1:


cmd/main.go
```diff
@@ -31,47 +33,49 @@ var Options struct {
...
 	if err != nil {
-		logrus.Fatal("Fail to connect to DB, ", err)
+		log.Fatal("Fail to connect to DB, ", err)
 	}
 	defer db.Close()
 
 	scheme := runtime.NewScheme()
 	if err := clientgoscheme.AddToScheme(scheme); err != nil {
-		logrus.Fatal()
+		log.Fatal()
```
filanov, line 58: probably should add the error to the log. i know i wasn't there but it should be.

cmd/main.go
```diff
@@ -31,47 +33,49 @@ var Options struct {
 }
 
 func main() {
+	log := logrus.New()
```
filanov, line 36: i think that you don't have to create a new logger if you don't set any specific configuration on it. 
if you like more the name log then you change the import to `log "github.com/sirupsen/logrus"`

PR #36, opened 2020-04-13
Osher/mgmt 659

oshercc:


swagger.yaml
```diff
@@ -175,11 +175,32 @@ paths:
         500:
           description: Internal server error
 
-  /clusters/{clusterId}/downloads/kubeconfig:
+  /clusters/{clusterId}/generate/kubeconfig:
```
filanov, line 178: As far as i understand the user should not get control of the files generation.
Those files need to be generated once when the cluster is ready for installation.
So the API is still download but it should return conflict until the installation is tarted. 
To be sure you can ask Avishay and Eran

internal/bminventory/inventory.go
```diff
@@ -767,41 +767,45 @@ func (b *bareMetalInventory) createKubeconfigJob(ctx context.Context, cluster *m
...
 		return inventory.NewDownloadClusterKubeconfigInternalServerError()
 	}
+
 	if resp.StatusCode != http.StatusOK {
-		defer resp.Body.Close()
-		b, _ := ioutil.ReadAll(resp.Body)
-		log.WithError(fmt.Errorf("%s", string(b))).
-			Errorf("Failed to get clusters %s kubeKonfig", params.ClusterID)
-		return inventory.NewDownloadClusterISOInternalServerError()
+		log.Errorf("Failed to get clusters %s %s file", params.ClusterID, params.FileName)
```
filanov, line 805: please take a look at the code that you have deleted. this code actually return the body from the error message so i think you should use it.

swagger.yaml
oshercc: @eranco74 @avishayt  what do you think ?

avishayt: When we generate the files is an implementation detail.  Users don't care - they just want to download.  Please leave it as it was.

internal/bminventory/inventory.go
```diff
@@ -342,6 +342,12 @@ func getImageName(clusterID strfmt.UUID) string {
 func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventory.InstallClusterParams) middleware.Responder {
 	log := logutil.FromContext(ctx, b.log)
 	var cluster models.Cluster
+
+	responder, done := generateClusterKubeconfig(b, cluster, params, log, ctx)
```
filanov, line 346: 1. Kubeconfig is just one of the files that are being generated, So maybe `generateClusterInstallConfig`? 
2. We should generate the files once we pass some validations. there is a comment in the code where it should be done.
3. this functions already generate install config yaml. so if you already have it in your code then you don't need it twice.

internal/bminventory/inventory.go
```diff
@@ -342,6 +342,12 @@ func getImageName(clusterID strfmt.UUID) string {
...
+	if done {
+		return responder
```
filanov, line 348: why do you return anything?

internal/bminventory/inventory.go
```diff
@@ -397,6 +403,24 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
 	return inventory.NewInstallClusterOK().WithPayload(&cluster)
 }
 
+func generateClusterKubeconfig(b *bareMetalInventory, cluster models.Cluster, params inventory.InstallClusterParams, log logrus.FieldLogger, ctx context.Context) (middleware.Responder, bool) {
+	if err := b.db.Preload("Hosts").First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
```
filanov, line 407: if you are doing it as part of the install functions there is already a code that get the cluster. so you don't really need to validate anything.

internal/bminventory/inventory.go
```diff
@@ -397,6 +403,24 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
...
+		log.WithError(err).Errorf("failed to get cluster %s", params.ClusterID)
+		return inventory.NewInstallClusterNotFound(), true
+	}
+
+	if err := b.createKubeconfigJob(ctx, &cluster); err != nil {
+		log.WithError(err).Error("Failed to create kubeconfig generation job")
```
filanov, line 413: cluster id can be helpful when you take a look at the logs.

internal/bminventory/inventory.go
```diff
@@ -397,6 +403,24 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
...
+		return inventory.NewInstallClusterInternalServerError(), true
+	}
+
+	if err := b.monitorJob(ctx, fmt.Sprintf("%s-%s", kubeconfigPrefix, params.ClusterID)); err != nil {
+		log.WithError(err).Error("Generating kubeconfig files failed")
```
filanov, line 418: cluster id can be helpful when you take a look at the logs.

swagger.yaml
```diff
@@ -175,11 +175,11 @@ paths:
         500:
           description: Internal server error
 
-  /clusters/{clusterId}/downloads/kubeconfig:
+  /clusters/{clusterId}/{fileName}/downloads/kubeconfig:
```
filanov, line 178: maybe filename should be a query param? 
@avishayt what do you think?

avishayt: Yes, query param

internal/bminventory/inventory.go
```diff
@@ -367,6 +368,10 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
 	fmt.Println("Install config: \n", string(cfg))
 
 	// generate ignition files from install-config.yaml
+	responder, done := generateClusterInstallConfig(b, cluster, params, log, ctx, cfg)
+	if done {
```
filanov, line 372: what do you return here?!

internal/bminventory/inventory.go
```diff
@@ -397,6 +402,20 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
...
 }
 
+func generateClusterInstallConfig(b *bareMetalInventory, cluster models.Cluster, params inventory.InstallClusterParams, log logrus.FieldLogger, ctx context.Context, cfg []byte) (middleware.Responder, bool) {
+
+	if err := b.createKubeconfigJob(ctx, &cluster, cfg); err != nil {
+		log.WithError(err).Errorf("Failed to create kubeconfig generation job for cluster %s", cluster.ID)
+		return inventory.NewInstallClusterInternalServerError(), true
+	}
+
+	if err := b.monitorJob(ctx, fmt.Sprintf("%s-%s", kubeconfigPrefix, params.ClusterID)); err != nil {
```
filanov, line 412: Maybe `createKubeconfigJob` should return a job name. because it's not clear why do you have `fmt.Sprintf("%s-%s", kubeconfigPrefix, params.ClusterID)`

subsystem/cluster_test.go
```diff
@@ -149,10 +149,10 @@ var _ = Describe("system-test cluster install", func() {
 		defer os.Remove(file.Name())
 
 		missingClusterId := strfmt.UUID(uuid.New().String())
-		_, err = bmclient.Inventory.DownloadClusterKubeconfig(ctx, &inventory.DownloadClusterKubeconfigParams{ClusterID: missingClusterId}, file)
-		Expect(err).Should(MatchError(inventory.NewDownloadClusterKubeconfigNotFound()))
+		_, err = bmclient.Inventory.DownloadClusterKubeconfig(ctx, &inventory.DownloadClusterKubeconfigParams{ClusterID: missingClusterId, FileName: "bootstrap.ign"}, file)
+		Expect(err).Should(MatchError(inventory.NewDownloadClusterKubeconfigInternalServerError()))
```
filanov, line 153: it's more of a conflict. because the installation didn't started the files didn't generated yet.

internal/bminventory/inventory.go
oshercc: the err

internal/bminventory/inventory.go
```diff
@@ -397,6 +401,21 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
 	return inventory.NewInstallClusterOK().WithPayload(&cluster)
 }
 
+func generateClusterInstallConfig(b *bareMetalInventory, cluster models.Cluster, params inventory.InstallClusterParams, log logrus.FieldLogger, ctx context.Context, cfg []byte) middleware.Responder {
```
filanov, line 404: this function can be part of the inventory no need to pass it as a parameter

swagger.yaml
```diff
@@ -179,7 +179,7 @@ paths:
     get:
       tags:
         - inventory
-      summary: Download the kubeconfig file for the specified cluster
+      summary: Download the kubeconfig files for the specified cluster
```
filanov, line 182: @avishayt this api is no longer download kubeconfig, maybe the name should change?

avishayt: As discussed on slack, let's make one API for all downloads:

GET /clusters/{clusterId}/files?<file name>

internal/bminventory/inventory.go
filanov: please don't resolve issues unless it absolutely clear.
this issue is still relevant. 
next time please leave the issue resolve to the one that wrote the issue.

internal/bminventory/inventory.go
```diff
@@ -432,6 +436,21 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
...
+
+	jobNmae, err := b.createKubeconfigJob(ctx, &cluster, cfg)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to create kubeconfig generation job for cluster %s", cluster.ID)
+		return inventory.NewInstallClusterInternalServerError()
+	}
+
+	if err := b.monitorJob(ctx, jobNmae); err != nil {
+		log.WithError(err).Errorf("Generating kubeconfig files failed for cluster %s", cluster.ID)
```
filanov, line 448: Please add job name to the log.

internal/bminventory/inventory.go
```diff
@@ -432,6 +436,21 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
 	return inventory.NewInstallClusterOK().WithPayload(&cluster)
 }
 
+func generateClusterInstallConfig(b *bareMetalInventory, cluster models.Cluster, params inventory.InstallClusterParams, log logrus.FieldLogger, ctx context.Context, cfg []byte) middleware.Responder {
+
+	jobNmae, err := b.createKubeconfigJob(ctx, &cluster, cfg)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to create kubeconfig generation job for cluster %s", cluster.ID)
```
filanov, line 443: please add job name to the log.

internal/bminventory/inventory.go
filanov: please don't resolve issues unless it absolutely clear.
this issue is still relevant.
next time please leave the issue resolve to the one that wrote the issue.

internal/bminventory/inventory.go
```diff
@@ -788,46 +802,29 @@ func (b *bareMetalInventory) createKubeconfigJob(ctx context.Context, cluster *m
...
+		return "", err
 	}
-	return nil
+	return jobName, nil
 }
 
-func (b *bareMetalInventory) DownloadClusterKubeconfig(ctx context.Context, params inventory.DownloadClusterKubeconfigParams) middleware.Responder {
-	log := logutil.FromContext(ctx, b.log)
-	var cluster models.Cluster
+func (b *bareMetalInventory) DownloadClusterFiles(ctx context.Context, params inventory.DownloadClusterFilesParams) middleware.Responder {
```
filanov, line 810: I think that there is a bug here.
if we start the installation and generate the files but then the api will fail because some of the hosts are in a problematic state. the the result is that you have a generated files that the user can download.
The user will download the files
Meanwhile he will handle the problematic hosts and try install again, so the files will be generated again with different hashes, secret etc. 
i think it will be difficult for the user to understand that he needs to get the files again.

Maybe we should enable download only if the cluster is in installing state so we know the he already started?

@ronniel1 @avishayt what do you think?

internal/bminventory/inventory.go
```diff
@@ -402,6 +403,9 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
 	fmt.Println("Install config: \n", string(cfg))
 
 	// generate ignition files from install-config.yaml
+	if err := generateClusterInstallConfig(b, cluster, params, log, ctx, cfg); err != nil {
+		return err
```
filanov, line 407: you have an active transaction. you need to call rollback before returning an error

internal/bminventory/inventory.go
avishayt: Discussed f2f, yes - files should be available for download only in installing/installed states.

internal/bminventory/inventory.go
oshercc: didn't push it, sorry, i'll make sure to pay more attention next time

internal/bminventory/inventory.go
```diff
@@ -414,7 +421,7 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
 	}
 
 	if err := tx.Model(&models.Cluster{}).Where("id = ?", cluster.ID.String()).
-		Update("status", "installing").Error; err != nil {
+		Update("status", swag.String(ClusterStatusInstalling)).Error; err != nil {
```
filanov, line 424: you don't need `swag.String` here.

internal/bminventory/inventory.go
```diff
@@ -788,46 +805,40 @@ func (b *bareMetalInventory) createKubeconfigJob(ctx context.Context, cluster *m
...
+		log.WithError(err).Errorf("failed to find cluster %s", params.ClusterID)
+		return inventory.NewDownloadClusterFilesInternalServerError()
 	}
-
-	if err := b.monitorJob(ctx, fmt.Sprintf("%s-%s", kubeconfigPrefix, params.ClusterID)); err != nil {
-		log.WithError(err).Error("Generating kubeconfig files failed")
-		return inventory.NewDownloadClusterKubeconfigInternalServerError()
+	clusterStatus := swag.StringValue(cluster.Status)
+	if clusterStatus != ClusterStatusInstalling && clusterStatus != ClusterStatusInstalled {
+		log.Errorf("Cluster %s is in %s state, files can be downloaded only in installing or installed state", params.ClusterID, clusterStatus)
```
filanov, line 824: it's not a real error. it can be a warning

subsystem/cluster_test.go
```diff
@@ -167,6 +167,14 @@ var _ = Describe("system-test cluster install", func() {
 		})
 		Expect(err).NotTo(HaveOccurred())
 
+		//Test downloading kubeconfig files in worng state
+		file, err := ioutil.TempFile("", "tmp")
```
filanov, line 171: this is already a long test. please split it to a different test.

internal/bminventory/inventory.go
```diff
@@ -793,45 +807,38 @@ func (b *bareMetalInventory) createKubeconfigJob(ctx context.Context, cluster *m
...
-	if err := b.db.Preload("Hosts").First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
-		log.WithError(err).Errorf("failed to get cluster %s", params.ClusterID)
-		return inventory.NewDownloadClusterKubeconfigNotFound()
-	}
-
-	jobName := fmt.Sprintf("%s-%s", kubeconfigPrefix, params.ClusterID)
-	if err := b.createKubeconfigJob(ctx, &cluster, jobName); err != nil {
-		log.WithError(err).Error("Failed to create kubeconfig generation job")
-		return inventory.NewDownloadClusterKubeconfigInternalServerError()
+	if err := b.db.First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
```
filanov, line 818: need to check the error type. it can be 404 or 500

internal/bminventory/inventory.go
```diff
@@ -793,45 +807,38 @@ func (b *bareMetalInventory) createKubeconfigJob(ctx context.Context, cluster *m
...
 	if resp.StatusCode != http.StatusOK {
 		defer resp.Body.Close()
 		b, _ := ioutil.ReadAll(resp.Body)
 		log.WithError(fmt.Errorf("%s", string(b))).
 			Errorf("Failed to get clusters %s kubeKonfig", params.ClusterID)
-		return inventory.NewDownloadClusterISOInternalServerError()
+		return inventory.NewDownloadClusterFilesConflict()
 	}
-	return filemiddleware.NewResponder(inventory.NewDownloadClusterKubeconfigOK().WithPayload(resp.Body), "kubeconfig")
+	return filemiddleware.NewResponder(inventory.NewDownloadClusterFilesOK().WithPayload(resp.Body), "kubeconfig")
```
filanov, line 843: please replace `kubeconfig` with `params.FileName`

subsystem/cluster_test.go
```diff
@@ -174,17 +174,54 @@ var _ = Describe("system-test cluster install", func() {
 		for _, host := range c.GetPayload().Hosts {
 			Expect(swag.StringValue(host.Status)).Should(Equal("installing"))
 		}
+	})
+
+	It("install download_config_files", func() {
```
filanov, line 179: if you copy a code then you can probably set it under a single `Context` and run it in `BeforeEach`
no need to copy past the same setup twice.

PR #37, opened 2020-04-16
Michaelf host-api

filanov:
Added a host state machine.
There is an interface that support several operation that can be performed on a host.
In general i use state pattern. Each state have its own implementation for the api, i hope that it will make things clearer.


internal/host/common.go
```diff
@@ -0,0 +1,61 @@
...
+	if time.Since(time.Time(h.UpdatedAt)) > 3*time.Minute {
+		return updateTo(log, hostStatusDisconnected, h, db)
+	}
+	return &UpdateReply{
+		State:     swag.StringValue(h.Status),
+		IsChanged: false,
+	}, nil
+}
+
+func updateToWithParams(log logrus.FieldLogger, status string, h *models.Host, db *gorm.DB, extra ...interface{}) (*UpdateReply, error) {
```
ronniel1, line 43: updateStateWithParams

internal/host/common.go
```diff
@@ -0,0 +1,61 @@
...
+	updates := map[string]interface{}{"status": status}
+	if len(extra)%2 != 0 {
+		return nil, errors.Errorf("invalid update extra parameters %+v", extra)
+	}
+	for i := 0; i < len(extra); i += 2 {
+		updates[extra[i].(string)] = extra[i+1]
+	}
+	if err := db.Model(&models.Host{}).Where("id = ? and cluster_id = ? and status = ?",
```
ronniel1, line 51: Maybe we should log which fields we are updating.
Since we cannot know from code what we are updating, we better have a log of it

internal/host/common.go
```diff
@@ -0,0 +1,61 @@
...
+		h.ID.String(), h.ClusterID.String(), swag.StringValue(h.Status)).
+		Updates(updates).Error; err != nil {
+		return nil, err
```
ronniel1, line 54: We might fail here for multiple reasons.
One of them is retryable (due to the where clause). Shouldn't we return a specific value, in that case, so users know that they should retry the operation?

Also, I don't exactly remember, but if the 'where' clause failed, doesn't it return success, and we need to look at the count of updated rows?

internal/host/disabled.go
```diff
@@ -0,0 +1,68 @@
...
+func (d *disabledState) RegisterHost(ctx context.Context, h *models.Host) (*UpdateReply, error) {
+	return nil, errors.Errorf("unable to register host <%s> in <%s> status",
+		h.ID, swag.StringValue(h.Status))
+}
+
+func (d *disabledState) UpdateHwInfo(ctx context.Context, h *models.Host, hwInfo string) (*UpdateReply, error) {
+	return nil, errors.Errorf("unable to update hardware info to host <%s> in <%s> status",
+		h.ID, swag.StringValue(h.Status))
+}
+
```
ronniel1, line 35: Maybe in these cases we should perform the update?
The reason for the disabled is that if a node is problematic, or doesn't pass the minimum requirements, it should not keep showing an error to the user.
The user can mark it as disabled, and not see the errors/warnings anymore.
However, if it updates us, maybe now it passed the minimum requirements, why not update the DB? The user will be able to see that the new HW specs and choose if it wants to re-enable it.

@avishayt  What do you think?

internal/host/disabled.go
```diff
@@ -0,0 +1,68 @@
...
+func (d *disabledState) UpdateRole(ctx context.Context, h *models.Host, role string, db *gorm.DB) (*UpdateReply, error) {
```
ronniel1, line 36: Why is it possible to update the role in this state?

internal/host/disabled.go
```diff
@@ -0,0 +1,68 @@
...
+func (d *disabledState) RefreshStatus(ctx context.Context, h *models.Host) (*UpdateReply, error) {
+	// State in the same state
+	return &UpdateReply{
+		State:     hostStatusDisabled,
+		IsChanged: false,
+	}, nil
+}
+
+func (d *disabledState) Install(ctx context.Context, h *models.Host, db *gorm.DB) (*UpdateReply, error) {
+	return nil, errors.Errorf("unable to update hardware info to host <%s> in <%s> status",
```
ronniel1, line 53: Wrong message

internal/host/disconnected.go
```diff
@@ -0,0 +1,78 @@
...
+	db          *gorm.DB
+	hwValidator hardware.Validator
+}
+
+func (d *disconnectedState) RegisterHost(ctx context.Context, h *models.Host) (*UpdateReply, error) {
+	return updateToWithParams(logutil.FromContext(ctx, d.log), hostStatusDiscovering, h, d.db,
+		"hardware_info", "")
+}
+
+func (d *disconnectedState) UpdateHwInfo(ctx context.Context, h *models.Host, hwInfo string) (*UpdateReply, error) {
```
ronniel1, line 34: If we got this , shouldn't the host become not disconnected first?

internal/host/disconnected.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	if !reply.IsSufficient {
+		return updateToWithParams(log, hostStatusInsufficient, h, d.db,
+			"hardware_info", hwInfo, "status_info", reply.Reason)
+	}
+	return updateToWithParams(log, hostStatusKnown, h, d.db, "hardware_info", hwInfo)
+}
+
+func (d *disconnectedState) UpdateRole(ctx context.Context, h *models.Host, role string, db *gorm.DB) (*UpdateReply, error) {
+	cdb := d.db
```
ronniel1, line 48: Is this valid at this state?

internal/host/common.go
```diff
@@ -0,0 +1,61 @@
...
+	hostStatusInstalled    = "installed"
+	hostStatusError        = "error"
+)
+
+type UpdateReply struct {
+	State     string
+	IsChanged bool
+}
+
+func updateTo(log logrus.FieldLogger, status string, h *models.Host, db *gorm.DB) (*UpdateReply, error) {
```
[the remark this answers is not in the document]
filanov: done

internal/host/common.go
filanov: done

internal/host/common.go
filanov: added the fields as well.

filanov: added all the fields to the log.

internal/host/common.go
filanov: Don't know about `Where` but i will add a check to see if there is an error and if nothing have changed.

internal/host/disabled.go
filanov: As far as i understand disable is for maintenance once the user enable the host again it will go to discovering so there is no reason to update the hw anyway.

internal/host/disabled.go
filanov: update hw is something that the agent will do. 
update role is a command that the user can perform. so for example the host is in maintenance but the user still should be able to change it's configuration no? 
after the host will be enabled again the state will remain.
host can be in a discovering mode and still have a role.

internal/host/disabled.go
filanov: fixed

internal/host/disconnected.go
filanov: if we have network issues the host will be disconnected.
if the network is back and the host will continue to ask for the next step he will get the hw as a reply and when he return the hw info it will be reconnected again.
should i add it as a comment?

internal/host/disconnected.go
filanov: why not? it's the same as i wrote regarding disabled `UpdateRole`
it's a state that will remain even after the host will be registered again.

internal/host/disabled.go
ronniel1: Lets say that the someone upgraded the hardware and restarted the node.
As it is, the user will not see that the HW was updated

internal/host/disabled.go
ronniel1: @avishayt  What do you think?

internal/host/disconnected.go
ronniel1: Isn't the "keepalive" that we get from the agent, is the "get-next-step" request?
If so, it should change the state to connected

internal/host/discovering.go
```diff
@@ -0,0 +1,90 @@
...
+		log:         log,
+		db:          db,
+		hwValidator: hwValidator,
+	}
+}
+
+type discoveringState struct {
+	log         logrus.FieldLogger
+	db          *gorm.DB
+	hwValidator hardware.Validator
```
ronniel1, line 26: These are the same in all states.
Should't we have something like this:

    type state struct {
        log
        ....
    }
    type discoveringState state
    type disconnectedState state

internal/host/discovering.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (d *discoveringState) RegisterHost(ctx context.Context, h *models.Host) (*UpdateReply, error) {
+	host := models.Host{}
```
ronniel1, line 30: This function seems common to multiple states.
Maybe it should be in common.go?

internal/host/discovering.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, err
+	}
+	return &UpdateReply{
+		State:     hostStatusDiscovering,
+		IsChanged: false,
+	}, nil
+}
+
+func (d *discoveringState) UpdateHwInfo(ctx context.Context, h *models.Host, hwInfo string) (*UpdateReply, error) {
+	log := logutil.FromContext(ctx, d.log)
```
ronniel1, line 50: same here

internal/host/error.go
```diff
@@ -0,0 +1,64 @@
...
+		log: log,
+		db:  db,
+	}
+}
+
+type errorState struct {
+	log logrus.FieldLogger
+	db  *gorm.DB
+}
+
```
ronniel1, line 24: How do we get out of error state?
Maybe if the node has registered again?
@avishayt ?

internal/host/installed.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func NewInstalledState(log logrus.FieldLogger, db *gorm.DB) API {
+	return &installedState{
+		log: log,
+		db:  db,
+	}
+}
+
+type installedState struct {
```
ronniel1, line 20: Isn't this the same as error state?

internal/host/disabled.go
filanov: If the host is disabled he will not get any requests so he should not notify the server and the service should ignore everything the host say no?
if the user will enable the host, he will go to discovery state then he will request the hw info.

internal/host/disconnected.go
filanov: There is no such state as connected but you are correct about the keepalive. 
Update hw info is not a keepalive but we can add an api like `KeepAlive` that will be activated every time the host ask for the next step there we will only update the field `updated_at` in the db. @ronniel1 what do you think?

internal/host/discovering.go
filanov: but then we don't have dependency injection no?

internal/host/discovering.go
filanov: Where? 
Maybe  you are talking about `updateHwInfo`?

internal/host/error.go
filanov: Good question.
Currently if something goes wrong you need to start over.
So delete the host and create a new cluster.
We need to find out how to handle installation errors.

internal/host/installed.go
filanov: the difference is the state on refresh and enable but it can be solved as well with `h.State`
but if we are going to go out form error state then the behavior will be different.
Do you want me to merge them to some idle state?

internal/host/disabled.go
avishayt: Again, agree with @filanov .  We might not have it in the UI (maybe disabled hosts will be in a separate area where the only action is to enable), but there is no harm or extra work in allowing it via API, so I don't see why not.  We will always need to make sure to ignore disabled hosts regardless, for example when making sure we have 3 masters.

internal/host/disconnected.go
avishayt: I don't understand why we need a new keepalive API.  I think what @ronniel1 meant (and if so I agree) that when we get a get-next-step request, we treat that as a keepalive.

internal/host/error.go
avishayt: Is there a way to handle errors?  I assumed that something went wrong during the installation that is irrecoverable and  beyond our control.  At this point the cluster installation might still succeed, and the node can be added again later.

internal/host/disabled.go
ronniel1: OK

internal/host/disabled.go
ronniel1: OK

internal/host/discovering.go
filanov: you are right, will do.

PR #39, opened 2020-04-16
Michaelf/hw validator

filanov:
This packaged is used in the host api (it's not integrated yet).
Once https://github.com/filanov/bm-inventory/pull/37 is integrated i will change and update subsystem to use correct parameters and bad flows as well.
(i will not resolve the ticket yet but this code can go in)

### DIFF
--- a/internal/hardware/validator.go
+++ b/internal/hardware/validator.go
@@ -1,6 +1,23 @@
 package hardware
 
-import "github.com/filanov/bm-inventory/models"
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/alecthomas/units"
+	"github.com/filanov/bm-inventory/models"
+)
+
+const (
+	// minimal CPU requirements
+	minCPUCores       = 2
+	minCpuCoresMaster = 4
+	minCpuCoresWorker = 2
+	// minimal RAM requirements
+	minRam       = int64(8 * units.Gibibyte)
+	minRamMaster = int64(16 * units.Gibibyte)
+	minRamWorker = int64(8 * units.Gibibyte)
+)
 
 type IsSufficientReply struct {
 	IsSufficient bool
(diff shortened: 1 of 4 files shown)


internal/hardware/validator.go
```diff
@@ -1,6 +1,23 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/alecthomas/units"
+	"github.com/filanov/bm-inventory/models"
+)
+
+const (
+	// minimal CPU requirements
+	minCPUCores       = 2
```
ronniel1, line 13: We should have a way to reduce these minimum requirements for testing

internal/hardware/validator.go
```diff
@@ -18,9 +35,52 @@ func NewValidator() Validator {
...
+	case "master":
+		minCpuCoresRequired = minCpuCoresMaster
+		minRamRequired = minRamMaster
+	case "worker":
+		minCpuCoresRequired = minCpuCoresWorker
+		minRamRequired = minRamWorker
+	}
+
+	if hwInfo.CPU.Cpus < minCpuCoresRequired {
+		reason += fmt.Sprintf("insufficient CPU cores expected: <%d> got <%d>", minCpuCoresRequired, hwInfo.CPU.Cpus)
```
ronniel1, line 61: Maybe also add the role to the message?

internal/hardware/validator.go
```diff
@@ -18,9 +35,52 @@ func NewValidator() Validator {
...
+	}
+
+	if total := sumMemory(hwInfo); total < minRamRequired {
+		reason += fmt.Sprintf(", insufficient RAM requirements <%d> < <%d>", total, minRamRequired)
```
ronniel1, line 65: Maybe also add the role to the message?

internal/hardware/validator.go
```diff
@@ -18,9 +35,52 @@ func NewValidator() Validator {
...
 	return &IsSufficientReply{
-		IsSufficient: true,
+		IsSufficient: isSufficient,
+		Reason:       reason,
 	}, nil
 }
+
+func sumMemory(hwInfo models.Introspection) int64 {
+	var sum int64
+	for i := range hwInfo.Memory {
```
ronniel1, line 82: Make sure that this is only looking at the total memory and not free memory

internal/hardware/validator.go
filanov: Those are the minimal requirements that @avishayt write i don't think that the installation will work with less than this requirement. 
Eran and Igal told me that they have those requirements in the lab

internal/hardware/validator.go
filanov: The role can be empty as well. so i will add this logic in case the hw insufficient.

internal/hardware/validator.go
filanov: It is in the next line. or maybe i don't understand?

filanov: Only the total

filanov: Fixed only the total of "Mem"

PR #41, opened 2020-04-16
Update cluster state machine

avishayt:


avishayt:
As discussed in the HLD doc

PR #46, opened 2020-04-20
Added first version of assisted_installer

eranco74:


eranco74:
This docker image will be used for preforming the installation of the cluster nodes.
The bm-inventory will send the each node [the install command](https://github.com/filanov/bm-inventory/blob/master/internal/bminventory/inventory.go#L461)  with the relevant role (master, worker, bootstrap)
* the node with the bootstrap role will pivot to master once we have 2 ready master nodes.

tools/install_node.sh
```diff
@@ -0,0 +1,137 @@
...
+  echo "Starting progress.service"
+  nsenter_run systemctl start progress.service
+
+  echo "Done Setting up bootstrap"
+}
+
+install_bootstrap() {
+  bootstrap
+  get_kubeconfig
+  wait_for_nodes
```
ronniel1, line 41: Why don't we need to wait for the master nodes to be ready?

tools/install_node.sh
```diff
@@ -0,0 +1,137 @@
...
+  ?) help_func ;; # Print help_func in case parameter is non-existent
+  esac
+done
+
+# Begin script in case all parameters are correct
+case "$NODE_ROLE" in
+bootstrap) install_bootstrap
+           reboot_node ;;
+worker) install_node worker
+        reboot_node ;;
```
ronniel1, line 133: reboot_node can be done after the select, since in all cases we must do it

eranco74:
> We need to understand if we should change it to python
> On one hand, in python it will be much more readable, but on the other hand, the bootstrap is also bash.
> Considering that will need to pass it sometime in the review process of OpenShift, maybe they prefer bash?, Maybe it has less dependencies?
> 
> What do you think?
I think it should be python/go 
> Also lets think how to test this?
> Do we want/need UTs?
Definitely need UT and system test

tools/install_node.sh
eranco74: Because they won't be ready until we have etcd.
So I just wait for them to show up, than patch etcd (so it will start on 2 masters and won't wait for 3).
The install_node master is here as an optimization since it takes a while.

tools/install_node.sh
eranco74: In case of unknown option I don't want it to reboot

Makefile
```diff
@@ -2,6 +2,7 @@ PWD = $(shell pwd)
 UID = $(shell id -u)
 
 SERVICE := $(or ${SERVICE},quay.io/mfilanov/bm-inventory:latest)
+INSTALLER := $(or ${INSTALLER},quay.io/eranco74/assisted_installer:latest)
```
filanov, line 5: if it's required we need to add it to the readme as well.

eranco74: It's optional, @ronniel1  do we want another repo?

PR #59, opened 2020-04-23
MGMT-694 Download image should use ssh key from parameters instead of…

oshercc:
Download image should use ssh key from parameters instead of

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -84,14 +84,14 @@ const ignitionConfigFormat = `{
         "name": "core",
         "passwordHash": "$6$MWO4bibU8TIWG0XV$Hiuj40lWW7pHiwJmXA8MehuBhdxSswLgvGxEh8ByEzeX2D1dk87JILVUYS4JQOP45bxHRegAB9Fs/SWfszXa5."
       }
-	 %s
+	 {{.userSshKey}}
     ]
   },
 "systemd": {
 "units": [{
 "name": "agent.service",
 "enabled": true,
-"contents": "[Service]\nType=simple\nExecStartPre=docker run --privileged --rm -v /usr/local/bin:/hostbin %s cp /usr/bin/agent /hostbin\nExecStart=/usr/local/bin/agent --host %s --port %s --cluster-id %s\n\n[Install]\nWantedBy=multi-user.target"
+"contents": "[Service]\nType=simple\nExecStartPre=docker run --privileged --rm -v /usr/local/bin:/hostbin {{.AgentDockerImg}} cp /usr/bin/agent /hostbin\nExecStart=/usr/local/bin/agent --host {{.InventoryURL}} --port {{.InventoryPort}} --cluster-id {{.clusterId}}\n\n[Install]\nWantedBy=multi-user.target"
 }]
 }
 }`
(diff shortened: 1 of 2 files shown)


internal/bminventory/inventory.go
```diff
@@ -199,21 +199,21 @@ func (b *bareMetalInventory) createImageJob(cluster *models.Cluster, name string
 	}
 }
 
-func (b *bareMetalInventory) formatIgnitionFile(cluster *models.Cluster) string {
-	return fmt.Sprintf(ignitionConfigFormat, b.getUserSshKey(cluster), b.AgentDockerImg, b.InventoryURL,
+func (b *bareMetalInventory) formatIgnitionFile(cluster *models.Cluster, params inventory.DownloadClusterISOParams) string {
```
filanov, line 202: maybe you should replace the sprintf to `html/template`

internal/bminventory/inventory.go
```diff
@@ -199,21 +199,36 @@ func (b *bareMetalInventory) createImageJob(cluster *models.Cluster, name string
...
+	var ignitionParams = map[string]string{
+		"userSshKey":     b.getUserSshKey(cluster, params),
+		"AgentDockerImg": b.AgentDockerImg,
+		"InventoryURL":   b.InventoryURL,
+		"InventoryPort":  b.InventoryPort,
+		"clusterId":      cluster.ID.String(),
+	}
+	tmpl, err := template.New("ignitionConfig").Parse(ignitionConfigFormat)
+	if err != nil {
+		panic(err)
```
filanov, line 212: NEVER PANIC UNLESS IT'S IN MAIN.GO

internal/bminventory/inventory.go
```diff
@@ -199,21 +199,36 @@ func (b *bareMetalInventory) createImageJob(cluster *models.Cluster, name string
...
+	}
+	buf := &bytes.Buffer{}
+	if err := tmpl.Execute(buf, ignitionParams); err != nil {
+		panic(err)
```
filanov, line 216: NEVER PANIC UNLESS IT'S IN MAIN.GO

oshercc:
recheck

internal/bminventory/inventory.go
```diff
@@ -297,7 +312,12 @@ func (b *bareMetalInventory) DownloadClusterISO(ctx context.Context, params inve
 	}
 	// max job name is 63 chars
 	jobName := fmt.Sprintf("create-image-%s-%s", cluster.ID.String(), uuid.New().String())[:63]
-	if err := b.job.Create(ctx, b.createImageJob(&cluster, jobName)); err != nil {
+	ignitionConfig, formatErr := b.formatIgnitionFile(&cluster, params)
+	if formatErr != nil {
+		log.WithError(formatErr).Error("failed to format ignition config file")
```
filanov, line 317: please add cluster id to the log.

tsorya:
@filanov @avishayt if we are using ssh key from download params it means that there is no point to set key in cluster create.
Lets decide how it must work and what point is to set key in cluster params

filanov:
> @filanov @avishayt if we are using ssh key from download params it means that there is no point to set key in cluster create.
> Lets decide how it must work and what point is to set key in cluster params

There are two ssh keys, in cluster create the ssh key that will be used when operating the cluster.
The key on the image is for debugging the installation. it's a way to connect to the host before it's installed.

tsorya:
@filanov as for me if no ssh key is passed to download api , lets use cluster ssh key

filanov:
@oshercc this request pending input from the UX team please don't push

PR #60, opened 2020-04-23
MGMT-697: Update swagger to split download action

avishayt:


tsorya:
Creating image will be sync api? what will happened if we run download before create?

avishayt:
> Creating image will be sync api? what will happened if we run download before create?

Yes, it's a sync API.  It returns an image ID, and then you can download using that image ID.  If you try to download before, you won't have a valid image ID to pass to the API, and you will get 400 if you don't pass anything, or 404 if you make one up that doesn't exist.

avishayt:
@filanov Note that we should store the image with both the cluster ID and the image ID in the name, so that we know what cluster each image belongs to.  This will improve our ability to manage the images as well as security (so that user A can't download with an image ID belonging to user B).

swagger.yaml
```diff
@@ -132,6 +132,38 @@ paths:
...
+      responses:
+        201:
+          description: Created ISO
+          schema:
+            type: object
+            properties:
+              imageId:
+                type: string
+                format: uuid
+        400:
```
filanov, line 162: please add 404, in case cluster doesn't exists.

swagger.yaml
```diff
@@ -132,6 +132,38 @@ paths:
...
+          description: The ID of the cluster whose ISO to create
+          type: string
+          format: uuid
+          required: true
+        - in: body
+          name: new-iso-params
+          description: New ISO parameters
+          required: true
+          schema:
+            $ref: '#/definitions/image-create-params'
```
filanov, line 152: `image-create-params` container cluster id but we already have cluster id as part of the path. maybe the in `image-create-params` should be removed?

swagger.yaml
avishayt: :+1:

swagger.yaml
avishayt: :+1:

swagger.yaml
```diff
@@ -132,6 +132,40 @@ paths:
...
+        - in: path
+          name: clusterId
+          description: The ID of the cluster whose ISO to create
+          type: string
+          format: uuid
+          required: true
+        - in: body
+          name: image-create-params
+          description: New ISO parameters
+          required: true
```
filanov, line 150: @avishayt why is it required? all the parameters are optional no?

avishayt: It could theoretically be optional but I chose not to because:
1. I wasn't sure if all the client/server code generation would handle it properly
2. In case we later add a required parameter and forget to update here
3. To avoid mistakes where the user forgets to pass parameters

Do you think it's important?

filanov: It's OK, i asked because i had to create a pointer when testing it even if i didn't set anything inside. 
But it's fine, just wanted to know the reason.

swagger.yaml
```diff
@@ -148,20 +182,10 @@ paths:
...
-          type: integer
-          minimum: 0
-          maximum: 65535
-        - in: query
-          name: sshPublicKey
-          description: SSH public key for debugging the installation
+          name: imageId
+          description: The ID of a previously-created image
           type: string
+          format: uuid
```
filanov, line 188: as @tsorya mentions, image ID should be required right?

swagger.yaml
```diff
@@ -132,6 +132,40 @@ paths:
           description: Internal server error
 
   /clusters/{clusterId}/downloads/image:
+    post:
```
tsorya, line 135: @avishayt maybe downloads is not a good name to this api? maybe generate?

swagger.yaml
filanov: I don't know about swagger 2.0 but in openapi version 3 i looks like they can be required

> Query Parameters
Query parameters are the most common type of parameters. They appear at the end of the request URL after a question mark (?), with different name=value pairs separated by ampersands (&). Query parameters can be required and optional.

https://swagger.io/docs/specification/describing-parameters/

swagger.yaml
tsorya: it is strange to call download_cluster_iso api and no download starts

PR #63, opened 2020-04-26
MGMT-652 Implement cluster state machine

oshercc:


internal/cluster/common.go
```diff
@@ -0,0 +1,46 @@
...
+	clusterStatusError        = "error"
+)
+
+type UpdateReply struct {
+	State     string
+	IsChanged bool
+}
+
+type baseState struct {
+	log logrus.FieldLogger
```
oshercc, line 25: @filanov am getting a "db & log is unused" lint err but I am i am definitely using them, can you explain this ?

filanov: you probably not using them in some of the states.

internal/bminventory/inventory.go
```diff
@@ -241,7 +245,7 @@ func (b *bareMetalInventory) RegisterCluster(ctx context.Context, params invento
 		Status:    swag.String(ClusterStatusReady),
 		UpdatedAt: strfmt.DateTime{},
 	}
-
+	fmt.Println(b.clusterApi.RegisterCluster(ctx, &cluster, b.db))
```
ronniel1, line 248: Use logger instead

internal/cluster/ready.go
```diff
@@ -0,0 +1,65 @@
...
+type readyState baseState
+
+func (i *readyState) RegisterCluster(ctx context.Context, c *models.Cluster, db *gorm.DB) (*UpdateReply, error) {
+	i.log.Println("adasdasdas")
+	return nil, errors.Errorf("unable to register cluster <%s> in <%s> status",
+		c.ID, swag.StringValue(c.Status))
+}
+
+func (i *readyState) RefreshStatus(ctx context.Context, c *models.Cluster, db *gorm.DB) (*UpdateReply, error) {
+	var cluster models.Cluster
```
ronniel1, line 31: What is the difference in in this function and insufficientState.RefrechStatus()?

internal/cluster/installing.go
```diff
@@ -0,0 +1,38 @@
...
+	context "context"
+
+	"github.com/filanov/bm-inventory/models"
+	"github.com/go-openapi/swag"
+	"github.com/jinzhu/gorm"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+func NewInstallingState(log logrus.FieldLogger, db *gorm.DB) *installingState {
```
ronniel1, line 13: @filanov How do we move from 'installing' state to 'installed'?

internal/cluster/insufficient.go
```diff
@@ -0,0 +1,66 @@
...
+	for i := range cluster.Hosts {
+		if cluster.Hosts[i].Role == "master" {
+			masterNodesIds = append(masterNodesIds, cluster.Hosts[i].ID)
+		}
+	}
+	minimumMasterNodes := 3
+	if len(masterNodesIds) >= minimumMasterNodes {
+		return updateState(logutil.FromContext(ctx, i.log), clusterStatusReady, c, i.db)
+	} else {
+		return &UpdateReply{
```
ronniel1, line 56: I think we should log exactly why we are not sufficient yet

internal/cluster/statemachine.go
```diff
@@ -0,0 +1,80 @@
...
+func NewState(log logrus.FieldLogger, db *gorm.DB) *State {
+	return &State{
+		insufficient: NewInsufficientState(log, db),
+		ready:        NewReadyState(log, db),
+		installing:   NewInstallingState(log, db),
+		installed:    NewInstalledState(log, db),
+		error:        NewErrorState(log, db),
+	}
+}
+
```
ronniel1, line 40: I recommend  that on each file with code that should implement the API interface, you'll add a [compile-time assertion](https://medium.com/@odeke_et/compile-type-assertions-to-the-rescue-6ddab4b8398b) in the form of:
    var _ API  = State{}

What this does, is try to declare an un-named variable of type API, initialized with an empty State struct.
If State does not implement all the methods of the interface, this will cause a compilation error.

This also has the benefit of documenting which interfaces we are implementing.

internal/cluster/installing.go
filanov: We probably will have to create a cluster monitor that will wait for all the hosts to be installed or at least for 3 masters to be installed.

internal/cluster/insufficient.go
filanov: +1

internal/cluster/insufficient.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (i *insufficientState) RegisterCluster(ctx context.Context, c *models.Cluster, db *gorm.DB) (*UpdateReply, error) {
+	return nil, errors.Errorf("unable to register cluster <%s> in <%s> status",
+		c.ID, swag.StringValue(c.Status))
+}
+
+func (i *insufficientState) RefreshStatus(ctx context.Context, c *models.Cluster, db *gorm.DB) (*UpdateReply, error) {
+	var cluster models.Cluster
+
+	tx := db.Begin()
```
filanov, line 33: why do you need a transaction?

internal/cluster/insufficient.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	var masterNodesIds []*strfmt.UUID
+	for i := range cluster.Hosts {
+		if cluster.Hosts[i].Role == "master" {
+			masterNodesIds = append(masterNodesIds, cluster.Hosts[i].ID)
+		}
+	}
+	minimumMasterNodes := 3
+	if len(masterNodesIds) >= minimumMasterNodes {
+		return updateState(logutil.FromContext(ctx, i.log), clusterStatusReady, c, i.db)
```
filanov, line 54: you are mixing transaction and non traction db operations - not a good idea

internal/cluster/insufficient.go
```diff
@@ -0,0 +1,66 @@
...
+			tx.Rollback()
+		}
+	}()
+
+	if err := tx.Preload("Hosts").First(&cluster, "id = ?", c.ID).Error; err != nil {
+		return nil, errors.Errorf("cluster %s not found", c.ID)
+	}
+	var masterNodesIds []*strfmt.UUID
+	for i := range cluster.Hosts {
+		if cluster.Hosts[i].Role == "master" {
```
filanov, line 48: why not just use a counter?

internal/cluster/statemachine.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/filanov/bm-inventory/models"
+	"github.com/go-openapi/swag"
+	"github.com/jinzhu/gorm"
+	"github.com/sirupsen/logrus"
+)
+
+//go:generate mockgen -source=statemachine.go -package=host -destination=mock_cluster_api.go
+type API interface {
+	// Register a new cluster
+	RegisterCluster(ctx context.Context, c *models.Cluster, db *gorm.DB) (*UpdateReply, error)
```
filanov, line 16: are you sure you need to pass a DB on every function? install will probably need one but why register cluster?

internal/bminventory/inventory.go
```diff
@@ -431,13 +435,13 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
 			return inventory.NewInstallClusterConflict()
 		}
 	}
-
-	if err := tx.Model(&models.Cluster{}).Where("id = ?", cluster.ID.String()).
-		Update("status", ClusterStatusInstalling).Error; err != nil {
-		log.WithError(err).Errorf("failed to update cluster %s to status installing", cluster.ID.String())
+	if _, err := b.clusterApi.Install(ctx, &cluster, b.db); err != nil {
```
filanov, line 438: why do you pass the DB?
why `clusterApi`don't have a db?

internal/bminventory/inventory.go
```diff
@@ -431,13 +435,13 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
...
+		log.Errorf("failed to update cluster %s to installing state", params.ClusterID)
 		tx.Rollback()
 		return inventory.NewInstallClusterInternalServerError()
 	}
+
+	//TODO roleback cluster state if installation failed to commit
```
filanov, line 444: i think that all the logic of installing cluster should be inside clusterApi not only changing the state.

internal/bminventory/inventory.go
```diff
@@ -572,6 +576,13 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params inventory
 		return inventory.NewUpdateClusterInternalServerError()
 	}
 
+	//TODO how to deal with rolback case
+	if _, err := b.clusterApi.RefreshStatus(ctx, &cluster, b.db); err != nil {
```
filanov, line 580: again update cluster should be part of cluster api.

filanov, line 580: the order here is incorrect. why do you commit if still have things to do?

internal/cluster/statemachine.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/filanov/bm-inventory/models"
+	"github.com/go-openapi/swag"
+	"github.com/jinzhu/gorm"
+	"github.com/sirupsen/logrus"
+)
+
+//go:generate mockgen -source=statemachine.go -package=cluster -destination=mock_cluster_api.go
+type API interface {
+	// Register a new cluster
+	RegisterCluster(ctx context.Context, c *models.Cluster) (*UpdateReply, error)
```
filanov, line 16: i think that you need to split this api into sub interfaces.
`RegisterCluster` is a function that can be called only once. because we are the one that create the id user can't register the same cluster twice.
So i suggest to split the interface into functions that affect the state and functions that are not affecting the state.
there is no reason to implement `RegisterCluster` more then once.

internal/bminventory/inventory.go
```diff
@@ -35,10 +37,11 @@ const kubeconfigPrefix = "generate-kubeconfig"
 const defaultJobNamespace = "default"
 
 const (
-	ClusterStatusReady      = "ready"
-	ClusterStatusError      = "error"
-	ClusterStatusInstalling = "installing"
-	ClusterStatusInstalled  = "installed"
+	ClusterStatusInsufficient = "insufficient"
```
filanov, line 40: the state shouldn't be defined here. it's internal to cluster api.

internal/bminventory/inventory.go
```diff
@@ -252,12 +257,12 @@ func (b *bareMetalInventory) RegisterCluster(ctx context.Context, params invento
 		PullSecret:               params.NewClusterParams.PullSecret,
 		ServiceNetworkCIDR:       params.NewClusterParams.ServiceNetworkCIDR,
 		SSHPublicKey:             params.NewClusterParams.SSHPublicKey,
-		// TODO: should start as insufficient
-		Status:    swag.String(ClusterStatusReady),
-		UpdatedAt: strfmt.DateTime{},
+		Status:                   swag.String(ClusterStatusInsufficient),
```
filanov, line 260: status should be set inside cluster-api

internal/cluster/cluster_test.go
```diff
@@ -0,0 +1,91 @@
...
+})
+
+func prepareDB() *gorm.DB {
+	db, err := gorm.Open("sqlite3", ":memory:")
+	Expect(err).ShouldNot(HaveOccurred())
+	db.AutoMigrate(&models.Cluster{})
+	return db
+}
+
+func TestSubsystem(t *testing.T) {
```
filanov, line 76: it's not subsystem

internal/cluster/cluster_test.go
```diff
@@ -0,0 +1,91 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "host state machine tests")
```
filanov, line 78: it's not host state machine

internal/cluster/common.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	for _, host := range cluster.Hosts {
+		if host.Role == "master" {
+			masterNodesIds = append(masterNodesIds, host.ID)
+		}
+	}
+	return masterNodesIds, nil
+}
+
+func isClusterReady(c *models.Cluster, db *gorm.DB, log logrus.FieldLogger) (bool, error) {
```
filanov, line 64: I think that ready is when at least 3 masters are in known state.
@avishayt how do define ready state for a cluster? if we have 3 known masters but all the other hosts are insufficient/disconnected or any other not known state is the cluster still ready?

internal/cluster/common.go
```diff
@@ -0,0 +1,109 @@
...
+	tx := db.Begin()
+
+	defer func() {
+		if txErr != nil {
+			tx.Rollback()
+		}
+	}()
+
+	for j := range c.Hosts {
+		if txErr = tx.Where("id = ? and cluster_id = ?", c.Hosts[j], c.ID).Delete(&models.Host{}).Error; txErr != nil {
```
filanov, line 90: please use host api.

internal/cluster/common.go
```diff
@@ -0,0 +1,109 @@
...
+	if len(masterNodesIds) < minimumMasterNodes {
+		log.Infof("cluster %s has %d master hosts which is less then the %d minimum needed for cluster installation",
+			c.ID, len(masterNodesIds), minimumMasterNodes)
+		return false, nil
+	} else {
+		return true, nil
+	}
+}
+
+func deregisterCluster(c *models.Cluster, db *gorm.DB) (*UpdateReply, error) {
```
filanov, line 79: we can't delete cluster that is being installed right now.

internal/cluster/statemachine.go
```diff
@@ -0,0 +1,90 @@
...
+		return s.installing, nil
+	case clusterStatusInstalled:
+		return s.installed, nil
+	case clusterStatusError:
+		return s.error, nil
+	}
+	return nil, fmt.Errorf("not supported cluster status: %s", status)
+}
+
+func (s *State) RegisterCluster(ctx context.Context, c *models.Cluster) (*UpdateReply, error) {
```
filanov, line 60: Can't register cluster twice. 
we initiate cluster id during the registration so there is no state yet.
take a look at a comment from a previous review.

internal/cluster/common.go
```diff
@@ -0,0 +1,109 @@
...
+	IsChanged bool
+}
+
+type baseState struct {
+	//TODO remove when res: https://github.com/golangci/golangci-lint/issues/537
+	log logrus.FieldLogger //nolint:structcheck
+	db  *gorm.DB           //nolint:structcheck
+}
+
+func updateState(state string, c *models.Cluster, db *gorm.DB) (*UpdateReply, error) {
```
filanov, line 31: I think that you should log all the changes here. then you will not need to log them anywhere else

subsystem/cluster_test.go
```diff
@@ -123,6 +125,7 @@ var _ = Describe("system-test cluster install", func() {
 		})
 		Expect(err).NotTo(HaveOccurred())
 		cluster = registerClusterReply.GetPayload()
+		//Expect(swag.StringValue(cluster.Status)).Should(Equal("insufficient"))
```
filanov, line 128: delete if not needed, else fix it.

subsystem/cluster_test.go
```diff
@@ -149,6 +152,8 @@ var _ = Describe("system-test cluster install", func() {
 				CPU:    &models.CPU{Cpus: 16},
 				Memory: []*models.Memory{{Name: "Mem", Total: int64(32 * units.GiB)}},
 			}
+			//Expect(swag.StringValue(cluster.Status)).Should(Equal("insufficient"))
```
filanov, line 155: delete if not needed, else fix it.

subsystem/cluster_test.go
```diff
@@ -168,9 +172,11 @@ var _ = Describe("system-test cluster install", func() {
 				ClusterID: clusterID,
 			})
 			Expect(err).NotTo(HaveOccurred())
+			Expect(swag.StringValue(c.GetPayload().Status)).Should(Equal("ready"))
 		})
 
 		It("install cluster", func() {
+			fmt.Println(swag.StringValue(cluster.Status))
```
filanov, line 179: please delete.

internal/cluster/common.go
oshercc: ok, so?
where do you see me deleting an installing cluster?

internal/cluster/common.go
oshercc: hostAPI has no remove method

internal/cluster/statemachine.go
oshercc: I am not sure what you mean by "split into sub interfaces"

filanov: There is no point for all the interfaces to implement register cluster even if it's just return an error because this function will never be called right? 
so there is no point having single interface that enforce all the states to implement a function that will never be used.

oshercc: this is true for installCluster isn't don't you think that to start deconstructing the statementing will make it less "aesthetic"

filanov: i think it's less aesthetic to have dead code no?

filanov: @ronniel1 what do you think?

internal/cluster/common.go
avishayt: Right now we know how to install on 3 masters, so it's enough to be "ready"

filanov: ok but still the state should be "known" we have nothing to do with 3 disconnected masters.

avishayt: Yes of course.  That's what you wrote before too - "if we have 3 known masters but..."

filanov: then @oshercc please check the status of the masters.

internal/cluster/installing.go
ronniel1: Should we have a separate task on that?

filanov: https://issues.redhat.com/browse/MGMT-747

PR #64, opened 2020-04-26
replace error return with the correct one

oshercc:


internal/bminventory/inventory.go
```diff
@@ -830,7 +830,7 @@ func (b *bareMetalInventory) DownloadClusterFiles(ctx context.Context, params in
 		if gorm.IsRecordNotFoundError(err) {
 			return inventory.NewDownloadClusterFilesNotFound()
 		} else {
-			return inventory.NewDownloadClusterISOInternalServerError()
+			return inventory.NewDownloadClusterFilesInternalServerError()
```
filanov, line 833: you can check error type with `BeAssignableToTypeOf`
https://onsi.github.io/gomega/#beassignabletotypeofexpected-interface
so please validate it in a test.

oshercc: do we have a way to simulate corruption in the db ?

PR #68, opened 2020-04-27
swagger: Add statusInfo to cluster

avishayt:


swagger.yaml
```diff
@@ -841,6 +841,8 @@ definitions:
               - error
               - installing
               - installed
+          statusInfo:
```
filanov, line 844: the default string length in the db is 255 i don't think it's going to be enough

PR #70, opened 2020-04-30
WIP Set all deploy login in makefile without external hack files

filanov:


filanov:
@ronniel1 you can ignore this one for now.
i'm working on setting the deployment option to openshift and to minikube

PR #72, opened 2020-04-30
MGMT-705 Choose boot device for the installation, taking first one

tsorya:
Adding boot device selection.
For now it will take the first one of type disk with lowest name, to match sda disk

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"text/template"
 
+	"github.com/filanov/bm-inventory/internal/hardware"
 	"github.com/filanov/bm-inventory/internal/host"
 	"github.com/filanov/bm-inventory/internal/installcfg"
 	"github.com/filanov/bm-inventory/models"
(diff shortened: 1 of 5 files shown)


tsorya:
@avishayt this is done for now. We need it to go ahead with installation on baremetal.
When we will add api to choose disk, lets just change this function.

tsorya:
as the first step the idea was to return sda in baremetal and vda for vms.
Till now we have vda hardcoded

tsorya:
maybe something like this is better ? 
https://github.com/filanov/bm-inventory/pull/73

internal/bminventory/inventory.go
```diff
@@ -472,27 +473,33 @@ func (b *bareMetalInventory) addInstallCommand(ctx context.Context, masterNodesI
...
 		}
 		data["ROLE"] = role
-
+		disks, err := hardware.GetHostDisks(cluster.Hosts[i])
+		if err != nil {
+			return err
+		}
+		data["BOOT_DEVICE"] = fmt.Sprintf("/dev/%s", disks[0].Name)
 		buf := &bytes.Buffer{}
+		log.Debugf("Start executing install command on host %s with params %s", cluster.Hosts[i].HostID, data)
```
eranco74, line 502: Please remove this log.
We are not executing anything on the host.

internal/bminventory/inventory.go
```diff
@@ -472,27 +473,33 @@ func (b *bareMetalInventory) addInstallCommand(ctx context.Context, masterNodesI
...
 	for i := range cluster.Hosts {
 		role := cluster.Hosts[i].Role
 		if cluster.Hosts[i].ID == bootstrapId {
 			role = bootstrap
 		}
 		data["ROLE"] = role
-
+		disks, err := hardware.GetHostDisks(cluster.Hosts[i])
+		if err != nil {
+			return err
```
eranco74, line 498: Log an error for a specific host.

internal/hardware/hardware_utils.go
```diff
@@ -0,0 +1,35 @@
...
+		return nil, err
+	}
+	disks := getDisksList(hwInfo)
+	if len(disks) == 0 {
+		return nil, fmt.Errorf("host %s doesn't have disks", host.HostID)
+	}
+	return disks, nil
+}
+
+func getDisksList(hwInfo models.Introspection) []*models.BlockDevice {
```
eranco74, line 23: Just call it listDisks

avishayt:
> maybe something like this is better ?
> #73

Yes that is awesome - thanks!

tsorya:
@avishayt i will reopen it with changes

PR #73, opened 2020-05-01
Igal/general 2

tsorya:
Add disk validator and choose default disk

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"text/template"
 
+	"github.com/filanov/bm-inventory/internal/hardware"
 	"github.com/filanov/bm-inventory/internal/host"
 	"github.com/filanov/bm-inventory/internal/installcfg"
 	"github.com/filanov/bm-inventory/models"
@@ -110,10 +111,11 @@ type bareMetalInventory struct {
 	log           logrus.FieldLogger
 	job           job.API
 	hostApi       host.API
+	validator     hardware.Validator
 }
 
 func NewBareMetalInventory(db *gorm.DB, log logrus.FieldLogger, hostApi host.API, cfg Config,
-	jobApi job.API) *bareMetalInventory {
+	jobApi job.API, validator hardware.Validator) *bareMetalInventory {
 
 	b := &bareMetalInventory{
 		db:          db,
@@ -122,6 +124,7 @@ func NewBareMetalInventory(db *gorm.DB, log logrus.FieldLogger, hostApi host.API
 		debugCmdMap: make(map[strfmt.UUID]debugCmd),
 		hostApi:     hostApi,
 		job:         jobApi,
+		validator:   validator,
 	}
 	if cfg.ImageBuilderCmd != "" {
 		b.imageBuildCmd = strings.Split(cfg.ImageBuilderCmd, " ")
@@ -472,18 +475,20 @@ func (b *bareMetalInventory) addInstallCommand(ctx context.Context, masterNodesI
 	bootstrapId := masterNodesIds[len(masterNodesIds)-1]
 	log.Debugf("Bootstrap ID is %s", bootstrapId)
 
-	const cmdTmpl = `sudo podman run -v /dev:/dev:rw -v /opt:/opt:rw --privileged --pid=host  {{.INSTALLER}} --role {{.ROLE}}  --cluster-id {{.CLUSTER_ID}}  --host {{.HOST}} --port {{.PORT}} --boot-device /dev/vda`
+	const cmdTmpl = `sudo podman run -v /dev:/dev:rw -v /opt:/opt:rw --privileged --pid=host  {{.INSTALLER}} --role {{.ROLE}}  --cluster-id {{.CLUSTER_ID}}  --host {{.HOST}} --port {{.PORT}} --boot-device {{.BOOT_DEVICE}}`
+
 	t, err := template.New("cmd").Parse(cmdTmpl)
 	if err != nil {
 		return err
 	}
 
 	data := map[string]string{
-		"HOST":       b.InventoryURL,
-		"PORT":       b.InventoryPort,
-		"CLUSTER_ID": string(params.ClusterID),
-		"ROLE":       "",
-		"INSTALLER":  b.Config.InstallerImage,
+		"HOST":        b.InventoryURL,
+		"PORT":        b.InventoryPort,
+		"CLUSTER_ID":  string(params.ClusterID),
+		"ROLE":        "",
+		"INSTALLER":   b.Config.InstallerImage,
+		"BOOT_DEVICE": "",
 	}
 	for i := range cluster.Hosts {
 		role := cluster.Hosts[i].Role
@@ -491,7 +496,12 @@ func (b *bareMetalInventory) addInstallCommand(ctx context.Context, masterNodesI
 			role = bootstrap
 		}
 		data["ROLE"] = role
-
+		disks, err := b.validator.GetHostValidDisks(cluster.Hosts[i])
+		if err != nil {
+			log.Errorf("Failed to get valid disks on host with id %s", cluster.Hosts[i].ID)
+			return err
+		}
+		data["BOOT_DEVICE"] = fmt.Sprintf("/dev/%s", disks[0].Name)
 		buf := &bytes.Buffer{}
 		if err := t.Execute(buf, data); err != nil {
 			return err
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -57,7 +57,7 @@ var _ = Describe("GenerateClusterISO", func() {
 		ctrl = gomock.NewController(GinkgoT())
 		db = prepareDB()
 		mockJob = job.NewMockAPI(ctrl)
-		bm = NewBareMetalInventory(db, getTestLog(), nil, cfg, mockJob)
+		bm = NewBareMetalInventory(db, getTestLog(), nil, cfg, mockJob, nil)
 	})
 
 	registerCluster := func() *models.Cluster {
--- a/internal/hardware/validator.go
+++ b/internal/hardware/validator.go
@@ -3,6 +3,7 @@ package hardware
 import (
 	"encoding/json"
 	"fmt"
+	"sort"
 
 	"github.com/alecthomas/units"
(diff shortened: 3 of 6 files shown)


internal/hardware/hardware_utils_test.go_old
```diff
@@ -0,0 +1,82 @@
+package hardware
```
tsorya, line 1: not relevant file

internal/hardware/validator_test.go
```diff
@@ -35,6 +35,9 @@ var _ = Describe("hardware_validator", func() {
 		hwInfo := &models.Introspection{
 			CPU:    &models.CPU{Cpus: 16},
 			Memory: []*models.Memory{{Name: "Mem", Total: int64(32 * units.GiB)}},
+			BlockDevices: []*models.BlockDevice{
```
filanov, line 38: if you have the same code why to write it several times?

internal/bminventory/inventory.go
```diff
@@ -472,27 +475,33 @@ func (b *bareMetalInventory) addInstallCommand(ctx context.Context, masterNodesI
...
 		if cluster.Hosts[i].ID == bootstrapId {
 			role = bootstrap
 		}
 		data["ROLE"] = role
-
+		disks, err := b.validator.GetHostValidDisks(cluster.Hosts[i])
+		if err != nil {
+			return err
+		}
+		data["BOOT_DEVICE"] = fmt.Sprintf("/dev/%s", disks[0].Name)
```
filanov, line 502: Is it always disk 0? 
In real HW it can be a different disk no?

internal/hardware/validator.go
```diff
@@ -95,3 +114,20 @@ func getTotalMemory(hwInfo models.Introspection) int64 {
 func gibToBytes(gib int64) int64 {
 	return gib * int64(units.GiB)
 }
+
+func getDisksList(hwInfo models.Introspection, minSizeRequired int64) []*models.BlockDevice {
+	var disks []*models.BlockDevice
+	for _, blockDevice := range hwInfo.BlockDevices {
+		// TODO add size validation
```
filanov, line 121: why not to do the validation?

internal/hardware/validator.go
```diff
@@ -95,3 +114,20 @@ func getTotalMemory(hwInfo models.Introspection) int64 {
...
+		if blockDevice.DeviceType == "disk" && blockDevice.RemovableDevice == 0 &&
+			!blockDevice.ReadOnly && blockDevice.Size >= minSizeRequired {
+
+			disks = append(disks, blockDevice)
+		}
+	}
+	// Sorting list by name
+	sort.Slice(disks, func(i, j int) bool {
+		return disks[i].Name < disks[j].Name
```
filanov, line 130: why sort by name and not by size?

internal/bminventory/inventory_test.go
```diff
@@ -44,20 +46,22 @@ func getTestLog() logrus.FieldLogger {
...
 	)
 
 	BeforeEach(func() {
 		Expect(envconfig.Process("test", &cfg)).ShouldNot(HaveOccurred())
 		ctrl = gomock.NewController(GinkgoT())
 		db = prepareDB()
 		mockJob = job.NewMockAPI(ctrl)
-		bm = NewBareMetalInventory(db, getTestLog(), nil, cfg, mockJob)
+		mockValidator = hardware.NewMockValidator(ctrl)
+		bm = NewBareMetalInventory(db, getTestLog(), nil, cfg, mockJob, mockValidator)
```
filanov, line 64: it's no used here so why to create the mock? just use `nil`

internal/bminventory/inventory.go
```diff
@@ -472,27 +475,33 @@ func (b *bareMetalInventory) addInstallCommand(ctx context.Context, masterNodesI
...
+		"BOOT_DEVICE": "",
 	}
 	for i := range cluster.Hosts {
 		role := cluster.Hosts[i].Role
 		if cluster.Hosts[i].ID == bootstrapId {
 			role = bootstrap
 		}
 		data["ROLE"] = role
-
+		disks, err := b.validator.GetHostValidDisks(cluster.Hosts[i])
```
filanov, line 498: i think it should be part of host api

internal/bminventory/inventory.go
```diff
@@ -472,27 +475,33 @@ func (b *bareMetalInventory) addInstallCommand(ctx context.Context, masterNodesI
...
+		if err != nil {
+			return err
+		}
+		data["BOOT_DEVICE"] = fmt.Sprintf("/dev/%s", disks[0].Name)
 		buf := &bytes.Buffer{}
+		log.Debugf("Start executing install command on host %s with params %s", cluster.Hosts[i].HostID, data)
```
filanov, line 504: why not info?

internal/bminventory/inventory.go
tsorya: The idea is that by default u will get "sda" or "vda", so u will get list of valid disks sorted by name and sda will be the first

internal/bminventory/inventory.go
tsorya: hmm was not sure it is ok to add it as info

internal/bminventory/inventory_test.go
tsorya: hmm, right

internal/hardware/validator.go
tsorya: added. need to delete comment

internal/hardware/validator.go
tsorya: as wrote previously as u know by default sda(the first added volume) is boot volume, so i am taking the first disk that matches.

internal/bminventory/inventory.go
filanov: ok

internal/bminventory/inventory.go
filanov: we don't run install every second so why not add some more info?

internal/bminventory/inventory.go
filanov: @ronniel1 what do you think?

internal/bminventory/inventory.go
```diff
@@ -472,27 +475,35 @@ func (b *bareMetalInventory) addInstallCommand(ctx context.Context, masterNodesI
...
 		data["ROLE"] = role
-
+		disks, err := b.validator.GetHostValidDisks(cluster.Hosts[i])
+		if err != nil {
+			log.Errorf("Failed to get valid disks on host with id %s", cluster.Hosts[i].ID)
+			return err
+		}
+		data["BOOT_DEVICE"] = fmt.Sprintf("/dev/%s", disks[0].Name)
 		buf := &bytes.Buffer{}
+		log.Infof("Start installation on host %s with params %s", cluster.Hosts[i].HostID, data)
```
eranco74, line 506: Remove the log (the execute here just populate the template).
The command get logged when we add the debug command.

internal/hardware/validator.go
```diff
@@ -29,6 +31,7 @@ type ValidatorCfg struct {
 	MinRamGib         int64 `envconfig:"HW_VALIDATOR_MIN_RAM_GIB" default:"8"`
 	MinRamGibWorker   int64 `envconfig:"HW_VALIDATOR_MIN_RAM_GIB_WORKER" default:"8"`
 	MinRamGibMaster   int64 `envconfig:"HW_VALIDATOR_MIN_RAM_GIB_MASTER" default:"16"`
+	MinDiskSizeGib    int64 `envconfig:"HW_VALIDATOR_MIN_DISK_SIZE_GIB" default:"120"`
```
eranco74, line 34: We will fail to install on test-infra (disk is 20 GB).
@tsorya don't forget to add the an override,

internal/hardware/validator_test.go
```diff
@@ -77,6 +77,41 @@ var _ = Describe("hardware_validator", func() {
 		sufficient(hwvalidator.IsSufficient(host))
 	})
 
+	It("insufficient_number_of_valid_disks", func() {
+		hwInfo.BlockDevices = []*models.BlockDevice{
+			// Not disk type
+			{DeviceType: "loop", Fstype: "squashfs", MajorDeviceNumber: 7, MinorDeviceNumber: 0, Mountpoint: "/sysroot", Name: "loop0", ReadOnly: true, RemovableDevice: 1, Size: 746217472},
+			// Not enough size
+			{DeviceType: "disk", Fstype: "iso9660", MajorDeviceNumber: 11, Mountpoint: "/test", Name: "sdb", Size: 118849018880},
```
eranco74, line 85: Please give name to the disks sizes.

internal/bminventory/inventory.go
tsorya: But @filanov  sai to leave it and to make it info

internal/hardware/validator.go
tsorya: sure :)

internal/bminventory/inventory.go
filanov: the command was to switch it from debug to info. if this log is duplication then you can remove it.

internal/hardware/validator_test.go
filanov: need to make sure that the info that you get from the agent is really bytes.

tsorya: i checked it, till now it was bytes. Maybe it will be changed while moving to IRONIC model

tsorya:
@filanov i don't see any validator test in subsystem
I have added all relevant test cases to validator unit-tests.

filanov:
> @filanov i don't see any validator test in subsystem
> I have added all relevant test cases to validator unit-tests.

you didn't checked erros from unmarshal (invalid hw info)
and you didn't checked empty disk list :)

tsorya:
i checked it in validator :) 
There no test for addInstallCommand :) u think it is fair to ask to add one for such a small change :)

tsorya:
@filanov  u are right will add it to validator test

internal/bminventory/inventory.go
ronniel1: It's not part of the state machine, but it should be part of the host.
The problem is that now the state machine must implement all the Host API.
This is indeed API of the host IMO, but not part of the state machine

filanov: you are right. currently i'm working above Liat's branch that split host api into state machine api and other host apis, for example get next step. he can put it there.

PR #74, opened 2020-05-03
Michaelf/deploy

filanov:


filanov:
> I think that you are abusing Makefile here.
> Scripts should be in external files, and just invoked from here.
> 
> We can discuss this f2f if you prefer

I wrote everything in the makefile because it's easier to see everything in one place. 
For me it was really hard to understand makefile that use only scripts.
But because i'm not sure what is the right way to do it i can change it scripts.
Do you think we should make it python?

PR #78, opened 2020-05-04
Use the proxy parameters when generation the ISO

eranco74:
It's not easy validating it in UT.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 	"text/template"
@@ -202,8 +203,8 @@ func (b *bareMetalInventory) formatIgnitionFile(cluster *models.Cluster, params 
 	var ignitionParams = map[string]string{
 		"userSshKey":     b.getUserSshKey(params),
 		"AgentDockerImg": b.AgentDockerImg,
-		"InventoryURL":   b.InventoryURL,
-		"InventoryPort":  b.InventoryPort,
+		"InventoryURL":   b.getURLForIngnition(params),
+		"InventoryPort":  b.getPortForIgnition(params),
 		"clusterId":      cluster.ID.String(),
 	}
 	tmpl, err := template.New("ignitionConfig").Parse(ignitionConfigFormat)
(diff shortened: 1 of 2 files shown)


internal/bminventory/inventory.go
```diff
@@ -231,6 +232,20 @@ func (b *bareMetalInventory) getUserSshKey(params inventory.GenerateClusterISOPa
 		"groups": [ "sudo" ]}`, sshKey)
 }
 
+func (b *bareMetalInventory) getTargetUrl(params inventory.GenerateClusterISOParams) string {
```
filanov, line 235: not sure about the name. 
maybe `getURLForIngnission`?

internal/bminventory/inventory.go
```diff
@@ -231,6 +232,20 @@ func (b *bareMetalInventory) getUserSshKey(params inventory.GenerateClusterISOPa
 		"groups": [ "sudo" ]}`, sshKey)
 }
 
+func (b *bareMetalInventory) getURLForIngnition(params inventory.GenerateClusterISOParams) string {
+	if params.ImageCreateParams.ProxyIP != "" {
```
ronniel1, line 236: No need for this. Just return swag.String(ProxyIP)

internal/bminventory/inventory.go
```diff
@@ -231,6 +232,20 @@ func (b *bareMetalInventory) getUserSshKey(params inventory.GenerateClusterISOPa
...
+		return string(params.ImageCreateParams.ProxyIP)
+	}
+	return b.InventoryURL
+}
+
+func (b *bareMetalInventory) getPortForIgnition(params inventory.GenerateClusterISOParams) string {
+	if params.ImageCreateParams.ProxyPort != nil {
```
ronniel1, line 243: same as above

internal/bminventory/inventory.go
filanov: he can't because the default if tit's not exists is `b.InventoryURL`

internal/bminventory/inventory.go
```diff
@@ -231,6 +232,20 @@ func (b *bareMetalInventory) getUserSshKey(params inventory.GenerateClusterISOPa
...
+		return strconv.FormatInt(*params.ImageCreateParams.ProxyPort, 10)
```
filanov, line 244: you can use `swag.IntValue`

filanov: or `fmt.Sprintf`

ronniel1: strconv is much better performance than fmt
However swag.IntValue is the good choice here, I think

internal/bminventory/inventory.go
ronniel1: Than maybe a single function:
   StringValue(str *string, default string) return string {
       v := swag.String(str)
       if v == "" {
           return default
       } else  {
           return v
       }
    }

eranco74: @ronniel1 I want it to be clear.

internal/bminventory/inventory.go
eranco74: I want a string
swag.IntValue will give me an int.
In addition I will get 0 in case of nil pointer.

filanov: good point

PR #81, opened 2020-05-04
Michaelf/progress api

filanov:


swagger.yaml
```diff
@@ -1022,3 +1051,6 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/connectivity-remote-host'
+
+  host-install-progress-params:
```
ronniel1, line 1055: Shouldn't this be an enum?

filanov: At first i was thinking the same but because the install script that @eranco74 is writing is independent from the service and it can add or remove steps without changing anything in the service, so this way it will be more flexible. 
@avishayt is the one that helped me define the api and he suggested it to be a string.
The only "magic" is the `done` progress string that will turn the host from "installing" to installed

swagger.yaml
```diff
@@ -366,6 +366,35 @@ paths:
         500:
           description: Internal server error
 
+  /clusters/{clusterId}/hosts/{hostId}/progress:
+    patch:
```
avishayt, line 370: We're overwriting it, so "put" is more appropriate

swagger.yaml
avishayt: :+1:

avishayt: The :+1: was meant for Ronnie's comment :)
Enum is also a string... it would be good to know what the values can be.
Also, is there any way to get some numerical progress?  (e.g., "5/7: installing foo")

filanov: @avishayt but the one that define the number of steps is a different service/tool if we set an enum then every change in the tool will require changes in the API no?

filanov: @eranco74 can you send me the list of steps so we can make it an enum?

avishayt: I think it should be pretty stable, and it's good to document the API.

swagger.yaml
filanov: fixed.

swagger.yaml
eranco74: @avishayt we don't know what it's going to be how can it be stable?
is it:
1: 12%, 33%, 98%. 
2: pulling ignition, pulling image, writing image to disk, restart ...
Different nodes will have different steps (according to the role).
Do we plan to add logic around this progress?

eranco74: I think we should start with free text and have a const for "Installed"/"Done".
So we can have some logic for knowing the installation is done.

avishayt: OK, leave it as is for now, we can add it later if it will make sense

PR #88, opened 2020-05-04
Added code generation verification

ronniel1:


.github/workflows/unit-test.yml
```diff
@@ -71,6 +71,12 @@ jobs:
         mkdir -p ~/.docker
         echo "{}" > ~/.docker/config.json
         touch ${HOME}/.gitconfig
+    - name: verify-code-generation-clean
+      run: |
+        git status -s
```
filanov, line 76: why do you have it before and after skipper make generate?

filanov: ahh if it's for the logs then it's find.

PR #89, opened 2020-05-04
Ronnie/codecov3

ronniel1:


.github/workflows/unit-test.yml
```diff
@@ -6,6 +6,7 @@ on:
       - master
   pull_request:
     branches:
+      - ronnie/coverage
```
filanov, line 9: it should be only master right?

PR #92, opened 2020-05-04
MGMT-705 Choose boot device for the installation, taking first one

tsorya:
This reverts commit 4149e2c731cbf602a2551d2d8c69efa6f9574d3e.

### DIFF
--- a/subsystem/cluster_test.go
+++ b/subsystem/cluster_test.go
@@ -98,8 +98,9 @@ var _ = Describe("Cluster tests", func() {
 
 var _ = Describe("system-test cluster install", func() {
 	var (
-		ctx     = context.Background()
-		cluster *models.Cluster
+		ctx           = context.Background()
+		cluster       *models.Cluster
+		validDiskSize = int64(128849018880)
 	)
 
 	AfterEach(func() {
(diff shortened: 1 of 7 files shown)


subsystem/cluster_test.go
```diff
@@ -98,8 +98,9 @@ var _ = Describe("Cluster tests", func() {
 
 var _ = Describe("system-test cluster install", func() {
 	var (
-		ctx     = context.Background()
-		cluster *models.Cluster
+		ctx           = context.Background()
+		cluster       *models.Cluster
+		validDiskSize = int64(128849018880)
```
filanov, line 103: how about adding a test with invalid disk size?

tsorya: how about not to do it? There is no point for it. This is the same as unitest no difference at all

filanov: it's not the same because here you test a full flow. 
you validate that all components are really used

tsorya: u just create same class and run same function, there is no difference. Current test case validates that validator insufficient is running. All other things are validated in unitest

filanov: no it's not, because it i create a dummy validation then the unit test will work the same but the system will not validate anything

PR #96, opened 2020-05-06
Michaelf/proxy

filanov:


swagger.yaml
```diff
@@ -569,14 +569,9 @@ definitions:
...
-      proxyIp:
+      proxyURL:
         type: string
-        description: The IP address of the HTTP proxy that agents should use to access the discovery service
-      proxyPort:
-        type: integer
-        description: The port of the HTTP proxy
-        minimum: 0
-        maximum: 65535
+        description: The URL of the HTTP/S proxy that agents should use to access the discovery service
```
avishayt, line 574: Please add the expected format.  Is it this?
"http://<user>:<password>@<ipaddr>:<port>"

filanov: I'm not sure how to define the pattern
do we enforce port or not?

avishayt: No need for a pattern right now - just put it in the description

filanov:
@tsorya wait for an ack from you before i push

internal/bminventory/inventory.go
```diff
@@ -87,7 +86,7 @@ const ignitionConfigFormat = `{
 "units": [{
 "name": "agent.service",
 "enabled": true,
-"contents": "[Service]\nType=simple\nExecStartPre=docker run --privileged --rm -v /usr/local/bin:/hostbin {{.AgentDockerImg}} cp /usr/bin/agent /hostbin\nExecStart=/usr/local/bin/agent --host {{.InventoryURL}} --port {{.InventoryPort}} --cluster-id {{.clusterId}}\n\n[Install]\nWantedBy=multi-user.target"
+"contents": "[Service]\nType=simple\nEnvironment=HTTPS_PROXY={{.httpsProxyURL}}\nEnvironment=HTTP_PROXY={{.httpProxyURL}}\nEnvironment=http_proxy=http://{{.httpProxyURL}}\nEnvironment=https_proxy={{.httpsProxyURL}}\nExecStartPre=docker run --privileged --rm -v /usr/local/bin:/hostbin {{.AgentDockerImg}} cp /usr/bin/agent /hostbin\nExecStart=/usr/local/bin/agent --host {{.InventoryURL}} --port {{.InventoryPort}} --cluster-id {{.clusterId}}\n\n[Install]\nWantedBy=multi-user.target"
```
romfreiman, line 89: I think NO_PROXY is required as well 
https://unix.stackexchange.com/questions/23452/set-a-network-range-in-the-no-proxy-environment-variable

romfreiman:
I think u will have to propagate it to the install-config as well (i opened a ticket for that). anyhow, it's out of the scope of this pull request

internal/bminventory/inventory.go
filanov: how do we know what is the mask ?

internal/bminventory/inventory.go
```diff
@@ -201,9 +200,11 @@ func (b *bareMetalInventory) formatIgnitionFile(cluster *models.Cluster, params
 	var ignitionParams = map[string]string{
 		"userSshKey":     b.getUserSshKey(params),
 		"AgentDockerImg": b.AgentDockerImg,
-		"InventoryURL":   b.getURLForIngnition(params),
-		"InventoryPort":  b.getPortForIgnition(params),
+		"InventoryURL":   b.InventoryURL,
```
romfreiman, line 203: is it logged somewhere? this is what we talked about this morning

filanov: Lets do it in a different ticket.

internal/bminventory/inventory.go
romfreiman: I think we can mask out the nodes ips. and localhost. @ori-amizur what did we do @ strato?
anyhow, lets not handle it now since u dont set it to the whole os but only for that process so it's not a real issue

filanov:
> I think u will have to propagate it to the install-config as well (i opened a ticket for that). anyhow, it's out of the scope of this pull request

it's a strange UX. because image create parameters are not stored in the DB.
The task should be adding proxy configuration to cluster.

internal/bminventory/inventory.go
romfreiman: why do we need a ticket to write logs? What am I missing? Just write logs :)

romfreiman:
> > I think u will have to propagate it to the install-config as well (i opened a ticket for that). anyhow, it's out of the scope of this pull request
> 
> it's a strange UX. because image create parameters are not stored in the DB.
> The task should be adding proxy configuration to cluster.

So we should. Why do we need to places to ask the user to fill same data?

filanov:
> > > I think u will have to propagate it to the install-config as well (i opened a ticket for that). anyhow, it's out of the scope of this pull request
> > 
> > 
> > it's a strange UX. because image create parameters are not stored in the DB.
> > The task should be adding proxy configuration to cluster.
> 
> So we should. Why do we need to places to ask the user to fill same data?

@avishayt should it be cluster configuration instead of just image configuration?

avishayt:
> > > I think u will have to propagate it to the install-config as well (i opened a ticket for that). anyhow, it's out of the scope of this pull request
> > 
> > 
> > it's a strange UX. because image create parameters are not stored in the DB.
> > The task should be adding proxy configuration to cluster.
> 
> So we should. Why do we need to places to ask the user to fill same data?

I didn't realize we need it for the install-config... so we need to get the proxy immediately upon cluster creation, and it is immutable, right?

filanov:
> > > > I think u will have to propagate it to the install-config as well (i opened a ticket for that). anyhow, it's out of the scope of this pull request
> > > 
> > > 
> > > it's a strange UX. because image create parameters are not stored in the DB.
> > > The task should be adding proxy configuration to cluster.
> > 
> > 
> > So we should. Why do we need to places to ask the user to fill same data?
> 
> I didn't realize we need it for the install-config... so we need to get the proxy immediately upon cluster creation, and it is immutable, right?

is it possible that single cluster will have more then one proxy? for example host in different locations etc.

filanov:
@romfreiman @avishayt when i think about it, why do we need it for the installation as well?
install command will have the same configuration no?
The cluster will not need the proxy to access the inventory service
so why do we need to store it?

romfreiman:
u need progress reporting and telemetry that are running from to cluster to
connect to the internet, no? how would it happen without a proxy? the user
just told us that it requires proxy to access the internet, didn't he? :)
see the link I posted in the ticket. we can ask on one of the slack
channels as well - I might be wrong of course.

filanov:
The proxy can be only for the installation, not all clusters are registered to redhat.com right? 
But maybe it's not our case. 
If we only have one install config then we can't support multiple proxy configurations and maybe it's not even relevant.
Anyway so we are starting with that user need to configure proxy when creating the cluster and he can't change that?

internal/bminventory/inventory.go
```diff
@@ -201,9 +200,11 @@ func (b *bareMetalInventory) formatIgnitionFile(cluster *models.Cluster, params
...
+		"InventoryPort":  b.InventoryPort,
 		"clusterId":      cluster.ID.String(),
+		"httpProxyURL":   b.setProxyPrefix(params.ImageCreateParams.ProxyURL, "http://"),
+		"httpsProxyURL":  b.setProxyPrefix(params.ImageCreateParams.ProxyURL, "https://"),
```
ori-amizur, line 207: This is usually wrong.  You need same values for HTTP_PROXY and HTTPS proxy.  In case of HTTPS_PROXY the client will do CONNECT method to the destination which will cause the PROXY to open a TCP connection.

filanov: fixed.

eranco74:
LGTM
This should allow us to register.
As Rom said this might not be enough for the full installation since the nodes need to pull images from the internet (after the reboot).
Yet I think it should be in another PR. first let's allow the nodes to register and start the install process.

internal/bminventory/inventory.go
filanov: done

PR #99, opened 2020-05-07
Swagger: change according to app-sre guidelines

avishayt:
- Removed tags from operations.
- Aligned descriptions to current api.openshift.
- Names of URL segments and JSON attributes should be all lower case and
  use snake_case.
- Names of kinds of objects should use CamelCase.
- The representation of relationships between resources should be done
  using links instead of embedding them.
- Format of returned HTTP errors according to doc.
- Catalog of errors.
- Descriptions for all properties, also end with period. Align
  descriptions with current APIs.

Still to be done: error catalog, paging, list filters and search,
sync/async.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -1,13 +1,13 @@
 swagger: '2.0'
 info:
-  description: 'Bare metal inventory'
+  description: 'Assisted installation'
   version: 1.0.0
-  title: BMInventory
+  title: AssistedInstall
 host: api.openshift.com
-basePath: /api/bm-inventory/v1
+basePath: /api/assisted-install/v1
 tags:
-  - name: Bare metal inventory
-    description: Manage bare metal inventory
+  - name: Assisted installation
+    description: Assisted bare metal installation
 
 schemes:
   - http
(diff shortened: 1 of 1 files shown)


igoihman:
@tzvatot PTAL

filanov:
@avishayt 
```
The swagger spec at "/home/mfilanov/work/go/src/github.com/filanov/bm-inventory/swagger.yaml" is invalid against swagger specification 2.0. see errors :
- some references could not be resolved in spec. First found: object has no key "Error"
```

filanov:
if you want to make sure that generation pass you can run `skipper make generate-from-swagger`
it will work without skipper as well. but you will need to have docker.

avishayt:
> @avishayt generation don't pass
> 
> ```
> The swagger spec at "/home/mfilanov/work/go/src/github.com/filanov/bm-inventory/swagger.yaml" is invalid against swagger specification 2.0. see errors :
> - some references could not be resolved in spec. First found: object has no key "Error"
> ```

Sorry about that, should work now

filanov:
@eranco74 @ori-amizur @igoihman this change will affect you as well. 
clients will change. so you will need to update your code as well.

swagger.yaml
```diff
@@ -1,13 +1,13 @@
 swagger: '2.0'
 info:
-  description: 'Bare metal inventory'
+  description: 'Assisted installation'
   version: 1.0.0
-  title: BMInventory
+  title: AssistedInstall
 host: api.openshift.com
-basePath: /api/bm-inventory/v1
+basePath: /api/assisted-install/v1
```
igoihman, line 7: assisted_install

PR #101, opened 2020-05-09
Osher/install cluster cluster api

oshercc:


internal/cluster/installer.go
```diff
@@ -0,0 +1,60 @@
...
+		return errors.Errorf("cluster %s has a error", c.ID)
+	default:
+		return errors.Errorf("cluster %s state is unclear - cluster state: %s", c.ID, swag.StringValue(c.Status))
+	}
+
+	clusterState, err := updateState(clusterStatusInstalling, c, db, i.log)
+	if err != nil {
+		return err
+	}
+	if clusterState.State != clusterStatusInstalling || !clusterState.IsChanged {
```
filanov, line 52: why do you need this part? why not just to return `err`?

internal/bminventory/inventory.go
```diff
@@ -374,41 +375,41 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
...
 				cluster.Hosts[i].ID.String(), cluster.ID.String())
 			tx.Rollback()
 			return inventory.NewInstallClusterConflict()
 		}
 	}
 
+	// create install-config.yaml
+	cfg, err := installcfg.GetInstallConfig(&cluster)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get install config for cluster %s", params.ClusterID)
```
filanov, line 403: rallback

internal/bminventory/inventory.go
```diff
@@ -368,54 +368,49 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
...
 		tx.Rollback()
-		return err
+		return inventory.NewInstallClusterConflict()
 	}
 
 	// Temporary hack - use debug API for setting the executing install command:
-	if err := b.addInstallCommand(ctx, masterNodesIds, log, params, cluster); err != nil {
+	masterNodesIds, err := b.clusterApi.GetMasterNodesIds(ctx, &cluster, tx)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get cluster %s master node id's", cluster.ID)
```
filanov, line 381: rollback

filanov:
what about inventory tests?

internal/bminventory/inventory.go
```diff
@@ -396,36 +396,28 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
...
-
 	if err := b.db.Preload("Hosts").First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
 		return inventory.NewInstallClusterInternalServerError()
 	}
-
 	return inventory.NewInstallClusterOK().WithPayload(&cluster)
 }
 
-func generateClusterInstallConfig(b *bareMetalInventory, cluster models.Cluster, log logrus.FieldLogger, ctx context.Context, cfg []byte) middleware.Responder {
+func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, cluster models.Cluster, log logrus.FieldLogger) middleware.Responder {
```
filanov, line 414: you have logger inside the context so no need to pass both. 
it's strange that once you are using `b.log` and the other time you are using `log` 
lets get the logger from context and use it.

internal/bminventory/inventory_test.go
```diff
@@ -6,6 +6,9 @@ import (
 	"io/ioutil"
 	"testing"
 
+	errHand "github.com/pkg/errors"
```
filanov, line 9: just use errors why do you need to give it names? 
or use `fmt.Errorf`

internal/bminventory/inventory_test.go
```diff
@@ -262,3 +265,152 @@ var _ = Describe("UpdateHostInstallProgress", func() {
...
+			Expect(err).ShouldNot(HaveOccurred())
+
+			for i := 0; i < 3; i++ {
+				addHost("master", "known", clusterID, db)
+			}
+		})
+
+		It("success", func() {
+
+			mockClusterApi.EXPECT().Install(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
```
filanov, line 329: if you have the same mocks i suggest to put them under the same context and write them once, otherwise if we change the api there is a lot of search and replace to do.
If it will make the code not clear then create  a function to set the mock this way you will have only once place to change.

internal/bminventory/inventory.go
```diff
@@ -411,7 +411,7 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params inventor
 	return inventory.NewInstallClusterOK().WithPayload(&cluster)
 }
 
-func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, cluster models.Cluster, log logrus.FieldLogger) middleware.Responder {
+func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, cluster models.Cluster) middleware.Responder {
```
filanov, line 414: you didn't extract the log from the context and you forgot line 418

internal/bminventory/inventory.go
```diff
@@ -336,7 +336,7 @@ func (b *bareMetalInventory) GenerateClusterISO(ctx context.Context, params inve
 	}
 
 	if err := b.job.Monitor(ctx, jobName, defaultJobNamespace); err != nil {
-		log.WithError(err).Error("image creation failed")
+		b.log.WithError(err).Error("image creation failed")
```
filanov, line 339: use log not b.log

PR #107, opened 2020-05-10
MGMT-765 Passing openshift version to assisted-installer

tsorya:


filanov:
@eranco74 you added a todo to support multiple openshift versions when you set `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` will there be a conflict with this PR?

eranco74:
> @eranco74 you added a todo to support multiple openshift versions when you set `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` will there be a conflict with this PR?

No conflict with this PR.
It just that currently we don't do anything with the openshift version we get from the user.
Passing it to the install command is 1 thing we should do.
In addition we will need to have the find the right OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE and the right installer image for the openshift versions we will support. 
For now we just install v4.4

filanov:
As long as nothing breaks i don't mind.
LGTM 
@tsorya @gamli75 you should sync because Liat is working on moving this code to the install command.

tsorya:
@filanov @eranco74 checked the full flow, everything looks ok

PR #110, opened 2020-05-12
MGMT-688 - Ironic output compatibility

ori-amizur:


swagger.yaml
```diff
@@ -1159,3 +1160,128 @@ definitions:
       reason:
         type: string
         description: Human readable description of the error.
+
+
+  interface:
```
avishayt, line 1165: Can we get the NIC's speed?  E.g., 1/10/40/100 GbE?

filanov:
@ori-amizur it looks good, will you wait for me to push my PR?

swagger.yaml
ori-amizur: Yes

PR #115, opened 2020-05-14
Ori/mgmt 688

ori-amizur:
Now above the master


filanov:
@ori-amizur so we will support only the new format?

PR #117, opened 2020-05-16
MGMT-845 move cluster id generating to clusterAPI

oshercc:


internal/cluster/registrar.go
```diff
@@ -83,3 +94,7 @@ func (r *registrar) DeregisterCluster(ctx context.Context, cluster *models.Clust
 	}
 	return nil
 }
+
+func buildHref(base, id string) *string {
```
ronniel1, line 98: If this is required in order top get the cluster's URL you can use:
    x := inventory.GetClusterURL{
	ClusterID: "1234556",
    }
    return .String()

filanov:
I think it looks strange that we give cluster API a cluster object and it need to generate the id.
it will look more natural if cluster api will get the parameters and return cluster. 
@ronniel1 @oshercc what do you think?

oshercc:
> I think it looks strange that we give cluster API a cluster object and it need to generate the id.
> it will look more natural if cluster api will get the parameters and return cluster.
> @ronniel1 @oshercc what do you think?

@filanov this change opens a door to a lot of changes and is not the main point of this task, 
I opened a additional task for this update.

PR #118, opened 2020-05-17
Fixed isntall command

eranco74:


eranco74:
We need to try to remove the `sudo` but for now this should fix the current failure.

PR #120, opened 2020-05-17
MGMT-747 Cluster status should be updated periodically to to reflect …

oshercc:
MGMT-747 Cluster status should be updated periodically  to reflect changed in host states

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -4,16 +4,8 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
+	"time"
 
-	"github.com/filanov/bm-inventory/internal/cluster"
-
-	"github.com/filanov/bm-inventory/internal/bminventory"
-	"github.com/filanov/bm-inventory/internal/hardware"
-	"github.com/filanov/bm-inventory/internal/host"
-	"github.com/filanov/bm-inventory/models"
-	"github.com/filanov/bm-inventory/pkg/job"
-	"github.com/filanov/bm-inventory/pkg/requestid"
-	"github.com/filanov/bm-inventory/restapi"
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
 	"github.com/jinzhu/gorm"
@@ -24,6 +16,16 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	"github.com/filanov/bm-inventory/internal/bminventory"
+	"github.com/filanov/bm-inventory/internal/cluster"
+	"github.com/filanov/bm-inventory/internal/hardware"
+	"github.com/filanov/bm-inventory/internal/host"
+	"github.com/filanov/bm-inventory/models"
+	"github.com/filanov/bm-inventory/pkg/job"
+	"github.com/filanov/bm-inventory/pkg/requestid"
+	"github.com/filanov/bm-inventory/pkg/thread"
+	"github.com/filanov/bm-inventory/restapi"
 )
 
 func init() {
@@ -31,12 +33,13 @@ func init() {
 }
 
 var Options struct {
-	BMConfig          bminventory.Config
-	DBHost            string `envconfig:"DB_HOST" default:"mariadb"`
-	DBPort            string `envconfig:"DB_PORT" default:"3306"`
-	HWValidatorConfig hardware.ValidatorCfg
-	JobConfig         job.Config
-	InstructionConfig host.InstructionConfig
+	BMConfig                    bminventory.Config
+	DBHost                      string `envconfig:"DB_HOST" default:"mariadb"`
+	DBPort                      string `envconfig:"DB_PORT" default:"3306"`
+	HWValidatorConfig           hardware.ValidatorCfg
+	JobConfig                   job.Config
+	InstructionConfig           host.InstructionConfig
+	clusterStateMonitorInterval time.Duration `envconfig:"CLUSTER_MONITOR_INTERVAL" default:"10s"`
 }
 
 func main() {
@@ -74,10 +77,16 @@ func main() {
 		log.Fatal("failed to auto migrate, ", err)
 	}
 
-	clusterApi := cluster.NewManager(log.WithField("pkg", "cluster-state"), db)
 	hwValidator := hardware.NewValidator(Options.HWValidatorConfig)
 	instructionApi := host.NewInstructionManager(log, db, hwValidator, Options.InstructionConfig)
 	hostApi := host.NewManager(log.WithField("pkg", "host-state"), db, hwValidator, instructionApi)
+	clusterApi := cluster.NewManager(log.WithField("pkg", "cluster-state"), db)
+
+	clusterStateMonitor := thread.New(
+		log.WithField("pkg", "cluster-monitor"), "State Monitor", Options.clusterStateMonitorInterval, clusterApi.ClusterMonitoring)
+	clusterStateMonitor.Start()
+	defer clusterStateMonitor.Stop()
+
 	jobApi := job.New(log.WithField("pkg", "k8s-job-wrapper"), kclient, Options.JobConfig)
 	bm := bminventory.NewBareMetalInventory(db, log.WithField("pkg", "Inventory"), hostApi, clusterApi, Options.BMConfig, jobApi)
 	h, err := restapi.Handler(restapi.Config{
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -5,12 +5,14 @@ import (
 	"fmt"
 
 	"github.com/go-openapi/strfmt"
-
-	"github.com/filanov/bm-inventory/models"
 	"github.com/go-openapi/swag"
 	"github.com/jinzhu/gorm"
 	"github.com/sirupsen/logrus"
+
+	"github.com/filanov/bm-inventory/models"
 )
+
+const minHostsNeededForInstallation = 3
 
 //go:generate mockgen -source=cluster.go -package=cluster -destination=mock_cluster_api.go
 
@@ -37,9 +39,12 @@ type API interface {
 	StateAPI
 	RegistrationAPI
 	InstallationAPI
+	ClusterMonitoring()
 }
 
 type Manager struct {
+	log             logrus.FieldLogger
+	db              *gorm.DB
 	insufficient    StateAPI
 	ready           StateAPI
 	installing      StateAPI
@@ -51,6 +56,8 @@ type Manager struct {
 
 func NewManager(log logrus.FieldLogger, db *gorm.DB) *Manager {
 	return &Manager{
+		log:             log,
+		db:              db,
 		insufficient:    NewInsufficientState(log, db),
 		ready:           NewReadyState(log, db),
 		installing:      NewInstallingState(log, db),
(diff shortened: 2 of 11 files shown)


cmd/main.go
```diff
@@ -74,10 +77,15 @@ func main() {
 		log.Fatal("failed to auto migrate, ", err)
 	}
 
-	clusterApi := cluster.NewManager(log.WithField("pkg", "cluster-state"), db)
 	hwValidator := hardware.NewValidator(Options.HWValidatorConfig)
 	instructionApi := host.NewInstructionManager(log, db, hwValidator, Options.InstructionConfig)
 	hostApi := host.NewManager(log.WithField("pkg", "host-state"), db, hwValidator, instructionApi)
+	clusterApi := cluster.NewManager(log.WithField("pkg", "cluster-monitor"), db)
```
ronniel1, line 83: This log is not only used by the monitor, but the entire cluster manager right?

internal/cluster/installing.go
```diff
@@ -23,13 +26,53 @@ type installingState baseState
 var _ StateAPI = (*Manager)(nil)
 
 func (i *installingState) RefreshStatus(ctx context.Context, c *models.Cluster, db *gorm.DB) (*UpdateReply, error) {
+	log := logutil.FromContext(ctx, i.log)
+	installationState, err := i.getClusterInstallationState(ctx, c)
+	if err != nil {
+		log.WithError(err).Errorf("cluster %s couldn't determine cluster installation state", c.ID)
```
ronniel1, line 32: What happens after logging the error?
We still follow through the switch statement, but with an invalid value

internal/cluster/installing.go
```diff
@@ -23,13 +26,53 @@ type installingState baseState
...
+	case clusterStatusError:
+		return updateState(clusterStatusError, c, i.db, log)
+	case clusterStatusInstalling:
+		return &UpdateReply{
+			State:     clusterStatusInstalling,
+			IsChanged: false,
+		}, nil
+	}
 	return &UpdateReply{
 		State:     clusterStatusInstalling,
```
ronniel1, line 46: If the status is not clear, how is it that we are returning "clusterStatusInstalling"?

internal/cluster/common.go
```diff
@@ -61,17 +61,18 @@ func getKnownMastersNodesIds(c *models.Cluster, db *gorm.DB) ([]*strfmt.UUID, er
...
-	if len(masterNodesIds) < minimumKnownMasterNodes {
-		log.Infof("cluster %s has %d known master hosts which is less then the %d minimum needed for cluster installation",
-			c.ID, len(masterNodesIds), minimumKnownMasterNodes)
-		return false, nil
-	} else {
-		return true, nil
+func mapMasterHostsByRole(c *models.Cluster) map[string][]*models.Host {
+	hostMap := make(map[string][]*models.Host)
+
+	for _, host := range c.Hosts {
```
ronniel1, line 67: From the name of this function, I understand that it returned a map of host for each role.
However, from the code I see that it also maps by statuses.

Maybe it should be named, mapMasterHostsByStatus ?

internal/cluster/insufficient.go
```diff
@@ -22,23 +24,24 @@ type insufficientState baseState
...
+			IsChanged: false}, errors.Errorf("cluster %s not found", c.ID)
 	}
-}
+	mappedMastersByRole := mapMasterHostsByRole(c)
 
-func (i *insufficientState) Install(ctx context.Context, c *models.Cluster) (*UpdateReply, error) {
-	return nil, errors.Errorf("unable to install cluster <%s> in <%s> status",
-		c.ID, swag.StringValue(c.Status))
+	// Cluster is ready
+	mastersInKnown, ok := mappedMastersByRole[host2.HostStatusKnown]
```
ronniel1, line 37: Don't we need to have hosts in 'Ready' state, and not 'Known' state?

internal/cluster/ready.go
```diff
@@ -22,23 +25,24 @@ type readyState baseState
...
+		return &UpdateReply{
+			State:     clusterStatusInsufficient,
+			IsChanged: false}, errors.Errorf("cluster %s not found", c.ID)
 	}
+	mappedMastersByRole := mapMasterHostsByRole(c)
 
-	if clusterIsReady {
+	// Cluster is insufficient
+	mastersInKnown := mappedMastersByRole[host2.HostStatusKnown]
+	if len(mastersInKnown) < minHostsNeededForInstallation {
```
ronniel1, line 39: Same as above

cmd/main.go
oshercc: yes

internal/cluster/insufficient.go
oshercc: hosts don't have ready state, known is the sufficient state for being installed

internal/cluster/common.go
oshercc: host is the role, updated to mapMasterHostsByStatus

internal/cluster/ready.go
oshercc: hosts don't have ready state, known is the sufficient state for being installed

cmd/main.go
```diff
@@ -24,19 +16,30 @@ import (
...
-	HWValidatorConfig hardware.ValidatorCfg
-	JobConfig         job.Config
-	InstructionConfig host.InstructionConfig
+	BMConfig                    bminventory.Config
+	DBHost                      string `envconfig:"DB_HOST" default:"mariadb"`
+	DBPort                      string `envconfig:"DB_PORT" default:"3306"`
+	HWValidatorConfig           hardware.ValidatorCfg
+	JobConfig                   job.Config
+	InstructionConfig           host.InstructionConfig
+	ClusterStateMonitorTimeLoop time.Duration `envconfig:"CLUSTER_MONITOR_TIME_LOOP" default:"10000"` // Microsecond
```
filanov, line 42: i think that `10s` will do the same.

cmd/main.go
```diff
@@ -46,7 +49,7 @@ func main() {
 		log.Fatal(err.Error())
 	}
 
-	port := flag.String("port", "8090", "define port that the service will listen to")
+	port := flag.String("port", "8090", "define port that the service will listen to") //nolint:staticcheck
```
filanov, line 52: don't use `//nolint:staticcheck`

cmd/main.go
filanov: so the name is not right.

cmd/main.go
```diff
@@ -65,7 +68,7 @@ func main() {
 		log.Fatal("Failed to add K8S scheme", err)
 	}
 
-	kclient, err := client.New(config.GetConfigOrDie(), client.Options{Scheme: scheme})
+	kclient, err := client.New(config.GetConfigOrDie(), client.Options{Scheme: scheme}) //nolint:staticcheck
```
filanov, line 71: don't use `//nolint:staticcheck`

cmd/main.go
ronniel1: so it should not be called cluster-monitor
instead, when starting the monitor thread, pass it a new logger with cluster-monitor name

internal/cluster/cluster.go
```diff
@@ -97,6 +104,32 @@ func (m *Manager) RefreshStatus(ctx context.Context, c *models.Cluster, db *gorm
...
+	if err := m.db.Find(&clusters).Error; err != nil {
+		m.log.WithError(err).Errorf("failed to get clusters")
+		return
+	}
+	for _, cluster := range clusters {
+		state, err := m.getCurrentState(swag.StringValue(cluster.Status))
+
+		if err != nil {
+			m.log.WithError(err).Errorf("failed to get cluster %s currentState", cluster.ID)
+			return
```
filanov, line 124: if you have one problematic cluster or problem in of the states you will have an infinite loop that will stop on this cluster and never monitor other clusters.
You need to continue not return errors her.

internal/cluster/cluster.go
```diff
@@ -97,6 +104,32 @@ func (m *Manager) RefreshStatus(ctx context.Context, c *models.Cluster, db *gorm
...
+		}
+		stateReply, err := state.RefreshStatus(context.Background(), cluster, m.db)
+		if err != nil {
+			m.log.WithError(err).Errorf("failed to refresh cluster %s state", cluster.ID)
+			return
```
filanov, line 129: same as the comment above.

internal/cluster/cluster_test.go
```diff
@@ -49,13 +50,117 @@ var _ = Describe("statemachine", func() {
...
+			Status: swag.String("installing"),
+		}
+
+		Expect(db.Create(&c).Error).ShouldNot(HaveOccurred())
+		clusterApi = NewManager(getTestLog().WithField("pkg", "cluster-monitor"), db)
+
+		Expect(err).ShouldNot(HaveOccurred())
+	})
+
+	Context("from installing state", func() {
```
filanov, line 92: how about know -> insufficient and the other way around?

pkg/thread/thread.go
```diff
@@ -9,10 +9,10 @@ import (
 // Thread provides a background, periodic thread, which invokes the given function every supplied interval.
 //
 // Sample usage:
-//    monitorFunc := fun() {
+//    monitorFunc := func() {
 //        //do monitoring logic
 //    }
-//    monitor := thread.New(log, "Health Monitor", time.Minute*2, monitorFunc)
+//    monitor := thread.NewNew(log, "Health Monitor", time.Minute*2, monitorFunc)
```
filanov, line 15: `NewNew`?

internal/cluster/cluster_test.go
```diff
@@ -63,6 +63,9 @@ All supported case options:
 installing -> installing
 installing -> installed
 installing -> error
+
```
filanov, line 66: please delete empty line

cmd/main.go
oshercc: @ronniel1 is this what you meant ?

internal/cluster/cluster_test.go
oshercc: left it in purpose, the installing state from the rest

cmd/main.go
```diff
@@ -24,19 +16,30 @@ import (
...
-	HWValidatorConfig hardware.ValidatorCfg
-	JobConfig         job.Config
-	InstructionConfig host.InstructionConfig
+	BMConfig                    bminventory.Config
+	DBHost                      string `envconfig:"DB_HOST" default:"mariadb"`
+	DBPort                      string `envconfig:"DB_PORT" default:"3306"`
+	HWValidatorConfig           hardware.ValidatorCfg
+	JobConfig                   job.Config
+	InstructionConfig           host.InstructionConfig
+	ClusterStateMonitorTimeLoop time.Duration `envconfig:"CLUSTER_MONITOR_TIME_LOOP" default:"10s"` // Microsecond
```
filanov, line 42: usually we call it interval `ClusterMonitorInterval`

cmd/main.go
```diff
@@ -74,10 +77,15 @@ func main() {
 		log.Fatal("failed to auto migrate, ", err)
 	}
 
-	clusterApi := cluster.NewManager(log.WithField("pkg", "cluster-state"), db)
 	hwValidator := hardware.NewValidator(Options.HWValidatorConfig)
 	instructionApi := host.NewInstructionManager(log, db, hwValidator, Options.InstructionConfig)
 	hostApi := host.NewManager(log.WithField("pkg", "host-state"), db, hwValidator, instructionApi)
+	clusterApi := cluster.NewManager(log.WithField("pkg", "cluster-state"), db)
+
+	clusterStateMonitor := thread.New(log.WithField("pkg", "cluster-monitor"), "State Monitor", Options.ClusterStateMonitorTimeLoop*time.Microsecond, clusterApi.ClusterMonitoring)
```
filanov, line 85: why not to break this line? it's too ling.

cmd/main.go
filanov, line 42: usually we call it interval `ClusterMonitorInterval`

PR #124, opened 2020-05-18
MGMT-848: Clean up ISOs that are older than a day

avishayt:
Adds a k8s CronJob that lists the objects in the S3 bucket that we use
for storing ISOs and deletes any that are older than 1 day. This is
because Scality does not support bucket lifecycle policies for
expiration. Deployments running in AWS shouldn't use this.

The expirer script has some environment variables that aren't passed
automatically. The defaults are currently suitable for our tests and can
be overidden in other deployments.

### DIFF
--- a/tools/expirer.py
+++ b/tools/expirer.py
@@ -1,0 +1,29 @@
+#!/usr/bin/python
+
+import boto3
+from datetime import datetime, timedelta
+import os
+import pytz
+
+S3_ENDPOINT_URL = os.environ.get('S3_ENDPOINT_URL', 'http://10.35.59.36:30925')
+S3_BUCKET = os.environ.get('S3_BUCKET', 'test')
+AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID', 'accessKey1')
+AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY', 'verySecretKey1')
+S3_OBJECT_RETENTION_DAYS = int(os.environ.get('S3_OBJECT_RETENTION_DAYS', '1'))
+ISO_PREFIX = os.environ.get('ISO_PREFIX', 'discovery-image')
+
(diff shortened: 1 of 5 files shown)


tools/expirer.py
```diff
@@ -0,0 +1,30 @@
...
+                      aws_access_key_id=AWS_ACCESS_KEY_ID,
+                      aws_secret_access_key=AWS_SECRET_ACCESS_KEY)
+paginator = client.get_paginator('list_objects')
+operation_parameters = {'Bucket': S3_BUCKET, 'Prefix': ISO_PREFIX}
+page_iterator = paginator.paginate(**operation_parameters)
+for page in page_iterator:
+    if not page.get('Contents'):
+        continue
+    for obj in page['Contents']:
+        if obj['LastModified'] < yesterday:
```
ronniel1, line 28: At 1 AM, wouldnt this delete objects that were created 2 hours ago?

deploy/s3-object-expirer-cron.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: s3-object-expirer-cron-job
+spec:
+  schedule: "0 * * * *"
```
ronniel1, line 6: I think that "@hourly" is clearer

tools/expirer.py
avishayt: It's comparing two datetimes, not just dates.  yesterday is 24 hours ago exactly.

deploy/s3-object-expirer-cron.yaml
avishayt: Will do

deploy/s3-object-expirer-cron.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1beta1
```
filanov, line 1: probably should be under s3 directory

PR #125, opened 2020-05-18
MGMT-850 VIP should validate IP format not hostname

eranco74:


filanov:
@eranco74 don't forget to generate and fix test if needed

PR #126, opened 2020-05-18
MGMT-849: Fix clusters href

avishayt:


internal/bminventory/inventory.go
```diff
@@ -223,7 +219,7 @@ func (b *bareMetalInventory) RegisterCluster(ctx context.Context, params install
 	log.Infof("Register cluster: %s with id %s", swag.StringValue(params.NewClusterParams.Name), id)
 	cluster := models.Cluster{
 		ID:                       &id,
-		Href:                     buildHref(ResourceKindCluster, id.String()),
+		Href:                     swag.String(fmt.Sprintf("%s/clusters/%s", baseHref, id.String())),
```
ronniel1, line 222: @avishayt , @filanov 
I don't think that this is the correct way to get the href.
Instead use: installer.GetClusterURL{ClusterID: id.String())

PR #128, opened 2020-05-18
MGMT-854 - Fixed commit after rollback in InstallCluster

ronniel1:
Also fixed the error handling in that function

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -11,6 +11,8 @@ import (
 	"sync"
 	"text/template"
 
+	"github.com/pkg/errors"
+
 	"github.com/filanov/bm-inventory/internal/cluster"
 	"github.com/filanov/bm-inventory/internal/host"
 	"github.com/filanov/bm-inventory/internal/installcfg"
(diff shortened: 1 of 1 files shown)


internal/bminventory/inventory.go
```diff
@@ -887,3 +888,13 @@ func generateError(id int32) *models.Error {
...
 	}
 }
+
+func generateInternalFromError(err error) *models.Error {
+	return &models.Error{
+		Code:   swag.String(string(http.StatusInternalServerError)),
+		Href:   swag.String(""),
+		ID:     swag.Int32(http.StatusInternalServerError),
+		Kind:   swag.String("Error"),
+		Reason: swag.String(err.Error()),
```
filanov, line 898: it will give the user internal information, i don't think we want to do it.
@avishayt what do you think?

ronniel1: I think that until we have a better way to handle this, this is better than nothing.
@avishayt ?

avishayt: It's OK for now, later we will want to map each 500 reason to an error ID and print something reasonable, and only log this info.

PR #131, opened 2020-05-18
Add to the image file with the git revision and push image also with git revision as a tag

ronniel1:


.github/workflows/image-push.yml
```diff
@@ -25,10 +28,14 @@ jobs:
...
       run: |
         skipper make build
+    - name: Get release version
+      id: get_version
+      run: echo ::set-env name=RELEASE_VERSION::$(echo ${GITHUB_SHA})
     - name: Publish to Registry
       uses: elgohr/Publish-Docker-Github-Action@2.14
       with:
-        name: bm-inventory:stable
+        name: bm-inventory
```
filanov, line 37: if there is not tag will it be latest?

eranco74: yes

Makefile
```diff
@@ -43,7 +43,11 @@ generate-from-swagger:
 	docker run -u $(UID):$(UID) -v $(PWD):$(PWD) -v /etc/passwd:/etc/passwd -w $(PWD) quay.io/goswagger/swagger generate client	--template=stratoscale -f swagger.yaml --template-dir=/templates/contrib
 	go generate $(shell go list ./client/... ./models/... ./restapi/...)
 
-update: build
+.PHONY: update_revision_file
+update_revision_file:
+	git rev-parse HEAD > build/git_revision
```
eranco74, line 48: You can add a label in the dockerfile https://docs.docker.com/engine/reference/builder/#label
It will show up in docker inspect, so there is no need to run the container or looking for this file. 
Also this is what they do for other openshift images:
```
        "Labels": {
            "architecture": "x86_64",
            "authoritative-source-url": "registry.access.redhat.com",
            "build-date": "2020-03-10T10:38:13.657446",
            "com.redhat.build-host": "cpt-1004.osbs.prod.upshift.rdu2.redhat.com",
            "com.redhat.component": "ubi7-container",
            "com.redhat.license_terms": "https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI",
            "description": "The Universal Base Image is designed and engineered to be the base layer for all of your containerized applications, middleware and utilities. This base image is freely redistributable, but Red Hat only supports Red Hat technologies through subscriptions for Red Hat products. This image is maintained by Red Hat and updated regularly.",
            "distribution-scope": "public",
            "io.k8s.description": "This is the base image from which all OpenShift images inherit.",
            "io.k8s.display-name": "OpenShift Base",
            "io.openshift.build.commit.author": "",
            "io.openshift.build.commit.date": "",
            "io.openshift.build.commit.id": "04c97c30e79430ad836fceff4c6f8c9429be6663",

```

PR #133, opened 2020-05-18
MGMT-853: Use PVs for mariadb and scality storage

avishayt:


tools/create_default_s3_bucket.py
```diff
@@ -1,6 +1,7 @@
 import os
 import utils
 import boto3
+import sys
```
ronniel1, line 4: Where is this used?

avishayt: Not used by this commit - it was missing previously

avishayt:
> Will this also work in minikube?

Yes, that's where I tested it.  Basically it creates /mnt/data/mariadb and /mnt/data/scality, and maps the directories to where the applications in the containers put the data.

filanov:
@avishayt probably need to add the cleanup to `clear-deployment`

avishayt:
> @avishayt probably need to add the cleanup to `clear-deployment`

Good call - added

PR #135, opened 2020-05-18
Add git revision to the container's label, instead of a file in the filesystem

ronniel1:
Getting the label can be done with the following command:
    docker inspect <image> | jq .[0].Config.Labels

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ endef
 endif
 
 SERVICE := $(or ${SERVICE},quay.io/ocpmetal/bm-inventory:stable)
+GIT_REVISION := $(shell git rev-parse HEAD)
 
 all: build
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -43,12 +44,8 @@ generate-from-swagger:
...
 	go generate $(shell go list ./client/... ./models/... ./restapi/...)
 
-.PHONY: update_revision_file
-update_revision_file:
-	git rev-parse HEAD > build/git_revision
-
-update: update_revision_file build
-	docker build -f Dockerfile.bm-inventory . -t $(SERVICE)
+update: build
+	GIT_REVISION=${GIT_REVISION} docker build --build-arg GIT_REVISION -f Dockerfile.bm-inventory . -t $(SERVICE)
```
eranco74, line 48: Should be:
```suggestion
	docker build --build-arg GIT_REVISION=${GIT_REVISION}  -f Dockerfile.bm-inventory . -t $(SERVICE)
```
No?

ronniel1: Not necessarily
when passed without a value to 'build-args' it means to take the value from host environment

PR #136, opened 2020-05-19
MGMT-808: add reason to all HTTP errors

filanov:
temporary adding the full error without filtering, later will filter out only the relevant error message for the user

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -11,8 +11,6 @@ import (
 	"sync"
 	"text/template"
 
-	"github.com/pkg/errors"
-
 	"github.com/filanov/bm-inventory/internal/cluster"
 	"github.com/filanov/bm-inventory/internal/host"
 	"github.com/filanov/bm-inventory/internal/installcfg"
@@ -26,6 +24,7 @@ import (
 	"github.com/go-openapi/swag"
 	"github.com/google/uuid"
 	"github.com/jinzhu/gorm"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	batch "k8s.io/api/batch/v1"
 	core "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


internal/bminventory/inventory.go
```diff
@@ -500,34 +499,34 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
 		tx.Rollback()
 		log.WithError(err).Errorf("failed to validate or update cluster %s state", params.ClusterID)
 		return installer.NewRegisterClusterInternalServerError().
-			WithPayload(generateError(http.StatusInternalServerError))
+			WithPayload(generateError(http.StatusInternalServerError, err))
 	}
 
 	if tx.Commit().Error != nil {
 		tx.Rollback()
 		return installer.NewUpdateClusterInternalServerError().
```
ronniel1, line 522: Don't we want this error in the log as well?

internal/bminventory/inventory.go
```diff
@@ -290,10 +289,10 @@ func (b *bareMetalInventory) DownloadClusterISO(ctx context.Context, params inst
 			Errorf("Failed to get ISO: %s", imgName)
 		if resp.StatusCode == http.StatusNotFound {
 			return installer.NewDownloadClusterISONotFound().
-				WithPayload(generateError(http.StatusNotFound))
+				WithPayload(generateError(http.StatusNotFound, errors.New(string(b))))
```
ronniel1, line 292: Dont we need b in the log as well?

filanov: it's already logged above 
```
		log.WithError(fmt.Errorf("%d - %s", resp.StatusCode, string(b))).
			Errorf("Failed to get ISO: %s", imgName)
```

internal/bminventory/inventory.go
filanov: you are right will add it.

PR #137, opened 2020-05-19
MGMT-761 move deploymnet to assisted-installer-namespace

filanov:
Will be pushed with https://github.com/tsorya/test-infra/pull/5
And after i sync with UI team

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,14 +2,15 @@ PWD = $(shell pwd)
 UID = $(shell id -u)
 
 TARGET := $(or ${TARGET},minikube)
+KUBECTL=kubectl -n assisted-installer
 
 ifeq ($(TARGET), minikube)
 define get_service
-minikube service --url $(1) | sed 's/http:\/\///g'
+minikube service --url $(1) -n assisted-installer | sed 's/http:\/\///g'
 endef
 else
 define get_service
-kubectl get service $(1) | grep $(1) | awk '{print $$4 ":" $$5}' | \
+kubectl get service $(1) -n assisted-installer | grep $(1) | awk '{print $$4 ":" $$5}' | \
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -51,9 +52,12 @@ update: update_revision_file build
 	docker build -f Dockerfile.bm-inventory . -t $(SERVICE)
 	docker push $(SERVICE)
 
-deploy-all: create-build-dir deploy-mariadb deploy-s3 deploy-service
+deploy-all: create-build-dir deploy-namespace deploy-mariadb deploy-s3 deploy-service
 	echo "Deployment done"
 
+deploy-namespace:
```
ronniel1, line 58: Doesn't this need to be done before any other deployment is done?

filanov: `deploy-all: create-build-dir deploy-namespace` it is, no?

ronniel1: there is not ordering between the prerequisites of a target
So deploy-mariadb might run before deploy-namespace
In order to set the ordering, you need to make the deploy-namespace a prerequisite of all the targets that depend on it

filanov: i can add `|` maybe? 
https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

ronniel1: The Makefile's way to do this is for each target to specify its dependencies.
In this case:
    deploy-mariadb: deploy namespace
    deploy-s3: deploy-namespace

etc

filanov: ok will change.

ronniel1:
Need to verify with test-infra

filanov:
> Need to verify with test-infra

Already did. will wait untill quay.io return to work then will ask from Jiri to support it as well, test again and push

PR #140, opened 2020-05-20
MGMT-820  - support platform none nd platform baremetal for ignition …

yevgeny-shnaidman:
…generation based on the Openshift Version

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -52,7 +52,7 @@ type Config struct {
 	ImageBuilder        string `envconfig:"IMAGE_BUILDER" default:"quay.io/oscohen/installer-image-build"`
 	ImageBuilderCmd     string `envconfig:"IMAGE_BUILDER_CMD" default:"echo hello"`
 	AgentDockerImg      string `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/oamizur/agent:latest"`
-	KubeconfigGenerator string `envconfig:"KUBECONFIG_GENERATE_IMAGE" default:"quay.io/oscohen/ignition-manifests-and-kubeconfig-generate"`
+	KubeconfigGenerator string `envconfig:"KUBECONFIG_GENERATE_IMAGE" default:"quay.io/ocpmetal/ignition-manifests-and-kubeconfig-generate:stable"`
 	InventoryURL        string `envconfig:"INVENTORY_URL" default:"10.35.59.36"`
 	InventoryPort       string `envconfig:"INVENTORY_PORT" default:"30485"`
 	S3EndpointURL       string `envconfig:"S3_ENDPOINT_URL" default:"http://10.35.59.36:30925"`
@@ -753,6 +753,10 @@ func (b *bareMetalInventory) EnableHost(ctx context.Context, params installer.En
 
 func (b *bareMetalInventory) createKubeconfigJob(cluster *models.Cluster, jobName string, cfg []byte) *batch.Job {
 	id := cluster.ID
+	overrideImageName := "registry.svc.ci.openshift.org/ocp/release:4.5.0-0.ci-2020-05-17-191604"
+	if cluster.OpenshiftVersion == models.ClusterOpenshiftVersionNr44 {
+		overrideImageName = "quay.io/openshift-release-dev/ocp-release:4.4.0-rc.7-x86_64"
+	}
 	return &batch.Job{
 		TypeMeta: meta.TypeMeta{
 			Kind:       "Job",
@@ -799,7 +803,7 @@ func (b *bareMetalInventory) createKubeconfigJob(cluster *models.Cluster, jobNam
 								},
 								{
 									Name:  "OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE",
-									Value: "quay.io/openshift-release-dev/ocp-release:4.4.0-rc.7-x86_64", //TODO: change this to match the cluster openshift version
+									Value: overrideImageName, //TODO: change this to match the cluster openshift version
 								},
 								{
 									Name:  "aws_access_key_id",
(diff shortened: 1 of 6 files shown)


internal/bminventory/inventory.go
```diff
@@ -52,7 +52,7 @@ type Config struct {
 	ImageBuilder        string `envconfig:"IMAGE_BUILDER" default:"quay.io/oscohen/installer-image-build"`
 	ImageBuilderCmd     string `envconfig:"IMAGE_BUILDER_CMD" default:"echo hello"`
 	AgentDockerImg      string `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/oamizur/agent:latest"`
-	KubeconfigGenerator string `envconfig:"KUBECONFIG_GENERATE_IMAGE" default:"quay.io/oscohen/ignition-manifests-and-kubeconfig-generate"`
+	KubeconfigGenerator string `envconfig:"KUBECONFIG_GENERATE_IMAGE" default:"quay.io/yshnaidm/ignition-manifests-and-kubeconfig-generate:baremetal_vm_v2"`
```
eranco74, line 55: Don't forget to push it to ocpmetal.

internal/bminventory/inventory.go
```diff
@@ -799,7 +799,7 @@ func (b *bareMetalInventory) createKubeconfigJob(cluster *models.Cluster, jobNam
 								},
 								{
 									Name:  "OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE",
-									Value: "quay.io/openshift-release-dev/ocp-release:4.4.0-rc.7-x86_64", //TODO: change this to match the cluster openshift version
+									Value: "registry.svc.ci.openshift.org/ocp/release:4.5.0-0.ci-2020-05-17-191604", //TODO: change this to match the cluster openshift version
```
eranco74, line 802: This will break soon, we need to find a 4.5 image from quay.io

internal/installcfg/installcfg.go
```diff
@@ -37,6 +37,55 @@ type InstallerConfig struct {
 	SSHKey     string `yaml:"sshKey"`
 }
 
+type InstallerConfigBaremetal struct {
```
eranco74, line 40: Since it's mainly the same struct (except for the platform) I think you can have a "base" struct with all the shared attributes.

internal/installcfg/installcfg.go
```diff
@@ -48,51 +97,224 @@ func countHostsByRole(cluster *models.Cluster, role string) int {
...
-				HostPrefix int    `yaml:"hostPrefix"`
-			} `yaml:"clusterNetwork"`
-			NetworkType    string   `yaml:"networkType"`
-			ServiceNetwork []string `yaml:"serviceNetwork"`
-		}{
-			ClusterNetwork: []struct {
-				Cidr       string `yaml:"cidr"`
-				HostPrefix int    `yaml:"hostPrefix"`
+	var cfg interface{}
+	if cluster.OpenshiftVersion != models.ClusterOpenshiftVersionNr44 {
```
eranco74, line 101: Must of it is common, 
You can call a method that returns the platform values, and create/update the config with it.

internal/installcfg/installcfg.go
```diff
@@ -48,51 +97,224 @@ func countHostsByRole(cluster *models.Cluster, role string) int {
...
+				Replicas:       countHostsByRole(cluster, "master"),
+			},
+			Metadata: struct {
+				Name string `yaml:"name"`
+			}{Name: cluster.Name},
+			Networking: struct {
+				ClusterNetwork []struct {
+					Cidr       string `yaml:"cidr"`
+					HostPrefix int    `yaml:"hostPrefix"`
+				} `yaml:"clusterNetwork"`
```
eranco74, line 299: Missing values here no?

internal/bminventory/inventory.go
yevgeny-shnaidman: i will do it in the next ticket

internal/installcfg/installcfg.go
yevgeny-shnaidman: since eventually we will delete the PlatformNone, and will be left only with Baremetal, then there is no need to have base and platform structs only for Baremetal

internal/installcfg/installcfg.go
yevgeny-shnaidman: again, once platform none will be removed, this code will be simplified

internal/installcfg/installcfg.go
yevgeny-shnaidman: what values? where?

internal/installcfg/installcfg.go
ronniel1, line 101: Please don't rely on the version.

I think that we need a flag, to be set by bm-inventory on which platform to use. bm-inventory can decide the value of the flag according to the version.

The benifit is that not every component, make the decision on which platform by itself, and later, if we realize that in some cases we need to use none, we only change bm-inventory

yevgeny-shnaidman: if we are going to keep none platform then we cannot rely anymore on the OpenshiftVersion, and need to pass the platform type as part of create cluster parameters. If we rely only on openshiftVersion, then the  only way to run none platform will be to set version to 4.4, in which case we will probably fail to start such a cluster after 6 month

ronniel1: I think that that is what I meant :)

eranco74: Why keep it?, What is it good for?
No one know what platform we are using, and I think we want to keep it that way.
Keeping it meaning testing it. otherwise it will not work.
We have enough problem testing (manually!) with one platform.
I think we should keep the none platform only for easy transition.

internal/installcfg/installcfg.go
eranco74: Sorry confused with this - https://github.com/filanov/bm-inventory/pull/140/files#diff-04c44b4c355d256a280425160fbed3dcR121
it's fine

PR #144, opened 2020-05-20
Ori/mgmt 822

ori-amizur:
Add support for new inventory structure.
For now, the old hardware-info, will be supported as well as the new inventory.
A new field in the host structure, called inventory, was added.
The IsSufficient logic was changed to use the new inventory

### DIFF
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -289,8 +289,8 @@ var _ = Describe("cluster", func() {
 		Expect(db.Create(&host).Error).ShouldNot(HaveOccurred())
 		return host
 	}
-	getDisk := func() *models.BlockDevice {
-		disk := models.BlockDevice{DeviceType: "loop", Fstype: "test", MajorDeviceNumber: 7, MinorDeviceNumber: 0, Mountpoint: "/sysroot", Name: "loop0", ReadOnly: true, RemovableDevice: 1, Size: 0}
+	getDisk := func() *models.Disk {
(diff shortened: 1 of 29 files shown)


avishayt:
Please add a short description to the commit messages so that people don't have to refer to the JIRA

internal/bminventory/inventory_test.go
```diff
@@ -289,8 +289,8 @@ var _ = Describe("cluster", func() {
 		Expect(db.Create(&host).Error).ShouldNot(HaveOccurred())
 		return host
 	}
-	getDisk := func() *models.BlockDevice {
-		disk := models.BlockDevice{DeviceType: "loop", Fstype: "test", MajorDeviceNumber: 7, MinorDeviceNumber: 0, Mountpoint: "/sysroot", Name: "loop0", ReadOnly: true, RemovableDevice: 1, Size: 0}
+	getDisk := func() *models.Disk {
+		disk := models.Disk{DriveType: "SSD", Name: "loop0", SizeBytes: 0}
```
filanov, line 293: Did you test it in system?

ori-amizur: test-infra.  Not full cycle, but saw that all hosts transfer to known

ori-amizur:
> @ori-amizur Does this also update the code to select installation disk to use new information?

Assuming that you mean the boot device then yes.
https://github.com/filanov/bm-inventory/blob/e84f19d616046077d7191fad6513bd046e9b9a10/internal/hardware/validator.go#L96

PR #146, opened 2020-05-20
MGMT-820 - add support for configuring machine network cidr in case o…

yevgeny-shnaidman:
…f cluster running in baremetal configuration

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ deploy-service: deploy-service-requirements deploy-role
 	python3 ./tools/deploy_assisted_installer.py
 
 deploy-expirer: deploy-role
-	python3 ./tools/deploy_s3_object_expirer.py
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -76,7 +76,7 @@ deploy-service: deploy-service-requirements deploy-role
 	python3 ./tools/deploy_assisted_installer.py
 
 deploy-expirer: deploy-role
-	python3 ./tools/deploy_s3_object_expirer.py
+	OBJEXP=$(OBJEXP)  python3 ./tools/deploy_s3_object_expirer.py
```
filanov, line 79: lets use parameter and not env

romfreiman:
It's not needed

eranco74:
This is the discussion https://issues.redhat.com/browse/MGMT-868
We can deduce the machine network from the api VIP and hardware info.

PR #152, opened 2020-05-21
Ori/mgmt 868 Add machine network parameter to cluster create and update params

ori-amizur:


internal/installcfg/installcfg.go
```diff
@@ -96,7 +102,43 @@ func countHostsByRole(cluster *models.Cluster, role string) int {
...
 }
 
+func getMachineCIDR(cluster *models.Cluster) (string, error) {
+	parsedVipAddr := net.ParseIP(string(cluster.APIVip))
+	if parsedVipAddr == nil {
+		errStr := fmt.Sprintf("Could not parse VIP ip %s", cluster.APIVip)
+		logrus.Warn(errStr)
+		return "", errors.New(errStr)
+	}
+	for _, h := range cluster.Hosts {
```
eranco74, line 112: Basically all hosts should have an interface with the machine network CIDR.  
Consider adding a log in case of failure to find it in one of the hosts.

eranco74: Actually leave it to MGMT-869

PR #153, opened 2020-05-24
Yevgeny/update release version

yevgeny-shnaidman:


internal/bminventory/inventory.go
```diff
@@ -771,7 +771,7 @@ func (b *bareMetalInventory) EnableHost(ctx context.Context, params installer.En
 
 func (b *bareMetalInventory) createKubeconfigJob(cluster *models.Cluster, jobName string, cfg []byte) *batch.Job {
 	id := cluster.ID
-	overrideImageName := "registry.svc.ci.openshift.org/ocp/release:4.5.0-0.ci-2020-05-17-191604"
+	overrideImageName := "registry.svc.ci.openshift.org/ocp/release:4.5.0-0.nightly-2020-05-21-015458"
```
romfreiman, line 774: @eranco74 @yevgeny-shnaidman @avishayt @filanov is there a task to parametrize it via bm-inventory?

filanov: The parameter should be openshift version no? the user will not know what version he need to select.

avishayt: After this commit we support 2 versions, and each version has its own image:
4.5: release:4.5.0-0.nightly-2020-05-21-015458
4.4: ocp-release:4.4.0-rc.7-x86_64

As I mentioned in my review, this small change is fine for now, but as we have more testing resources and support more versions, we will need to make the code clearer instead of if/else if/else if/else if/...
We should probably store the mapping in a yaml file and load it into a map or something...

yevgeny-shnaidman: our goal should be to use the image based on the openshift version that is passed. So ideally we should use the image contained in the openshift-install. But we are not there yet

PR #154, opened 2020-05-24
Change register host to new state machine model

filanov:
Change register host to new state machine model
    Fix registration API - Register will check the current state before activating state machine
    Cleanup existing code

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -557,16 +557,8 @@ func (b *bareMetalInventory) GetCluster(ctx context.Context, params installer.Ge
 
 func (b *bareMetalInventory) RegisterHost(ctx context.Context, params installer.RegisterHostParams) middleware.Responder {
 	log := logutil.FromContext(ctx, b.log)
-	url := installer.GetHostURL{ClusterID: params.ClusterID, HostID: *params.NewHostParams.HostID}
-	host := &models.Host{
-		ID:        params.NewHostParams.HostID,
-		Href:      swag.String(url.String()),
-		Kind:      swag.String(ResourceKindHost),
-		Status:    swag.String("discovering"),
-		ClusterID: params.ClusterID,
-	}
-
-	log.Infof("Register host: %+v", host)
+	var host models.Host
+	log.Infof("Register host: %+v", params)
 
 	if err := b.db.First(&models.Cluster{}, "id = ?", params.ClusterID.String()).Error; err != nil {
 		log.WithError(err).Errorf("failed to get cluster: %s", params.ClusterID.String())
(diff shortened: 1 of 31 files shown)


internal/host/transition.go
```diff
@@ -0,0 +1,67 @@
...
+	params, ok := args.(*TransitionArgsRegisterHost)
+	if !ok {
+		return errors.New("PostRegisterHost invalid argument")
+	}
+
+	host := models.Host{}
+	log := logutil.FromContext(params.ctx, th.log)
+
+	// if already exists, reset role and hw info
+	if err := th.db.First(&host, "id = ? and cluster_id = ?", sHost.host.ID, sHost.host.ClusterID).Error; err == nil {
```
ori-amizur, line 45: You may use gorm.FirstOrCreate

internal/bminventory/inventory.go
```diff
@@ -557,31 +557,38 @@ func (b *bareMetalInventory) GetCluster(ctx context.Context, params installer.Ge
...
+	log.Infof("Register host: %+v", params)
 
 	if err := b.db.First(&models.Cluster{}, "id = ?", params.ClusterID.String()).Error; err != nil {
 		log.WithError(err).Errorf("failed to get cluster: %s", params.ClusterID.String())
 		return installer.NewRegisterHostBadRequest().
 			WithPayload(generateError(http.StatusBadRequest, err))
 	}
 
-	if _, err := b.hostApi.RegisterHost(ctx, host); err != nil {
+	err := b.db.First(&host, "id = ? && cluster_id = ?", *params.NewHostParams.HostID, params.ClusterID).Error
```
ori-amizur, line 569: I think that one of the advantages of state machine is minimize the locations where decisions are taken.  So if we can fill arguments and let the state machine do all the decisions it will make things clearer.

internal/host/statehost.go
```diff
@@ -0,0 +1,28 @@
+package host
+
+import (
+	"github.com/filanov/bm-inventory/models"
+	"github.com/filanov/stateswitch"
+	"github.com/go-openapi/swag"
+)
+
+type stateHost struct {
+	srcState string
```
ori-amizur, line 10: Seems there is duplication here  srcState <-> h.Status.  Why is it needed?

internal/bminventory/inventory.go
filanov: Did it because we don't know the state of the host to start with.
Usually all the apis are called on existing object. 
But i can move this code to host API if needed.

internal/host/common.go
```diff
@@ -69,6 +69,32 @@ func updateStateWithParams(log logrus.FieldLogger, status, statusInfo string, h
 	}, nil
 }
 
+func updateHostStateWithParams(log logrus.FieldLogger, srcStatus, statusInfo string, h *models.Host, db *gorm.DB,
```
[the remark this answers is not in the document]
filanov: it's not the same function, and i didn't want to break anything or change existing behavior. after i finish the full transition i will probably delete `updateStateWithParams`

internal/host/statehost.go
filanov: when i get to `PostTrantions` the state in `h.Status` is already the new state. i want to get the source state so i can make the change the db with the right source. from state x to state y.

internal/host/transition.go
filanov: I will need to copy all the parameters.
If i create a new one no problem.
if i get the host then i will loos the target state. that is set into `shost.host`

internal/bminventory/inventory.go
filanov: Looked at it again, and i like the idea. moved some of the logic to host api.

PR #155, opened 2020-05-24
MGMT-905 assisted install container should be named explicitly

eranco74:


eranco74:
Surprisingly all UT and subsystem passed.
`
Ran 17 of 17 Specs in 69.005 seconds
SUCCESS! -- 17 Passed | 0 Failed | 0 Pending | 0 Skipped
`
We have no coverage for the install command?

yevgeny-shnaidman:
probably there is a lot of coverage missing. Maybe, after 1of June, we should take a couple of days to update test coverage for most of the repositories

filanov:
@eranco74 isn't there UT test that can validate that we pass the right format?

filanov:
subsystem don't check it because it's not actually running this code.

eranco74:
All UT passed, I didn't changed them - hence no coverage.

filanov:
so maybe you should add a test that check that we don't break the format of the command?

eranco74:
Can I get an approve PR this PR regardless of the missing tests?

PR #156, opened 2020-05-24
MGMT-796 Implement connectivity check getNextStep

gamli75:
@ronniel1 @ori-amizur @filanov please review

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -719,55 +719,79 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params installer.
 }
 
 func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer.PostStepReplyParams) middleware.Responder {
+	var err error
 	log := logutil.FromContext(ctx, b.log)
 	log.Infof("Received step reply <%s> from cluster <%s> host <%s>  exit-code <%d> stdout <%s> stderr <%s>", params.Reply.StepID, params.ClusterID,
 		params.HostID, params.Reply.ExitCode, params.Reply.Output, params.Reply.Error)
 
+	//check the output exit code
+	if params.Reply.ExitCode != 0 {
+		err = fmt.Errorf("Exit code is %d reply error is %s for %s reply for host %s cluster %s",
+			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
+		log.WithError(err).Errorf("Exit code is <%d> , reply error is <%s> for <%s> reply for host <%s> cluster <%s>",
+			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
+		return installer.NewPostStepReplyBadRequest().
+			WithPayload(common.GenerateError(http.StatusBadRequest, err))
+	}
+
 	var host models.Host
-	if err := b.db.First(&host, "id = ? and cluster_id = ?", params.HostID, params.ClusterID).Error; err != nil {
+	if err = b.db.First(&host, "id = ? and cluster_id = ?", params.HostID, params.ClusterID).Error; err != nil {
 		log.WithError(err).Errorf("Failed to find host <%s> cluster <%s> step <%s>",
 			params.HostID, params.ClusterID, params.Reply.StepID)
 		return installer.NewPostStepReplyNotFound().
 			WithPayload(common.GenerateError(http.StatusNotFound, err))
 	}
 
-	if strings.HasPrefix(params.Reply.StepID, string(models.StepTypeHardwareInfo)) {
-		// To make sure we store only information defined in swagger we unmarshal and marshal hw info.
-		hwInfo, err := filterReply(&models.Introspection{}, params.Reply.Output)
-		if err != nil {
-			log.WithError(err).Errorf("Failed decode <%s> reply for host <%s> cluster <%s>",
-				params.Reply.StepID, params.HostID, params.ClusterID)
-			return installer.NewPostStepReplyBadRequest().
-				WithPayload(common.GenerateError(http.StatusBadRequest, err))
-		}
-
-		if _, err := b.hostApi.UpdateHwInfo(ctx, &host, hwInfo); err != nil {
-			log.WithError(err).Errorf("Failed to update host <%s> cluster <%s> step <%s>",
-				params.HostID, params.ClusterID, params.Reply.StepID)
-			return installer.NewPostStepReplyInternalServerError().
-				WithPayload(common.GenerateError(http.StatusInternalServerError, err))
-		}
+	var stepReply string
+	stepReply, err = filterReplyByType(params)
+	if err != nil {
+		log.WithError(err).Errorf("Failed decode <%s> reply for host <%s> cluster <%s>",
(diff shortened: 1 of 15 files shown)


internal/hardware/validator.go
```diff
@@ -105,6 +106,17 @@ func (v *validator) GetHostValidDisks(host *models.Host) ([]*models.Disk, error)
 	return disks, nil
 }
 
+func (v *validator) GetHostValidNics(host *models.Host) ([]*models.Nic, error) {
+	var hwInfo models.Introspection
```
ronniel1, line 110: This should be from the inventory

internal/host/host.go
```diff
@@ -33,6 +33,8 @@ type StateAPI interface {
 	EnableHost(ctx context.Context, h *models.Host) (*UpdateReply, error)
 	// Disable host from getting any requests
 	DisableHost(ctx context.Context, h *models.Host) (*UpdateReply, error)
+	// Set a new Connectivity report information
+	UpdateConnectivityReport(ctx context.Context, h *models.Host, connectivityReport string) (*UpdateReply, error)
```
ronniel1, line 37: Why is this part of the StateAPI?
I don't think we need to do something different in each state.
Just update the DB with the connectivity report we got.

internal/bminventory/inventory.go
```diff
@@ -656,52 +656,62 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params installer.
...
+		stepReply, err = filterReply(&models.Inventory{}, params.Reply.Output)
+	}
 
-		if _, err := b.hostApi.UpdateInventory(ctx, &host, inventory); err != nil {
-			log.WithError(err).Errorf("Failed to update host <%s> cluster <%s> step <%s>",
-				params.HostID, params.ClusterID, params.Reply.StepID)
-			return installer.NewPostStepReplyInternalServerError().
-				WithPayload(generateError(http.StatusInternalServerError, err))
-		}
+	if err != nil {
```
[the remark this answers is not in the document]
gamli75: I know :) thought about it a lot - I think that this code is better

internal/host/host.go
gamli75: We talked about it in the first meeting - I think we can ask the same question about the update inventory/ HW info. @filanov

internal/hardware/validator.go
gamli75: @ori-amizur can I add both IPv4 & IPv6 to models.ConnectivityCheckHost.IPAddresses?

ori-amizur: No.  IPv4 only for now.

internal/host/connectivitycheckconvertor.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"encoding/json"
+
+	"github.com/filanov/bm-inventory/internal/hardware"
+	"github.com/filanov/bm-inventory/models"
+	"github.com/go-openapi/strfmt"
+)
+
+func convertHostsToConnectivityParamsHosts(currentHostId *strfmt.UUID, hosts []*models.Host, hwValidator hardware.Validator) (string, error) {
+	var connectivityCheckHosts []models.ConnectivityCheckHost
```
ori-amizur, line 12: Why not ConnectivityCheckParams as it appears in the model?

internal/bminventory/inventory.go
```diff
@@ -656,52 +656,62 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params installer.
...
 
 func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer.PostStepReplyParams) middleware.Responder {
+	var err error
 	log := logutil.FromContext(ctx, b.log)
 	log.Infof("Received step reply <%s> from cluster <%s> host <%s>  exit-code <%d> stdout <%s> stderr <%s>", params.Reply.StepID, params.ClusterID,
 		params.HostID, params.Reply.ExitCode, params.Reply.Output, params.Reply.Error)
 
+	//check the output exit code
+	if params.Reply.ExitCode != 0 {
+		log.WithError(err).Errorf("Exit code is <%d> for <%s> reply for host <%s> cluster <%s>",
```
ori-amizur, line 666: It might be useful to add the params.reply.Error in the log message since it is the error string

internal/host/host.go
gamli75: @ronniel1 @filanov I didn't implement it yet but we said that if there is no network connection to all other hosts on all interfaces - we will update the state to insufficient and not wait till the cluster monitor will change the host state. But now that I'm thinking about it - maybe it's not a good idea that we will have 2 places that change the host state to insufficient.

filanov: I think that first step should be just keep it in the DB, without doing anything active like changing states. 
Regarding your comment @gamli75, if we are moving this logic to a monitor it will probably need to handle hardware as well. but we can do it later.

internal/bminventory/inventory.go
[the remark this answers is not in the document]
gamli75: split it to 2 methods

internal/hardware/validator.go
gamli75: :+1:

internal/host/host.go
gamli75: removed it from stateAPI

internal/host/connectivitycheckconvertor.go
gamli75: fixed

internal/bminventory/inventory.go
gamli75: fixed

internal/host/connectivitycheckcmd_test.go
```diff
@@ -22,7 +24,7 @@ var _ = Describe("connectivitycheckcmd", func() {
 
 	BeforeEach(func() {
 		db = prepareDB()
-		connectivityCheckCmd = NewConnectivityCheckCmd(getTestLog(), db)
+		connectivityCheckCmd = NewConnectivityCheckCmd(getTestLog(), db, mockValidator)
```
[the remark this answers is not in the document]
gamli75: fixed

gamli75:
@ronniel1 @filanov @ori-amizur I fixed the remarks, please review again

internal/host/connectivitycheckconvertor.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func convertInterfacesToConnectivityCheckHost(hostId *strfmt.UUID, interfaces []*models.Interface) *models.ConnectivityCheckHost {
+	var connectivityHost models.ConnectivityCheckHost
+	connectivityHost.HostID = *hostId
+	for _, hostInterface := range interfaces {
+		var connectivityNic models.ConnectivityCheckNic
+		connectivityNic.Mac = hostInterface.MacAddress
+		connectivityNic.Name = hostInterface.Name
+		connectivityNic.IPAddresses = hostInterface.IPV4Addresses
```
ori-amizur, line 36: You need to strip the mask from each ip-address

gamli75: fixed

gamli75:
@ori-amizur @filanov @ronniel1 please approve

PR #158, opened 2020-05-25
mgmt-797: Store image properties, one image per cluster

avishayt:


internal/bminventory/inventory.go
```diff
@@ -296,24 +297,69 @@ func (b *bareMetalInventory) DownloadClusterISO(ctx context.Context, params inst
...
+	tx := b.db.Begin()
+	defer func() {
+		if r := recover(); r != nil {
+			log.Error("generate cluster ISO failed")
+			tx.Rollback()
+		}
+	}()
+
+	if tx.Error != nil {
+		log.WithError(tx.Error).Error("failed to start transaction")
```
filanov, line 317: Need to return an error, and please fix the `UpdateCluster` place as well.

internal/bminventory/inventory.go
```diff
@@ -296,24 +297,69 @@ func (b *bareMetalInventory) DownloadClusterISO(ctx context.Context, params inst
...
-	imgName := getImageName(params.ClusterID, imgId)
-	// max job name is 63 chars
-	jobName := fmt.Sprintf("create-image-%s-%s", cluster.ID, imgId)[:63]
+
+	/* We need to ensure that the metadata in the DB matches the image that will be uploaded to S3,
+	so we check that at least 10 seconds have past since the previous request to reduce the chance
+	of a race between two consecutive requests.
+	*/
+	now := time.Now()
+	previousCreatedAt := time.Time(cluster.ImageInfo.CreatedAt)
```
filanov, line 332: i think that i can be nill. and it if is you will have panic.

pkg/job/job.go
```diff
@@ -45,15 +49,15 @@ func (k *kubeJob) Create(ctx context.Context, obj runtime.Object, opts ...client
 	return k.kube.Create(ctx, obj, opts...)
 }
 
-// Monitor k8s job
-func (k *kubeJob) Monitor(ctx context.Context, name, namespace string) error {
-	log := logutil.FromContext(ctx, k.log)
-	var job batch.Job
+func getJob(k *kubeJob, ctx context.Context, job *batch.Job, name, namespace string) error {
```
filanov, line 52: why not `func (k KubeJob) getJob(ctx ...)`?

pkg/job/job.go
```diff
@@ -94,3 +104,39 @@ func (k *kubeJob) Monitor(ctx context.Context, name, namespace string) error {
...
 	return nil
 }
+
+// Delete k8s job
+func (k *kubeJob) Delete(ctx context.Context, name, namespace string) error {
+	log := logutil.FromContext(ctx, k.log)
+	var job batch.Job
+
+	if err := getJob(k, ctx, &job, name, namespace); err != nil {
+		if strings.HasSuffix(err.Error(), " not found") {
```
filanov, line 114: they have no error type? 
it's really risky to do this way.

pkg/job/job.go
```diff
@@ -94,3 +104,39 @@ func (k *kubeJob) Monitor(ctx context.Context, name, namespace string) error {
...
+			log.Infof("Job <%s> was not found for deletion, probably already completed", name)
+			return nil
+		}
+		log.WithError(err).Errorf("Failed to get job <%s> for deletion", name)
+		return err
+	}
+
+	// not deleting a job if it failed
+	if job.Status.Failed >= swag.Int32Value(job.Spec.BackoffLimit)+1 {
```
filanov, line 123: it's a general comment, but we don't have garbage collector for failed jobs. maybe we need to  open a ticket for it.
Or store the logs of the pods somehow and delete the job.

internal/bminventory/inventory.go
avishayt: Ah so you found where I copy-pasted from :)

internal/bminventory/inventory.go
avishayt: No, all date-times are initiated to 0001-01-01T00:00:00.  So for a new cluster it will have this previous time, we will try to kill that job that started 2019 years ago, not find it, and move on.  It keeps the same flow for the first image as for the next ones which I think is nice.

pkg/job/job.go
avishayt: Because I'm not good at Go yet :)
Fixing...

internal/bminventory/inventory.go
filanov: i will probably do the same :)

internal/bminventory/inventory.go
filanov: Date-time yes, but `ImageInfo` is a pointer to struct.

avishayt: It is never nil in this code - if it was then generateImage would fail every time.  Here is what's returned when creating a new cluster:

{"created_at":"2020-05-25T05:22:39","hosts":null,"href":"/api/assisted-install/v1/clusters/9cd28ce9-ccdd-4fe2-b749-86feda0d47fa","id":"9cd28ce9-ccdd-4fe2-b749-86feda0d47fa","image_info":{"created_at":"0001-01-01T00:00:00"},"install_completed_at":"0001-01-01T00:00:00","install_started_at":"0001-01-01T00:00:00","kind":"Cluster","name":"bar","openshift_version":"4.4","status":"insufficient","status_info":"insufficient","updated_at":"2020-05-25T05:22:39"}

pkg/job/job.go
avishayt: Fixed

pkg/job/job.go
avishayt: Yes agree, though I guess it should be fairly rare so not high priority

PR #159, opened 2020-05-25
MGMT-915 - adjust hosts section of installconfig to reflect the corre…

yevgeny-shnaidman:
…ct number of hosts intances

### DIFF
--- a/internal/installcfg/installcfg.go
+++ b/internal/installcfg/installcfg.go
@@ -11,6 +11,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+// [TODO] - remove once we stop supporting none platform
 type InstallerConfigNone struct {
 	APIVersion string `yaml:"apiVersion"`
 	BaseDomain string `yaml:"baseDomain"`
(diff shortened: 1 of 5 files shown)


internal/installcfg/installcfg.go
```diff
@@ -134,185 +143,133 @@ func getMachineCIDR(cluster *models.Cluster) (string, error) {
...
+	}
+	hwMac[len(hwMac)-1] = hwMac[len(hwMac)-1] + byte(count)
+	return hwMac.String(), nil
+}
+
+func setPlatformInstallconfig(cluster *models.Cluster, cfg *InstallerConfigBaremetal) error {
+	// set hosts
+	numMasters := countHostsByRole(cluster, "master")
+	numWorkers := countHostsByRole(cluster, "worker")
+	logrus.Infof("Number of masters in cluster i %d, number of worker is %d", numMasters, numWorkers)
```
romfreiman, line 213: workers

internal/installcfg/installcfg.go
```diff
@@ -134,185 +143,133 @@ func getMachineCIDR(cluster *models.Cluster) (string, error) {
...
+	masterCount := 0
+	workerCount := 0
+	hosts := make([]host, numWorkers+numMasters)
+
+	// dummy MAC and port, once we start using real BMH, those values should be set from cluster
+	dummyMAC := "00:aa:39:b3:51:10"
```
romfreiman, line 219: maybe dont set same mac but ++ every time?

yevgeny-shnaidman: that's what i do, see function getDummyMAC

internal/installcfg/installcfg.go
yevgeny-shnaidman: removed this printout , it was for debug purposes

romfreiman: why? Lets have those logs

yevgeny-shnaidman: Believe me , this one was really not needed :)

internal/installcfg/installcfg.go
```diff
@@ -134,185 +143,132 @@ func getMachineCIDR(cluster *models.Cluster) (string, error) {
...
+	dummyPort := 6230
+
+	for i := range hosts {
+		logrus.Infof("Setting master, host %d, master count %d", i, masterCount)
```
eranco74, line 222:
```suggestion
		logrus.Infof("Setting host info, host %d, master count %d", i, masterCount)
```
You are setting both workers and masters

yevgeny-shnaidman: as explained above, this is a debug log, not needed anymore, i removed it

PR #162, opened 2020-05-25
Avishay/mgmt 908 - APIs for kubeconfig

avishayt:


swagger.yaml
```diff
@@ -264,6 +264,73 @@ paths:
           schema:
             $ref: '#/definitions/error'
 
+  /clusters/{cluster_id}/downloads/kubeconfig:
```
tsorya, line 267: don't u think better to call it kubeconfig-withingress ?
less changes :)

avishayt: No, don't be lazy ;)
Users expect kubeconfig to work with ingress.  They shouldn't need to know about it even.  The strange one is kubeconfig-noingress which will not work for ingress.

swagger.yaml
```diff
@@ -264,6 +264,73 @@ paths:
...
+        - multipart/form-data
+      parameters:
+        - in: path
+          name: cluster_id
+          type: string
+          format: uuid
+          required: true
+        - in: formData
+          name: upfile
+          type: file
```
tsorya, line 316: why file? lets just make it string

tsorya: it is relatively small amount of data, not sure it worth to be file, it will never saved as file

filanov:
is that a WIP? where is the implementation?

PR #163, opened 2020-05-25
Change user from systemUser to core

eranco74:


eranco74:
`
[root@hpe-dl580gen9-01 test-infra]# ssh -i ssh_key/key core@192.168.126.12
Fedora CoreOS 31.20200319.dev.1
Tracker: https://github.com/coreos/fedora-coreos-tracker
Discuss: https://discussion.fedoraproject.org/c/server/coreos/

`

internal/bminventory/inventory.go
```diff
@@ -64,15 +64,7 @@ const ignitionConfigFormat = `{
 "ignition": { "version": "3.0.0" },
   "passwd": {
     "users": [
-      {
-        "groups": [
-          "sudo",
-          "docker"
-        ],
-        "name": "core",
-        "passwordHash": "$6$MWO4bibU8TIWG0XV$Hiuj40lWW7pHiwJmXA8MehuBhdxSswLgvGxEh8ByEzeX2D1dk87JILVUYS4JQOP45bxHRegAB9Fs/SWfszXa5."
```
romfreiman, line 73: What is it?

eranco74: qwerty

eranco74:
In case the user didn't pass ssh public key I will not create the user.
I can create the user and allow login with password but unsure we want to.
@ronniel1 , @avishayt , @filanov , @romfreiman @oshercc , did we have the user password here on purpose?

avishayt:
> In case the user didn't pass ssh public key I will not create the user.
> I can create the user and allow login with password but unsure we want to.
> @ronniel1 , @avishayt , @filanov , @romfreiman @oshercc , did we have the user password here on purpose?

No, we don't want password-based login

romfreiman:
is it console password? if so, we might want for now - in care the vm will
not get network

eranco74:
OK,
Currently the status is this:
We have password based-login both for core user.
In addition if the user supply  ssh public key we are adding ssh & password based login for systemUser user.
After this PR is merged we will have:
In case the user provided ssh public key we will have ssh and password based login.
As Rom said we might want the password based login for debuging network issues.
Note that this is only for the ISO in live CD, once the install start and the host reboot this configuration is deleted.

avishayt:
The password-based login is via SSH, console, or both?  Is SSH configured to allow password auth?

eranco74:
ssh is configured not to allow password based login.
The password will work only from console

avishayt:
> ssh is configured not to allow password based login.
> The password will work only from console

OK excellent

PR #165, opened 2020-05-25
Osher/subsystem fix

oshercc:


subsystem/cluster_test.go
```diff
@@ -207,14 +208,24 @@ var _ = Describe("system-test cluster install", func() {
...
-				By("update host progress to done")
+
 				for _, host := range c.Hosts {
 					updateProgress(*host.ID, "Done")
 				}
-				//rep, err = bmclient.Installer.GetCluster(ctx, &installer.GetClusterParams{ClusterID: clusterID})
-				//Expect(err).NotTo(HaveOccurred())
-				//c = rep.GetPayload()
-				//Expect(swag.StringValue(c.Status)).Should(Equal("installed"))
+				rep, err = bmclient.Installer.GetCluster(ctx, &installer.GetClusterParams{ClusterID: clusterID})
```
filanov, line 215: probably should be a function something like wait for cluster state.

filanov:
@oshercc i see that in gomega there is something called `Eventually` maybe you can use it 
https://onsi.github.io/gomega/

subsystem/cluster_test.go
```diff
@@ -460,3 +460,19 @@ var _ = Describe("system-test cluster install", func() {
 		Expect(*h1.Status).Should(Equal("insufficient"))
 	})
 })
+
+func waitForClusterState(ctx context.Context, clusterID strfmt.UUID, fromState, toState string) {
```
filanov, line 464: you don't really need the from. just the target state no?

PR #166, opened 2020-05-25
Ori/mgmt 911 Move calculation of machine CIDR to pre-install phase

ori-amizur:
Add validations for that APIVip is valid during cluster update
Calculate machine network cidr during cluster update and save in database.
Add validation that all masters have IP address belonging to machine network cidr before installation.


internal/bminventory/inventory.go
```diff
@@ -343,6 +342,11 @@ func getImageName(clusterID, id strfmt.UUID) string {
 	return fmt.Sprintf("discovery-image-%s-%s", clusterID.String(), id)
 }
 
+func logAndGenerateError(id int32, errStr string) *models.Error {
```
filanov, line 345: maybe it should be 
```suggestion
func logAndGenerateError(id int32, err error) *models.Error {
```
Then when you are using it you can use errors.Wrap
there are some places that you didn't log the original error. 

Another solution is maybe add the log to `generateError` but then you will need to pass context or logger.
@ronniel1 what do you think? 
if are already using generateErr then maybe the log should be there. and we will not need to log and create error separately.

internal/bminventory/inventory.go
```diff
@@ -366,6 +370,49 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
...
+		for _, h := range hosts {
+			if *h.ID == id {
+				return true
+			}
+		}
+		return false
+	}
+
+	for _, id := range masterNodesIds {
+		if !hostIDInCidrHosts(*id, machineCidrHosts) {
```
filanov, line 408: `tx.Rollback()`

internal/bminventory/machine_network_cidr.go
```diff
@@ -0,0 +1,85 @@
+package bminventory
+
+import (
+	"encoding/json"
+	"errors"
```
filanov, line 5: suggesting to switch to `"github.com/pkg/errors"`

internal/bminventory/inventory.go
```diff
@@ -366,6 +370,49 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
 			WithPayload(generateError(http.StatusNotFound, err))
 	}
 
+	cidr, err := calculateMachineNetworkCIDR(&cluster)
+	if err != nil {
+		tx.Rollback()
+		return installer.NewInstallClusterInternalServerError().WithPayload(generateError(http.StatusInternalServerError, err))
```
filanov, line 376: isn't it conflict or bad request?

internal/bminventory/machine_network_cidr.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/filanov/bm-inventory/models"
+	"github.com/sirupsen/logrus"
+)
+
+func calculateMachineNetworkCIDR(cluster *models.Cluster) (string, error) {
+	parsedVipAddr := net.ParseIP(string(cluster.APIVip))
+	if parsedVipAddr == nil {
+		errStr := fmt.Sprintf("Could not parse VIP ip %s", cluster.APIVip)
+		logrus.Warn(errStr)
```
filanov, line 17: Usually we log errors in the api. otherwise you will have the same log multiple times. 
I'm talking only about the cases where you return an error.

internal/bminventory/machine_network_cidr.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		for _, intf := range inventory.Interfaces {
+			for _, ipv4addr := range intf.IPV4Addresses {
+				_, ipnet, err := net.ParseCIDR(ipv4addr)
+				if err != nil {
+					logrus.WithError(err).Warnf("Could not parse cidr %s", ipv4addr)
+					continue
+				}
+				if ipnet.Contains(parsedVipAddr) {
+					return ipnet.String(), nil
```
filanov, line 35: it calculate the CIDR based on one host?

internal/bminventory/machine_network_cidr_test.go
```diff
@@ -0,0 +1,108 @@
...
+	createCluster := func(apiVip string, machineCidr string, inventories ...string) *models.Cluster {
+		return &models.Cluster{
+			APIVip:             strfmt.IPv4(apiVip),
+			MachineNetworkCidr: machineCidr,
+			Hosts:              createHosts(inventories...),
+		}
+	}
+	Context("calculateMachineNetworkCIDR", func() {
+		It("happpy flow", func() {
+			cluster := createCluster("1.2.5.6", "",
```
filanov, line 43: looks like something that can be a table test.

internal/bminventory/machine_network_cidr.go
ori-amizur: The first one it finds with matching CIDR

internal/bminventory/inventory.go
ori-amizur: Probably BadRequest

internal/bminventory/inventory.go
ori-amizur: @filanov I think you should consider using gorm.DB.Transaction.  This will prevent this type of error

internal/bminventory/machine_network_cidr.go
ori-amizur: On the first match found

internal/bminventory/machine_network_cidr_test.go
ori-amizur: @filanov What do u mean by table test?

filanov: table driven

internal/bminventory/machine_network_cidr.go
filanov: you probably should add a comment to explain it.

internal/bminventory/machine_network_cidr_test.go
ori-amizur: @filanov Example?

internal/common/machine_network_cidr.go
```diff
@@ -0,0 +1,132 @@
...
+			continue
+		}
+		for _, intf := range inventory.Interfaces {
+			for _, ipv4addr := range intf.IPV4Addresses {
+				_, ipnet, err := net.ParseCIDR(ipv4addr)
+				if err != nil {
+					continue
+				}
+				if ipnet.Contains(parsedVipAddr) {
+					return ipnet.String(), nil
```
gamli75, line 45: I think that you should add info log here

internal/common/machine_network_cidr.go
```diff
@@ -0,0 +1,132 @@
...
+		return false
+	}
+	for _, intf := range inventory.Interfaces {
+		for _, ipv4addr := range intf.IPV4Addresses {
+			ip, _, err := net.ParseCIDR(ipv4addr)
+			if err != nil {
+				logrus.WithError(err).Warnf("Could not parse cidr %s", ipv4addr)
+				continue
+			}
+			if machineIpnet.Contains(ip) {
```
gamli75, line 101: @ronniel1 this should be machineIpnet.String() == ip.String() and not contains, right?

internal/common/machine_network_cidr.go
```diff
@@ -0,0 +1,132 @@
...
+	} else if cluster.IngressVip != "" {
+		ip = cluster.IngressVip.String()
+	} else {
+		return "", nil
+	}
+	parsedVipAddr := net.ParseIP(ip)
+	if parsedVipAddr == nil {
+		return "", fmt.Errorf("Could not parse VIP ip %s", cluster.APIVip)
+	}
+	for _, h := range cluster.Hosts {
```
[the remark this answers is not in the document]
ori-amizur: The return value is different.  Here we return the CIDR and belongsToNetwork return bool

internal/common/machine_network_cidr.go
ori-amizur: It is logged on the API level

internal/common/machine_network_cidr.go
ori-amizur: The Contains method refers to ip-address contained in network.  @gamli75 What you suggested will not work here.

internal/host/host.go
```diff
@@ -26,6 +26,8 @@ type StateAPI interface {
 	UpdateRole(ctx context.Context, h *models.Host, role string, db *gorm.DB) (*UpdateReply, error)
 	// check keep alive
 	RefreshStatus(ctx context.Context, h *models.Host) (*UpdateReply, error)
+	// Refresh the current state
+	RefreshState(ctx context.Context, h *models.Host, db *gorm.DB) (*UpdateReply, error)
```
gamli75, line 30: I don't think it should be another method - your code should be inside the refreshStatus method

ori-amizur: We have to think about it because what RefreshStatus does is check the keep alive and move to disconnected when there is not connection for some period.
The RefreshState updates the state because some attributes have changed that have changed may cause state change.

internal/bminventory/inventory.go
```diff
@@ -581,7 +664,26 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
+			if _, err = b.hostApi.RefreshState(ctx, &host, tx); err != nil {
+				tx.Rollback()
+				log.WithError(err).Errorf("failed to refresh state of host %s cluster %s", *h.ID, params.ClusterID)
+				return installer.NewRegisterClusterInternalServerError().
+					WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+			}
+		}
+	}
+
+	if _, err = b.clusterApi.RefreshStatus(ctx, &cluster, tx); err != nil {
```
eranco74, line 686: Extract to separate function

internal/bminventory/inventory.go
```diff
@@ -581,7 +664,26 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
+				h.ID.String(), params.ClusterID).Error; err != nil {
+				tx.Rollback()
+				log.WithError(err).Errorf("failed to find host <%s> in cluster <%s>",
+					h.ID.String(), params.ClusterID)
+				return installer.NewUpdateClusterNotFound().WithPayload(common.GenerateError(http.StatusNotFound, err))
+			}
+			if _, err = b.hostApi.RefreshState(ctx, &host, tx); err != nil {
+				tx.Rollback()
+				log.WithError(err).Errorf("failed to refresh state of host %s cluster %s", *h.ID, params.ClusterID)
+				return installer.NewRegisterClusterInternalServerError().
```
eranco74, line 680: Bad error, should be NewUpdateCluster...

internal/bminventory/inventory.go
```diff
@@ -537,24 +590,54 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
+	updateIPv4 := func(target *strfmt.IPv4, source *strfmt.IPv4) {
+		if source != nil {
+			*target = *source
+		}
+	}
+	updateString := func(target *string, source *string) {
+		if source != nil {
+			*target = *source
+		}
+	}
```
ronniel1, line 607: Isn't this is what swag.String() does?

internal/bminventory/inventory.go
ronniel1, line 680: If this logic is for updating the status of all the hosts, in the cluster, maybe we should have this kind of function in cluster.go?

internal/common/machine_network_cidr.go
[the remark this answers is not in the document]
ronniel1: Maybe one can use the other?
Seems like many lines are duplicated

swagger.yaml
```diff
@@ -1120,6 +1118,12 @@ definitions:
         format: date-time
         x-go-custom-tag: gorm:"type:datetime;default:0"
         description: The time that this cluster completed installation.
+      host_networks:
```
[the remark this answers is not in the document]
filanov: unless @ori-amizur handle it with multiple tables

swagger.yaml
```diff
@@ -964,44 +956,50 @@ definitions:
...
-      dns_vip:
-        type: string
-        format: ipv4
-        description: Virtual IP used internally by the cluster for automating internal DNS requirements.
+        x-nullable: true
       ingress_vip:
         type: string
         format: ipv4
         description: Virtual IP used for cluster ingress traffic.
+        x-nullable: true
```
filanov, line 989: i think that it's true but default, unless it required.

internal/bminventory/inventory.go
ori-amizur: Already suggested how to handle ir. See https://github.com/filanov/bm-inventory/pull/186

internal/bminventory/inventory.go
```diff
@@ -435,6 +439,54 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
 			WithPayload(common.GenerateError(http.StatusNotFound, err))
 	}
 
+	cidr, err := common.CalculateMachineNetworkCIDR(&cluster)
+	if err != nil {
+		tx.Rollback()
```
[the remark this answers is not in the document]
ori-amizur: This is how the function was written in the first place. This is how I suggest to handle transactions and errors. https://github.com/filanov/bm-inventory/pull/186

internal/bminventory/inventory.go
ori-amizur: No.  This is a set function

internal/bminventory/inventory.go
```diff
@@ -537,24 +590,54 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
+	updateIPv4(&cluster.APIVip, params.ClusterUpdateParams.APIVip)
+	updateString(&cluster.BaseDNSDomain, params.ClusterUpdateParams.BaseDNSDomain)
+	updateString(&cluster.ClusterNetworkCidr, params.ClusterUpdateParams.ClusterNetworkCidr)
+	if params.ClusterUpdateParams.ClusterNetworkHostPrefix != nil {
+		cluster.ClusterNetworkHostPrefix = *params.ClusterUpdateParams.ClusterNetworkHostPrefix
+	}
+	updateIPv4(&cluster.IngressVip, params.ClusterUpdateParams.IngressVip)
+	updateString(&cluster.PullSecret, params.ClusterUpdateParams.PullSecret)
+	updateString(&cluster.ServiceNetworkCidr, params.ClusterUpdateParams.ServiceNetworkCidr)
+	updateString(&cluster.SSHPublicKey, params.ClusterUpdateParams.SSHPublicKey)
```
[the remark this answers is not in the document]
ori-amizur: It is the same as a.x = b.x but with a function to verify that only provided parameters are set

internal/bminventory/inventory.go
ori-amizur: OK

ori-amizur: We need to be consistent, because couple of lines above we update the roles of the hosts, and this is in this function.

internal/bminventory/inventory.go
ori-amizur: OK

swagger.yaml
ori-amizur: It lets you distinguish between parameters that are default value("", 0 ), and parameters that were not passed to the update command.

internal/common/machine_network_cidr.go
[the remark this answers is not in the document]
ori-amizur: It seems duplicated but it is not.
Using parts of the functions will make things more complicated (I think).
One function selects one of the vips, and tries to calculate the machine cidr from the selected vip and the ipv4-addresses of the hosts.
The other takes a cidr (usually the machine cidr) and tries to find if a host has an ip-address that belongs to that cidr.

internal/bminventory/machine_network_cidr_test.go
filanov: https://github.com/filanov/bm-inventory/blob/master/internal/host/transition_test.go#L83

internal/bminventory/inventory.go
```diff
@@ -435,6 +439,54 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
...
+
+	for _, id := range masterNodesIds {
+		if !hostIDInCidrHosts(*id, machineCidrHosts) {
+			tx.Rollback()
+			return installer.NewInstallClusterBadRequest().WithPayload(
+				logAndGenerateError(http.StatusBadRequest,
+					fmt.Sprintf("Master id %s does not have an interface with IP belonging to machine CIDR %s",
+						*id, cluster.MachineNetworkCidr)))
+		}
+	}
```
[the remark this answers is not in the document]
filanov: maybe it should be part of some pre-install validation.

internal/bminventory/inventory.go
```diff
@@ -436,6 +440,54 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
...
+				fmt.Sprintf("Could not get machine CIDR hosts for cluster %s", params.ClusterID)))
+	}
+	masterNodesIds, err := b.clusterApi.GetMasterNodesIds(ctx, &cluster, tx)
+	if err != nil {
+		tx.Rollback()
+		return installer.NewInstallClusterInternalServerError().WithPayload(
+			logAndGenerateError(http.StatusInternalServerError,
+				fmt.Sprintf("Could not get master node ids cluster %s: %s", params.ClusterID, err.Error())))
+	}
+	hostIDInCidrHosts := func(id strfmt.UUID, hosts []*models.Host) bool {
```
ronniel1, line 472: I'm guessing that this is a check that all masters belong to the correct machine CIDR.
Lets make a function out of it

internal/bminventory/inventory.go
```diff
@@ -519,9 +571,29 @@ func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, c
 	return nil
 }
 
+func (b *bareMetalInventory) refreshClusterHosts(ctx context.Context, cluster *models.Cluster, tx *gorm.DB, log logrus.FieldLogger) middleware.Responder {
```
ronniel1, line 574: Shouldn't this be under cluster.go?

internal/bminventory/inventory.go
ronniel1: Why do we need this?
swag.StringValue() will return empty string or the string value

internal/bminventory/inventory.go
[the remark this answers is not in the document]
ronniel1: Why do we care about setting, or not setting only the given values?
At least for strings

Gorm will handle this for us

internal/host/common.go
```diff
@@ -57,7 +63,7 @@ func updateStateWithParams(log logrus.FieldLogger, status, statusInfo string, h
 		return nil, errors.Wrapf(dbReply.Error, "failed to update host %s from cluster %s state from %s to %s",
 			h.ID.String(), h.ClusterID, swag.StringValue(h.Status), status)
 	}
-	if dbReply.RowsAffected == 0 {
+	if dbReply.RowsAffected == 0 && swag.StringValue(h.Status) != status {
```
ronniel1, line 66: Even if the status is the same we failed, maybe the user wanted to update some other params, so shouldn't this be an error even if the state did not change?

internal/bminventory/machine_network_cidr.go
ori-amizur: Added a comment

internal/bminventory/inventory.go
```diff
@@ -538,24 +610,49 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
 			WithPayload(common.GenerateError(http.StatusInternalServerError, errors.New("DB error, failed to start transaction")))
 	}
 
-	if err := tx.First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
+	if err = tx.Preload("Hosts").First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
```
ronniel1, line 613: If this was added for the host-networks, I don't thnk that this will work
host-networks is not an array of hosts but an array of host-networks.

internal/bminventory/machine_network_cidr_test.go
ronniel1: Let's not handle this for now

internal/host/host.go
ronniel1: changed all to refreshStatus

internal/host/common.go
ronniel1: Spoke with @filanov about this.
This is indeed buggy.
However, the entire way of handling state change in a CAS update, and adding more data on top of that is buggy, and we need to handle this differently.
This is out of scope for this review
Opened https://issues.redhat.com/browse/MGMT-982 on this issue

swagger.yaml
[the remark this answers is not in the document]
ori-amizur: Host networks are not saved in the database.  They are created on the fly.

internal/bminventory/inventory.go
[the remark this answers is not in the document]
ori-amizur: Done in https://github.com/filanov/bm-inventory/pull/195

internal/bminventory/inventory.go
ori-amizur: Will be reverted in a separate branch

internal/bminventory/inventory.go
ori-amizur: Done in https://github.com/filanov/bm-inventory/pull/195

internal/bminventory/inventory.go
ori-amizur: Yes.  Will be done in a separate branch

internal/bminventory/inventory.go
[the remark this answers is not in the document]
ori-amizur: In general I think we need to update all provided values even if they are default values.
Without such change, we will not be able to distinguish if a default value was provided by the user, or not. 
Gorm just "knows" not to update default values. It is not always the desired behaviour.
We probably need to address it separately.  This is not the best way to approach it.
Therefore, the code will be reverted.

internal/bminventory/machine_network_cidr.go
ori-amizur: OK

PR #167, opened 2020-05-26
Fix envconfig parsing for ClusterStateMonitorInterval

avishayt:
It doesn't work if the variable starts with a lower-case letter,
resulting in an interval of 0s instead of the desired 10s. This causes
zero waiting between consecutive calls which is not what we want.


filanov:
@oshercc make sure it will not make the test flaky again.

PR #168, opened 2020-05-26
MGMT-914 Add API on bm-inventory to get the admin credentials

eranco74:
WIP:
 - Adding UT
 - Adding subsystem


internal/bminventory/inventory.go
```diff
@@ -871,13 +871,55 @@ func (b *bareMetalInventory) DownloadClusterFiles(ctx context.Context, params in
...
+		if gorm.IsRecordNotFoundError(err) {
+			return installer.NewGetKubeadminPasswordNotFound().
+				WithPayload(generateError(http.StatusNotFound, err))
+		} else {
+			return installer.NewGetKubeadminPasswordInternalServerError().
+				WithPayload(generateError(http.StatusInternalServerError, err))
+		}
+	}
+	clusterStatus := swag.StringValue(cluster.Status)
+	if clusterStatus != ClusterStatusInstalling && clusterStatus != ClusterStatusInstalled {
```
filanov, line 896: i think that this validation should be part of cluster api.
otherwise there is no point in having state machine.

filanov:
@eranco74 please add at lest subsystem test for it.

internal/bminventory/inventory.go
eranco74: Unsure what you mean.

filanov: I think that all state related logic should be in one place.
So there should be an api like `IsInstalltionStarted` or something like this. that validate the states.

ronniel1: I guess that makes sense

filanov: so why resolve?

subsystem/cluster_test.go
```diff
@@ -281,6 +281,29 @@ var _ = Describe("system-test cluster install", func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(s.Size()).ShouldNot(Equal(0))
 		})
+		It("Get kubeadmin password", func() {
+			//Test happy flow
+
+			By("Test getting kubeadmin password for not found cluster")
+			{
```
filanov, line 288: you don't really need the `{`

swagger.yaml
```diff
@@ -263,6 +263,35 @@ paths:
...
+        - in: path
+          name: cluster_id
+          type: string
+          format: uuid
+          required: true
+      responses:
+        200:
+          description: Success.
+          schema:
+            type: string
```
ronniel1, line 282: @avishayt , @eranco74 , Jiri thinks that this API should also return the username and the console URL.

About the username, even if its always the same, I think that the backed should return it and not the UI have to know it.

eranco74: Well I think the UI can just use the download/file API and read the content (honestly don't see a problem with it).
But we added an API just for this specific password and we called it get kubeadmin-password.
So it only make sense that the username is `kubeadmin`.
If we want to change it, I guess we need:
1. Different name for the API.
2. Schema for the response.

@avishayt?

avishayt: GET /clusters/{cluster_id}/kubeadmin-credentials

200: {"username": <string>, "password": <string>, "console_url": <string>}

PR #172, opened 2020-05-26
Added metrics and health endpoints

ronniel1:
Metrics endpoint dumps the basic service metrics in Prometheus format


ronniel1:
> Nice! Do we need it to be under /api/assisted-install/v1/ for SRE guidelines?

I don't know about SRE guidelines, but for [openshift monitoring conventions](https://kubernetesbyexample.com/healthz/) it does not

cmd/main.go
```diff
@@ -121,3 +122,13 @@ func withMetricsMiddleware(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+func withHealthMiddleware(next http.Handler) http.Handler {
```
ori-amizur, line 126: Why not add new handler with http.HandleFunc or http.Handle ?

ronniel1: What is wrong with the middleware approach?
Adding a new handler will require another port

ori-amizur: It is more in line with the http library approach.  You let the http library do the dispatching.
Adding new handler does not require a new port

cmd/main.go
```diff
@@ -93,10 +96,39 @@ func main() {
...
+
 	h = requestid.Middleware(h)
 	if err != nil {
 		log.Fatal("Failed to init rest handler,", err)
 	}
 
 	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", swag.StringValue(port)), h))
 }
+
+func withMetricsMiddleware(next http.Handler) http.Handler {
```
filanov, line 116: suggesting to move this to /pkg
lets have a clean main.

cmd/main.go
```diff
@@ -93,10 +96,39 @@ func main() {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodGet && r.URL.Path == "/metrics" {
+			promhttp.Handler().ServeHTTP(w, r)
+			return
+		}
+		next.ServeHTTP(w, r)
+	})
+}
+
+func withHealthMiddleware(next http.Handler) http.Handler {
```
filanov, line 126: suggesting to move this to /pkg
lets have a clean main.

cmd/main.go
ronniel1: I don't agree that is is more inline.
Adding a handler is one acceptable way, and doing it this way is another
This has the benefit of not requiring another port

filanov: i like the middleware solution this way we can add more functionality if needed.

cmd/main.go
ronniel1: done

cmd/main.go
ronniel1: done

PR #174, opened 2020-05-26
Update README service update with namespace

oshercc:


ronniel1:
@oshercc I just sent a pull request with the same change :)

PR #181, opened 2020-05-27
installation can occur only when exactly 3 known masters are present

oshercc:


oshercc:
> No need to change tests?
> If so we are missing lots of tests

PR #182, opened 2020-05-27
Avishay/mgmt 656: Candidate disconnection monitoring

avishayt:


internal/host/disconnected.go
```diff
@@ -47,7 +48,11 @@ func (d *disconnectedState) UpdateRole(ctx context.Context, h *models.Host, role
 }
 
 func (d *disconnectedState) RefreshStatus(ctx context.Context, h *models.Host) (*UpdateReply, error) {
-	// State in the same state
+	log := logutil.FromContext(ctx, d.log)
+	if time.Since(time.Time(h.CheckedInAt)) < 3*time.Minute {
+		return updateState(log, HostStatusDiscovering, statusInfoDiscovering, h, d.db)
```
gamli75, line 53: are you sure that we want to return it to discovering state? If the host status was installed before we moved it to disconnect - I think that we want to return the host status to installed and not to discovering.

avishayt: The monitor only affects hosts in the following states: discovering, known, disconnected, insufficient.  You can see this in the DB query in HostMonitoring():

+       monitorStates := []string{HostStatusDiscovering, HostStatusKnown, HostStatusDisconnected, HostStatusInsufficient}
+       if err := m.db.Where("status IN (?)", monitorStates).Find(&hosts).Error; err != nil {

We no longer monitor a host in "installing" because it will reboot and disconnect anyway which is confusing.

gamli75: :+1:

swagger.yaml
```diff
@@ -766,6 +766,10 @@ definitions:
         type: string
         format: date-time
         x-go-custom-tag: gorm:"type:datetime"
+      checked_in_at:
```
filanov, line 769: need to add description?

internal/host/host.go
```diff
@@ -256,5 +257,31 @@ func (m *Manager) UpdateConnectivityReport(ctx context.Context, h *models.Host,
 		}
 	}
 	return nil
+}
+
+func (m *Manager) HostMonitoring() {
```
filanov, line 262: suggesting to move this code to a different file. `monitor.go` ?

filanov:
@avishayt what about subsystem tests?

internal/host/host.go
avishayt: Necessary?  It matches the ClusterMonitoring method in cluster.go.

avishayt:
> @avishayt what about subsystem tests?

Too complicated because the monitor runs periodically.  Added unit tests.

internal/host/host.go
filanov: I don't like it either, but if it's a problem then leave it here.

avishayt: I'll try to do it in a follow-up commit.  The only problem is that everything in Go still takes me a long time...

avishayt: OK did it

PR #184, opened 2020-05-28
MGMT-916 - remove commands when host is in disconnected state

gamli75:


avishayt:
I think the code was actually fine as is.  If the agent was disconnected and now contacts the service, we may as well ask it for the information.  It will optimize the amount of time instead of waiting for the host monitor to put it in "discovering" and waiting for the agent to contact the service again.

gamli75:
@avishayt your remark is correct in the current code when the host status will change after getting hw info data co but I'm going to change it for the next release that the host status will be changed only from the host monitor. I also think that there is an option that we will do the hw info, inventory & connectivity check twice: the agent contacts the service in disconnected state and will start the commands and after few seconds the host monitor will change the status from disconnected to discovering and the agent will perform the commands again.

I can leave it in this release but I think for the next release we will have to decrease the host monitor interval and we will not need this code.

ronniel1:
@gamli75 Can we close this PR?

PR #186, opened 2020-05-31
Suggestion using transaction and logging

ori-amizur:
Suggestion for using gorm.DB.Transaction


internal/bminventory/inventory.go
```diff
@@ -411,62 +411,54 @@ func getImageName(clusterID strfmt.UUID) string {
...
+}
 
-	if err = tx.Preload("Hosts").First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
-		return installer.NewInstallClusterNotFound().
-			WithPayload(common.GenerateError(http.StatusNotFound, err))
+func newInvetoryError(httpStatus int32, err error) *inventoryError {
+	return &inventoryError{httpStatus: httpStatus, err: err}
+}
+
+type clusterInstaller struct {
```
ronniel1, line 431: Why do we need this struct?
Why not pass this to the install method?

internal/bminventory/inventory.go
```diff
@@ -479,46 +471,69 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
...
+		}
+	}()
+	err = b.db.Transaction(clusterInstaller{
+		ctx:    ctx,
+		b:      b,
+		params: params,
+	}.install)
+
+	if err != nil {
+		log.WithError(err).Warn("Cluster install")
```
ronniel1, line 519: Where is the rollback being done?

filanov: +1

internal/bminventory/inventory.go
```diff
@@ -411,62 +411,54 @@ func getImageName(clusterID strfmt.UUID) string {
 	return fmt.Sprintf("discovery-image-%s", clusterID.String())
 }
 
-func logAndGenerateError(id int32, errStr string) *models.Error {
-	logrus.Error(errStr)
-	return common.GenerateError(id, errors.New(errStr))
+type inventoryError struct {
```
filanov, line 414: please move to a different package or file.

internal/bminventory/inventory.go
filanov: +1

internal/bminventory/inventory.go
ori-amizur: Look at https://gorm.io/docs/transactions.html
If non-nil error is returned, then it is rollback
If nil commit

internal/bminventory/inventory.go
ori-amizur: OK

internal/bminventory/inventory.go
ori-amizur: The signature of the function that is passed to the gorm.DB.Transaction is 
func(tx *gorm.DB) error.  If I want to pass additional parameters, then I need additional struct.
We can add our own Transaction function with our parameters, and maybe our response value that may be middleware.Responder, but it is a different approach.

internal/bminventory/inventory.go
ori-amizur: @ronniel1 This PR is not up-to-date.
Please look at https://github.com/filanov/bm-inventory/pull/195.
Sometimes, like in install-cluster you want to do post commit transactions.
Then you cannot rely on commit being delayed.
Besides, commit may fail - deadlocks or some other reason.  You will not be able to catch that.

ronniel1: You can remove the commit from the defer and do it where ever you want.
The important part is that the rollback is in the defer

filanov: i think that if we have a solution that don't require a struct that need to implement something it will be better.

ori-amizur: I think it is preferable to have it as a reusable function that performs the begin, rollback, and commit.  If the function is tested, then there will not be a case that the user forgets to commit or to rollback.
This method also lets the user log the error in a single place.

ronniel1: That is a good point.
However the code is still much less understandable.
To prove the point, how long it took both @filanov and me to understand what you did there.
So while we might not be the brightest, we are the kind of people who need to consume this

ori-amizur:
I think you didn't expect it.
After you are aware of it, how long will it take you to understand a
similar change with the same pattern?

internal/bminventory/inventory.go
ori-amizur: I think you didn't expect it.
After you are aware of it, how long will it take you to understand a
similar change with the same pattern?

PR #188, opened 2020-05-31
MGMT-832 Assisted-Installer service hosting for beta

filanov:
Support ui deployment.
The service creates s3 bucket, not the deployment.
Add KUBECONFIG evn and mount.


cmd/main.go
```diff
@@ -6,29 +6,30 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/aws/aws-sdk-go/service/s3"
```
ronniel1, line 9: is this goimport-ed?

cmd/main.go
```diff
@@ -57,6 +59,17 @@ func main() {
...
+	if err != nil {
+		log.Fatal(errors.Wrap(err, "failed to create s3 client"))
+	}
+
+	if _, err = s3client.CreateBucket(&s3.CreateBucketInput{
+		Bucket: swag.String(Options.S3Config.S3Bucket),
+	}); err != nil {
+		log.Fatal(errors.Wrap(err, "failed to create bucket"))
+	}
+
```
ronniel1, line 72: Maybe this should be in a function? to make main shorter?

deploy/s3/scality-configmap.yaml
```diff
@@ -7,6 +7,6 @@ metadata:
     app: scality
 data:
   S3_ENDPOINT_URL: REPLACE_URL
-  HOST_NAME: REPLACE_HOST_NAME
+#  HOST_NAME: REPLACE_HOST_NAME
```
ronniel1, line 10: if not needed, please remove

deploy/s3/scality-deployment.yaml
```diff
@@ -15,6 +15,7 @@ spec:
       labels:
         app: scality
     spec:
+      hostname: cloudserver-front
```
ronniel1, line 18: Maybe worth a comment

deploy/s3/scality-deployment.yaml
```diff
@@ -35,7 +36,7 @@ kind: Service
 metadata:
   labels:
     app: scality
-  name: scality
+  name: cloudserver-front
```
ronniel1, line 39: Why is the name changed as well?
We have multiple places where 'scality' as the name of the service is referenced. Do we need to change them all?

deploy/s3/scality-ingress.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Ingress
```
ronniel1, line 1: Why do we have ingress for scality?
Aren't we accessing it only from withing the cluster?

tools/deploy_scality_configmap.py
```diff
@@ -1,20 +1,24 @@
 import os
-import sys
 import utils
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--target")
+
+args = parser.parse_args()
+
```
ronniel1, line 9: Where is that used?

tools/deploy_s3.py
```diff
@@ -1,14 +1,20 @@
 import os
-import sys
 import utils
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--target")
+
+args = parser.parse_args()
+
```
ronniel1, line 9: Where is that used?

cmd/main.go
filanov: what does it mean?

deploy/s3/scality-configmap.yaml
filanov: it will remind us about the hostname that need to be set if we are working with load-balancer or other way to access scality. 
but i can remove it.

deploy/s3/scality-deployment.yaml
filanov: no, just the service. 
when defining a service it automatically added to the dns

deploy/s3/scality-ingress.yaml
filanov: leftover - will remove

cmd/main.go
filanov: I was planning to use this client for other operations in the service but it we can do it later. 
So i will add a single function.

tools/deploy_scality_configmap.py
filanov: will remove.

cmd/main.go
ronniel1: looks like this is before running the tool goimport

deploy/s3/scality-deployment.yaml
filanov: added comment above.

tools/deploy_s3.py
filanov: removed

deploy/s3/scality-deployment.yaml
ronniel1: Don't see the comment

filanov: sorry probably deleted in the rebase. 
added again.

filanov: Sorry i don't think i need it. 
i added the comment to the service.

filanov: yep don't need it. will remove.

filanov:
tested with test-infra

PR #191, opened 2020-05-31
MGMT-970 - update status_updated_at attributes for host and cluster

ronniel1:


internal/cluster/common.go
```diff
@@ -36,7 +39,8 @@ type baseState struct {
 }
 
 func updateState(state string, statusInfo string, c *models.Cluster, db *gorm.DB, log logrus.FieldLogger) (*UpdateReply, error) {
-	updates := map[string]interface{}{"status": state, "status_info": statusInfo}
+	fmt.Printf("Update time: %s\n", strfmt.DateTime(time.Now()).String())
```
filanov, line 42: was it for debug?

ronniel1: yes
Will remove

PR #193, opened 2020-05-31
Igal/ingress ca

tsorya:


internal/bminventory/inventory.go
```diff
@@ -1114,14 +1114,24 @@ func (b *bareMetalInventory) UploadClusterIngressCert(ctx context.Context, param
 		msg := fmt.Sprintf("Cluster %s is in %s state, upload ingress ca can be done only in installed state", params.ClusterID, clusterStatus)
 		log.Warn(msg)
 		return installer.NewUploadClusterIngressCertBadRequest().
-			WithPayload(common.GenerateError(http.StatusConflict, errors.New(msg)))
+			WithPayload(common.GenerateError(http.StatusBadRequest, errors.New(msg)))
```
filanov, line 1117: i think that usually when there is something that the user can do to change the state that allow him to perform this operation then the reply is conflict. 
@avishayt probably know best

internal/bminventory/inventory.go
```diff
@@ -1114,14 +1114,24 @@ func (b *bareMetalInventory) UploadClusterIngressCert(ctx context.Context, param
...
-	resp, err := b.s3Client.DownloadFileFromS3(ctx, fileName, b.S3Bucket)
+	exists, err := b.s3Client.DoesObjectExists(ctx, fileName, b.S3Bucket)
+	if err != nil {
+		return installer.NewUploadClusterIngressCertInternalServerError().
+			WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+	}
+
+	if exists {
+		log.Warnf("Ingress ca for cluster %s already exists, returning bad request", cluster.ID)
+		return installer.NewUploadClusterIngressCertBadRequest().
```
filanov, line 1129: is that a problem? we don't need the option to overwrite?

internal/bminventory/inventory_test.go
```diff
@@ -579,12 +583,34 @@ var _ = Describe("UploadClusterIngressCert test", func() {
...
+		})
+		Expect(generateReply).Should(BeAssignableToTypeOf(installer.NewUploadClusterIngressCertInternalServerError()))
+	})
 	It("UploadClusterIngressCert s3download failure", func() {
 		status := ClusterStatusInstalled
 		cluster.Status = &status
 		db.Save(&cluster)
-		fileName := fmt.Sprintf("%s/%s", clusterID, kubeconfig)
-		mockS3Client.EXPECT().DownloadFileFromS3(ctx, fileName, "test").Return(nil, errors.Errorf("dummy"))
+		mockS3Client.EXPECT().DoesObjectExists(ctx, kubeconfigObject, "test").Return(false, nil).Times(1)
```
filanov, line 612: if you have the same mock over and over again i suggest to make a single function that set it.
it can be an unnamed function. there are multiple examples in this file.

pkg/s3Client/s3Client.go
```diff
@@ -56,3 +57,18 @@ func (s s3Client) DownloadFileFromS3(ctx context.Context, fileName string, s3Buc
...
+func (s s3Client) DoesObjectExists(ctx context.Context, objectName string, s3Bucket string) (bool, error) {
+	log := logutil.FromContext(ctx, s.log)
+	log.Infof("Verifying if %s exists in %s", objectName, s3Bucket)
+	_, err := s.client.StatObject(s3Bucket, objectName, minio.StatObjectOptions{})
+	if err != nil {
+		errResponse := minio.ToErrorResponse(err)
+		if errResponse.Code == "NoSuchKey" {
+			return false, nil
+		}
+		log.WithError(err).Infof("Failed to get %s from %s", objectName, s3Bucket)
```
filanov, line 70: we  already talked about error logs. should be once in the api level.

subsystem/cluster_test.go
```diff
@@ -363,6 +363,10 @@ var _ = Describe("system-test cluster install", func() {
 
 				Expect(s.Size()).ShouldNot(Equal(sni.Size()))
 
+				// Try to upload ingress ca second time, must fail
+				_, err = bmclient.Installer.UploadClusterIngressCert(ctx, &installer.UploadClusterIngressCertParams{ClusterID: clusterID, IngressCertParams: models.IngressCertParams(ingressCa)})
```
filanov, line 367: if you have a long test i suggest to use `By` to mark multiple steps. it will be easier to look and at the test and debug it.

internal/bminventory/inventory.go
tsorya: nope, surely it will not overwrite it, it will just add it. 
This operation is one time thing

internal/bminventory/inventory.go
tsorya: @avishayt ?

pkg/s3Client/s3Client.go
tsorya: logger is passed through context, so why it matters?

filanov: because we will have the same log multiple times. 
imagine looking at a log like this:
```
Failed to get %s from %s
Failed to get %s from %s
Failed to get %s from %s
Failed to get %s from %s
Failed to get %s from %s
```
if you have extra information to add you can add it to the error with `errorrs.Wrap`

tsorya: not sure i understand why it is different but i can move it

filanov: if we will have an error log in every level in the code it will make the logs messy and it will repeat itself.
you don't have to remove it if you don't want to.
but eventually it will be hard to look at the logs like this.

tsorya: thats why i didn't add log inside api but i will move it. 
I am adding logs inside functions cause api func is very long, every log line makes it even longer and less readable

filanov: but if you have internal log then anyone that adding new code will need to know that you have a log somewhere in the code. if you have it in the api level it will be only in one place.

internal/bminventory/inventory.go
eranco74, line 1129: Why return an error?
Just log "nothing to do".

PR #194, opened 2020-06-01
Eran/mgmt 981

eranco74:


subsystem/cluster_test.go
```diff
@@ -109,11 +109,68 @@ var _ = Describe("Cluster tests", func() {
 	})
 })
 
+func waitForClusterState(ctx context.Context, clusterID strfmt.UUID, state string) {
```
filanov, line 112: i really don''t like it. i wrote it to @oshercc as well.
we probably should use something like https://onsi.github.io/gomega/#eventually

subsystem/cluster_test.go
```diff
@@ -620,57 +639,118 @@ var _ = Describe("system-test cluster install", func() {
 	})
 })
 
-func waitForClusterState(ctx context.Context, clusterID strfmt.UUID, state string) {
-	for start := time.Now(); time.Since(start) < 10*time.Second; {
+var _ = Describe("system-test cluster install, with default network params", func() {
```
filanov, line 642: you don't need this part `system-test `

subsystem/cluster_test.go
eranco74: Didn't touch it. it just GitHub diff acting out

PR #197, opened 2020-06-01
Use Eventually instead of for loop

eranco74:
Ran 21 of 21 Specs in 93.678 seconds
SUCCESS! -- 21 Passed | 0 Failed | 0 Pending | 0 Skipped


filanov:
much better!

PR #198, opened 2020-06-01
Osher/mgmt 984

oshercc:
MGMT-984 workaround for the first ISO download timeout, assisted-installer should create a dummy ISO


filanov:
@ronniel1 @oshercc how about instead of doing something like this we can create a dummy job as part of the deployment?

internal/bminventory/inventory.go
```diff
@@ -144,11 +144,22 @@ func NewBareMetalInventory(
 	if cfg.ImageBuilderCmd != "" {
 		b.imageBuildCmd = strings.Split(cfg.ImageBuilderCmd, " ")
 	}
+	//	Run first ISO dummy for image pull
```
filanov, line 147: please explain here the it is done so that the image will be pulled and the api will take less time.

PR #200, opened 2020-06-01
Revert "Use Eventually instead of for loop"

ronniel1:
This reverts commit de5577cddf46b0b63e0c501c28de53b0c5e7861c.


filanov:
why did you revert it?

ronniel1:
It caused the subsystem to fail

PR #205, opened 2020-06-02
Avishay/mgmt 1001

avishayt:


avishayt:
Subsystem failing - do not review yet

ronniel1:
Also need to test in test-infa

internal/host/common.go
```diff
@@ -49,7 +49,16 @@ func updateByKeepAlive(log logrus.FieldLogger, h *models.Host, db *gorm.DB) (*Up
...
+	updates := map[string]interface{}{}
+	now := strfmt.DateTime(time.Now())
+	if swag.StringValue(h.Status) != status {
+		updates["status"] = status
+		updates["status_updated_at"] = now
+	}
+	if *h.StatusInfo != statusInfo {
+		updates["status_info"] = statusInfo
+		updates["status_updated_at"] = now
+	}
```
ronniel1, line 61: I don't understand how this solves things.
In case the status does not change, but the status_info does, we will update the DB and this can cause a parallel install transaction to fail.

What am I missing?

filanov: i suggest to have specific update function for the monitor. at least until we make the transaction shorter.

avishayt: This solves things because we won't move from "known" to "known".

PR #206, opened 2020-06-02
MGMT-832 Set new default role, service account and role binding that …

filanov:
…enable basic operations for non root user


filanov:
Tested subsystem and test-infra.

PR #208, opened 2020-06-03
Preparing the Role and service to be fully compatible with Prometheus monitoring

jparrill:
Those changes are necesary in order to add the monitoring part on the Minikube side.
After these ones, you need to follow this steps:

- Deploy OLM
- Deploy Prometheus operator
- Deploy Prometheus instance
- Create the service of the Assisted Installer for Prometheus
- Create the ServiceMonitor

With that you will be able to see the monitoring screen on your cluster using a port-forward

### DIFF
--- a/deploy/bm-inventory-service.yaml
+++ b/deploy/bm-inventory-service.yaml
@@ -7,7 +7,8 @@ metadata:
   namespace: assisted-installer
 spec:
   ports:
-    - port: 8090
+    - name: bm-inventory
+      port: 8090
       protocol: TCP
       targetPort: 8090
   selector:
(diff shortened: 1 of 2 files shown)


filanov:
is it tested with test-infra?

jparrill:
It's not a destructive change, it just ask for more permissions on the Role creation and gives a name for the port in the svc case. Should work in the same way.

romfreiman:
@jparrill just verifying that it will run on ocp with all the restrictions there

jparrill:
Humm It should, this seems like an error from the Jenkins api token against GitHub, could be?

jparrill:
is this a permissions issue on the destination CI/CD platform?

filanov:
@jparrill i added you as a contributor. lets see if it helps.

PR #211, opened 2020-06-03
MGMT-996 - Validate the format of the pull-secret

ronniel1:


internal/bminventory/inventory.go
```diff
@@ -279,7 +281,15 @@ func (b *bareMetalInventory) RegisterCluster(ctx context.Context, params install
...
 	}
-	setPullSecret(&cluster, params.NewClusterParams.PullSecret)
+	if params.NewClusterParams.PullSecret != "" {
+		err := validations.ValidatePullSecret(params.NewClusterParams.PullSecret)
+		if err != nil {
+			log.WithError(err).Errorf("Pull-secret for new cluster has invalid format")
+			return installer.NewRegisterClusterBadRequest().
+				WithPayload(common.GenerateError(http.StatusBadRequest, errors.New("Pull-secret has invalid format")))
+		}
+		setPullSecret(&cluster, params.NewClusterParams.PullSecret)
```
avishayt, line 291: You changed the behavior here - now if pull secret is empty the old one won't be deleted.  Is that on purpose?

ronniel1:
Actually, you are not right,.
This is only in RegisterCluster

avishayt:
> Actually, you are not right,. This is only in RegisterCluster

Yes I also realized it a second later - that's why I deleted the comment :)

PR #212, opened 2020-06-03
Moved the logic that check the cluster status when trying to download…

eranco74:
… a file to clustr.go


eranco74:
Opening a different PR to master

PR #213, opened 2020-06-03
Moved the logic that check the cluster status when trying to download…

eranco74:
… a file or get credentials to clustrApi

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -37,7 +37,6 @@ import (
 	"github.com/jinzhu/gorm"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"github.com/thoas/go-funk"
 	batch "k8s.io/api/batch/v1"
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


tsorya:
To say the truth it doesn't look good. Not sure why moving this part is so crucial, i think it looks even less readable than now and much more complicated.
Not better just to add function :
validate_cluster_state(cluster.., allowed_states) ?

eranco74:
> validate_cluster_state(cluster.., allowed_states) ?

I agree, but than the inventory needs to know (as it did till now) what are the allowed statuses for each action.
@filanov ?

filanov:
@eranco74 is that just the first step? because it's strange that get function return just an error.

filanov:
if it is the first step then please add a TODO in the API or somewhere so it will be clear.

filanov:
> To say the truth it doesn't look good. Not sure why moving this part is so crucial, i think it looks even less readable than now and much more complicated.
> Not better just to add function :
> validate_cluster_state(cluster.., allowed_states) ?

As @eranco74 said, cluster status is something internal to the cluster. so it should be handled by the cluster.

eranco74:
Next step:
https://issues.redhat.com/browse/MGMT-1031

PR #217, opened 2020-06-04
Fix db transaction race

filanov:
We still have two places related to cluster update and download image. 
but it's less urgent so it will be fixed in a different PR
This is the fix:
```
if err := tx.Model(&host).Update("checked_in_at", host.CheckedInAt).Error; err != nil {
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -72,6 +72,9 @@ func main() {
 		log.Fatal("Fail to connect to DB, ", err)
 	}
 	defer db.Close()
+	db.DB().SetMaxIdleConns(0)
+	db.DB().SetMaxOpenConns(0)
+	db.DB().SetConnMaxLifetime(0)
 
 	scheme := runtime.NewScheme()
 	if err = clientgoscheme.AddToScheme(scheme); err != nil {
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -891,8 +891,7 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params installer.
 	}
 
 	host.CheckedInAt = strfmt.DateTime(time.Now())
-
-	if err := tx.Model(&host).Update(host).Error; err != nil {
+	if err := tx.Model(&host).Update("checked_in_at", host.CheckedInAt).Error; err != nil {
 		tx.Rollback()
 		log.WithError(err).Errorf("failed to update host: %s", params.ClusterID)
 		return installer.NewGetNextStepsInternalServerError()
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -418,7 +418,7 @@ func (b *bareMetalInventory) GenerateClusterISO(ctx context.Context, params inst
 	// This job name is exactly 63 characters which is the maximum for a job - be careful if modifying
 	jobName := fmt.Sprintf("createimage-%s-%s", cluster.ID, now.Format("20060102150405"))
 	imgName := getImageName(params.ClusterID)
-	log.Info("Creating job %s", jobName)
+	log.Infof("Creating job %s", jobName)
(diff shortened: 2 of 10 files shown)


filanov:
subsystem and test-infra passed.
will run test-infra multiple times.

internal/cluster/installing.go
```diff
@@ -47,10 +47,10 @@ func (i *installingState) RefreshStatus(ctx context.Context, c *models.Cluster,
 	return nil, errors.Errorf("cluster % state transaction is not clear, installation state: %s ", c.ID, installationState)
 }
 
-func (i *installingState) getClusterInstallationState(ctx context.Context, c *models.Cluster) (string, string, error) {
+func (i *installingState) getClusterInstallationState(ctx context.Context, c *models.Cluster, db *gorm.DB) (string, string, error) {
 	log := logutil.FromContext(ctx, i.log)
 
-	if err := i.db.Preload("Hosts").First(&c, "id = ?", c.ID).Error; err != nil {
+	if err := db.Preload("Hosts").First(&c, "id = ?", c.ID).Error; err != nil {
```
eranco74, line 53: Why not use the installer db?

filanov: Because refresh can run as part of transaction, for example in cluster update.
It's not changing anything right now but if this code will move to be as part of a transaction we will have a bug.
so it's just something that will prevent it in the future.

cmd/main.go
```diff
@@ -72,6 +72,9 @@ func main() {
 		log.Fatal("Fail to connect to DB, ", err)
 	}
 	defer db.Close()
+	db.DB().SetMaxIdleConns(0)
```
ori-amizur, line 75: Why?

internal/cluster/installing.go
```diff
@@ -28,7 +28,7 @@ var _ StateAPI = (*Manager)(nil)
 
 func (i *installingState) RefreshStatus(ctx context.Context, c *models.Cluster, db *gorm.DB) (*UpdateReply, error) {
 	log := logutil.FromContext(ctx, i.log)
-	installationState, StateInfo, err := i.getClusterInstallationState(ctx, c)
+	installationState, StateInfo, err := i.getClusterInstallationState(ctx, c, db)
```
ori-amizur, line 31: Couple of lines below (38, 40) , you are still using i.db instead db.  It seems to me that it has to change there as well

internal/bminventory/inventory.go
```diff
@@ -890,8 +891,7 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params installer.
 	}
 
 	host.CheckedInAt = strfmt.DateTime(time.Now())
-
-	if err := tx.Model(&host).Update(host).Error; err != nil {
+	if err := tx.Model(&host).Update("checked_in_at", host.CheckedInAt).Error; err != nil {
```
ori-amizur, line 894: Q.  Update also modifies updated_at column.  Maybe you want to use UpdateColumn instead?

cmd/main.go
filanov: we had in strato i don't remember all the details but it solves some issues for us.

internal/bminventory/inventory.go
filanov: why is that a problem to update updated_at as well?

cmd/main.go
ori-amizur: Seems like it disables connection pooling.

internal/bminventory/inventory.go
ori-amizur: No problem.  Just, your correction was to update a specific field.  Maybe you want to be more specific.

filanov: updated_at is a field managed by gorm. we want to know when the object host was last updated , this is the only purpose of this field.

cmd/main.go
filanov: is that a problem for us? not sure if we use it.

PR #222, opened 2020-06-04
MGMT-779 Prevent host registration after installation has started - WIP

eranco74:


filanov:
@eranco74 it's already handled by the state machine. 
https://github.com/filanov/bm-inventory/blob/master/internal/host/statemachine.go#L12

internal/cluster/cluster.go
```diff
@@ -199,3 +200,12 @@ func (m *Manager) UploadIngressCert(c *models.Cluster) (err error) {
 
 	return err
 }
+
+func (m *Manager) VerifyRegisterHost(c *models.Cluster) (err error) {
+	clusterStatus := swag.StringValue(c.Status)
+	allowedStatuses := []string{clusterStatusInsufficient, clusterStatusReady}
+	if !funk.ContainsString(allowedStatuses, clusterStatus) {
+		err = fmt.Errorf("Cluster %s is in %s state, host can register only in one of %s", c.ID, clusterStatus, allowedStatuses)
```
filanov, line 208: please use `github.com/pkg/errors` and not fmt. 
this package give you a stack trace and we use it as a standard error package.

eranco74:
> @eranco74 it's already handled by the state machine.
> https://github.com/filanov/bm-inventory/blob/master/internal/host/statemachine.go#L12

This is the cluster state machine...

internal/bminventory/inventory.go
```diff
@@ -790,14 +790,19 @@ func (b *bareMetalInventory) GetCluster(ctx context.Context, params installer.Ge
...
 	log.Infof("Register host: %+v", params)
 
-	if err := b.db.First(&models.Cluster{}, "id = ?", params.ClusterID.String()).Error; err != nil {
+	if err := b.db.First(&cluster, "id = ?", params.ClusterID.String()).Error; err != nil {
 		log.WithError(err).Errorf("failed to get cluster: %s", params.ClusterID.String())
 		return installer.NewRegisterHostBadRequest().
 			WithPayload(common.GenerateError(http.StatusBadRequest, err))
 	}
+	if err := b.clusterApi.VerifyRegisterHost(&cluster); err != nil {
+		return installer.NewRegisterHostBadRequest().
```
filanov, line 802: the user will not see the issue, need to add a log. maybe even event? @ronniel1?

internal/cluster/cluster_test.go
```diff
@@ -286,6 +287,41 @@ var _ = Describe("cluster monitor", func() {
...
+	)
+
+	BeforeEach(func() {
+		db = prepareDB()
+		id = strfmt.UUID(uuid.New().String())
+		clusterApi = NewManager(getTestLog().WithField("pkg", "cluster-monitor"), db, nil)
+	})
+
+	checkVerifyRegisterHost := func(clusterStatus string, expectedErr error) {
+		Context(fmt.Sprintf("Register host while cluster in %s state", clusterStatus), func() {
```
filanov, line 307: i think that all tests should be under `It` but not sure.

filanov:
> > @eranco74 it's already handled by the state machine.
> > https://github.com/filanov/bm-inventory/blob/master/internal/host/statemachine.go#L12
> 
> This is the cluster state machine...

right sorry.

internal/cluster/cluster_test.go
```diff
@@ -286,6 +287,41 @@ var _ = Describe("cluster monitor", func() {
...
+			c = models.Cluster{ID: &id, Status: swag.String("ready")}
+			Expect(db.Create(&c).Error).ShouldNot(HaveOccurred())
+			Expect(err).ShouldNot(HaveOccurred())
```
filanov, line 310: from where is the error that you compare?

internal/cluster/cluster_test.go
```diff
@@ -286,6 +287,41 @@ var _ = Describe("cluster monitor", func() {
...
+		clusterApi *Manager
+	)
+
+	BeforeEach(func() {
+		db = prepareDB()
+		id = strfmt.UUID(uuid.New().String())
+		clusterApi = NewManager(getTestLog().WithField("pkg", "cluster-monitor"), db, nil)
+	})
+
+	checkVerifyRegisterHost := func(clusterStatus string, expectedErr error) {
```
filanov, line 306: i think if will be easier to understand with a table driven test, but it's fine like this as well

internal/bminventory/inventory.go
```diff
@@ -790,14 +790,20 @@ func (b *bareMetalInventory) GetCluster(ctx context.Context, params installer.Ge
...
 	log.Infof("Register host: %+v", params)
 
-	if err := b.db.First(&models.Cluster{}, "id = ?", params.ClusterID.String()).Error; err != nil {
+	if err := b.db.First(&cluster, "id = ?", params.ClusterID.String()).Error; err != nil {
 		log.WithError(err).Errorf("failed to get cluster: %s", params.ClusterID.String())
 		return installer.NewRegisterHostBadRequest().
 			WithPayload(common.GenerateError(http.StatusBadRequest, err))
 	}
-
+	if err := b.clusterApi.VerifyRegisterHost(&cluster); err != nil {
```
filanov, line 801: when i think about it. maybe we do want to register the host but in error state or some state that will ignore the this host?

avishayt: I think it might be weird to have new hosts show up in the list after installation already started.  I would return 409 to the agent which should cause it to stop trying to register, and raise an event.

filanov: What about the user? He will not understand why one of the hosts that he booted don't appear

avishayt: I assumed that the user will wait for all of the hosts that he expected before clicking "install", and that any hosts that show up later are probably by mistake (e.g., someone booted the wrong ISO).

eranco74: I agree.
In addition we need the agent to log the error and stop trying to register.

filanov: In small clusters i agree. But in large cluster it will be an issue. 
but we probably support mostly small cluster. 
So if @avishayt agree i don't have objections.
Still if the user will not know the reason i think that we will be called to debug.

avishayt: Yes agree, when the agent sees 409 it should stop trying.  Retry should only be on 500 errors.

eranco74: https://issues.redhat.com/browse/MGMT-1041

eranco74:
Ran 22 of 22 Specs in 103.968 seconds
SUCCESS! -- 22 Passed | 0 Failed | 0 Pending | 0 Skipped

PR #224, opened 2020-06-06
Fix flaky test, (download_config_files in error state)

eranco74:
The test fail in case the node that fail to install is the worker.
Updated the test to fail a master node


eranco74:
Ran 22 of 22 Specs in 163.624 seconds
SUCCESS! -- 22 Passed | 0 Failed | 0 Pending | 0 Skipped

PR #227, opened 2020-06-07
MGMT-1042 - There are no logs in the service when we deregister a clu…

gamli75:
…ster/ host
+ Added logs to other methods

@ronniel1 @filanov please review


filanov:
> Is there a way to add logs to the start and end of each non-GET API call in the infra, including parameters and return codes (rather than doing it manually)?

We can create a wrapper to the api that will log everything before and after the api, but it will have less info then we can set in the api. 
@ronniel1 what do you think does it worth adding something like this?

ronniel1:
> > Is there a way to add logs to the start and end of each non-GET API call in the infra, including parameters and return codes (rather than doing it manually)?
> 
> We can create a wrapper to the api that will log everything before and after the api, but it will have less info then we can set in the api.
> @ronniel1 what do you think does it worth adding something like this?

I think that since we will not have all the relevant info, it's better to have it in the API level and not as a wrapper

PR #229, opened 2020-06-07
Return Forbidden in case a host tries to register while cluster is already installing

eranco74:


filanov:
how about adding some tests?

PR #230, opened 2020-06-07
Igal/hostnames

tsorya:


subsystem/cluster_test.go
```diff
@@ -663,6 +680,72 @@ var _ = Describe("cluster install", func() {
 		Expect(err).NotTo(HaveOccurred())
 		h1 = getHost(clusterID, *h1.ID)
 		Expect(*h1.Status).Should(Equal("insufficient"))
+
+	})
+	It("unique hostname validation", func() {
+		clusterID := *cluster.ID
+
+		hwInfo := &models.Inventory{
```
filanov, line 688: there is a variable with this value. `validHwInfo`

internal/bminventory/inventory.go
```diff
@@ -503,6 +503,24 @@ func (c *clusterInstaller) installHosts(cluster *models.Cluster, tx *gorm.DB) er
 	return nil
 }
 
+func (c clusterInstaller) validateHostsInventory(cluster *models.Cluster) error {
+	for _, chost := range cluster.Hosts {
+		sufficient, err := c.b.hostApi.ValidateCurrentInventory(chost, cluster)
```
filanov, line 508: you are running inside a transaction, need to make sure that you are working with the transaction db.

subsystem/cluster_test.go
```diff
@@ -663,6 +680,72 @@ var _ = Describe("cluster install", func() {
 		Expect(err).NotTo(HaveOccurred())
 		h1 = getHost(clusterID, *h1.ID)
 		Expect(*h1.Status).Should(Equal("insufficient"))
+
+	})
+	It("unique hostname validation", func() {
```
filanov, line 685: if you have a code that repeat itself please don't copy it. just put it under the same context inside a before each

tsorya: can u specify what specific code?

internal/bminventory/inventory.go
ronniel1: nice catch @filanov

ronniel1: Why do we need this validation inside the transaction?
Already the transaction is too long

tsorya: @ronniel1 we don't need it to much but this is what was decided. i can delete it

ronniel1: I don't mean that we don't need the validation.
We do need it but I don't think it should be done within the transaction, but as a validation done before starting the transaction.

Or am I misunderstanding what you mean @tsorya ?

subsystem/cluster_test.go
filanov: https://github.com/filanov/bm-inventory/blob/master/subsystem/cluster_test.go#L485

internal/bminventory/inventory.go
```diff
@@ -503,6 +503,24 @@ func (c *clusterInstaller) installHosts(cluster *models.Cluster, tx *gorm.DB) er
...
+		if err != nil {
+			msg := fmt.Sprintf("failed to validate host <%s> in cluster: %s", chost.ID.String(), cluster.ID.String())
+			c.b.log.WithError(err).Warn(msg)
```
ori-amizur, line 511: You don't need the log.  The error is logged in the InstallCluster.  You just need to format the error with errors.Wrapf

internal/host/common.go
```diff
@@ -114,6 +114,9 @@ func getCluster(clusterID strfmt.UUID, db *gorm.DB) (*models.Cluster, error) {
 	if err := db.Take(&cluster, "id = ?", clusterID.String()).Error; err != nil {
 		return nil, err
 	}
+	if err := db.Preload("Hosts").First(&cluster, "id = ?", clusterID).Error; err != nil {
```
ori-amizur, line 117: Why query cluster twice? Take is the same as First but without sorting

internal/bminventory/inventory.go
```diff
@@ -503,6 +503,24 @@ func (c *clusterInstaller) installHosts(cluster *models.Cluster, tx *gorm.DB) er
...
+			return common.NewApiError(http.StatusInternalServerError, errors.Wrapf(err, msg))
+		}
+		if !sufficient.IsSufficient {
+			msg := fmt.Sprintf("host <%s>  failed to pass hardware validation in cluster: %s. Reason %s",
+				chost.ID.String(), cluster.ID.String(), sufficient.Reason)
+			c.b.log.Warn(msg)
```
ori-amizur, line 517: Same as above

internal/bminventory/inventory.go
ori-amizur: It seems that you need to do it in the transaction because the cluster and host information might be different if you do it outside.
Seems that ValidateCurrentInventory doesn't need to query the database because it gets both the cluster and host.  It probably doesn't need anything more.   @tsorya is it true?

tsorya: yap, i can do it before transaction, it will require preload of hosts though

ori-amizur: They are already preloaded in the start of the transaction

internal/bminventory/inventory.go
```diff
@@ -538,19 +556,24 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
...
 
+	cInstaller := clusterInstaller{
+		ctx:    ctx,
+		b:      b,
+		params: params,
+	}
+	if err = b.db.Preload("Hosts").First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
+		return common.NewApiError(http.StatusNotFound, err)
+	}
+	if err = cInstaller.validateHostsInventory(&cluster); err != nil {
```
ori-amizur, line 567: Why outside the transaction?  We do the same query in the transaction.  What is the benefit?

tsorya: @ori-amizur to make transaction shorter

tsorya: @ronniel1

internal/bminventory/inventory.go
tsorya: don't see where it will e logged.

internal/bminventory/inventory.go
ori-amizur: Transaction length is measured by the number of queries and the queries themselves.
Here the transaction length remains the same.

ronniel1: @ori-amizur this is not entirely true. Since transactions are serialized, the time that a transaction takes is also important

internal/bminventory/inventory.go
ronniel1: The hostnames do not change often so doing it in a transaction is not really important

internal/bminventory/inventory.go
ori-amizur: Transactions are serialized only in case of conflict.  Conflicts happen when you update.  Since we are talking about the first query in transaction, and you do not perform select for update, then you will not have any conflict event between transactions.

internal/bminventory/inventory.go
ori-amizur: @ronniel1 But here you query the database twice instead of once. 
In the install we also verify vips in the transaction.  Should we also move it outside the transaction?
Sane is true for the UpdateCluster

internal/bminventory/inventory.go
ori-amizur: You are right.
Still error can be returned and logged once in the caller instead of twice here.

internal/bminventory/inventory.go
filanov: i'm not sure why are you arguing, all the validation are done at the begging of the transaction, it takes probably less than 100 milliseconds. if something fails we close the transaction.

ori-amizur: Less and simpler code.  I prefer not to add code that is not needed.

internal/bminventory/inventory.go
ronniel1: That is not what we are seeing

PR #231, opened 2020-06-07
Osher/mgmt 1038 - Should not allow to update hosts when the installation has started

oshercc:
MGMT-1038 prevent update while cluster is installing

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -691,6 +691,12 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
 		log.WithError(err).Errorf("failed to get cluster: %s", params.ClusterID)
 		return installer.NewUpdateClusterNotFound().WithPayload(common.GenerateError(http.StatusNotFound, err))
 	}
+
+	if err = b.clusterApi.VerifyClusterUpdatability(&cluster); err != nil {
+		log.WithError(err).Errorf("cluster %s can't be updated in current state", params.ClusterID)
+		return installer.NewUpdateClusterConflict().WithPayload(common.GenerateError(http.StatusConflict, err))
+	}
+
 	updateString := func(target *string, source *string) {
 		if source != nil {
 			*target = *source
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -454,19 +454,26 @@ var _ = Describe("cluster", func() {
 			})
 		}
 	})
-
 	Context("Update", func() {
-		BeforeEach(func() {
+		It("update_cluster_while_installing", func() {
 			clusterID = strfmt.UUID(uuid.New().String())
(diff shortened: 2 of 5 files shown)


internal/bminventory/inventory.go
```diff
@@ -656,6 +656,12 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
 		tx.Rollback()
 		return installer.NewUpdateClusterNotFound().WithPayload(common.GenerateError(http.StatusNotFound, err))
 	}
+	if b.clusterApi.IsClusterInstallationInProgress(&cluster) {
```
avishayt, line 659:
```suggestion
	if b.clusterApi.HasClusterInstallationStarted(&cluster) {
```

And check if it's installing-in-progress, installing, installed, or error.  User shouldn't be allowed to change the cluster in these states.

oshercc: from the ticket, I understand that the only in installing state nothing can be updated, no ?

avishayt:
The ticket says "after starting the installation", which is all of the states that I mentioned.  Also note that this relates to MGMT-923.

FYI @filanov

filanov:
> The ticket says "after starting the installation", which is all of the states that I mentioned. Also note that this relates to MGMT-923.
> 
> FYI @filanov

Agree with @avishayt

oshercc:
@avishayt @filanov   What about cluster in error ?

filanov:
> @avishayt @filanov What about cluster in error ?

same, can't update

internal/bminventory/inventory.go
ronniel1: @oshercc What @avishayt said :)

internal/bminventory/inventory.go
```diff
@@ -658,6 +658,12 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
 		tx.Rollback()
 		return installer.NewUpdateClusterNotFound().WithPayload(common.GenerateError(http.StatusNotFound, err))
 	}
+
+	if err = b.clusterApi.VerifyClusterUpdatability(&cluster); err != nil {
```
filanov, line 662: rollback

internal/bminventory/inventory.go
```diff
@@ -658,6 +658,13 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
+		log.WithError(err).Errorf("cluster %s can't be updated in current state", params.ClusterID)
+		tx.Rollback()
```
filanov, line 664: if you rebased above master you will not need this line. i added you the review.

oshercc:
@filanov can you please re-approve, had to push conflict fixes.

filanov:
> @filanov can you please re-approve, had to push conflict fixes.

done

PR #233, opened 2020-06-08
Ori/mgmt 1013 Change type of aip-vip and ingress-vip to string and provide regex pattern validation

ori-amizur:


swagger.yaml
```diff
@@ -958,7 +958,7 @@ definitions:
         default: "172.30.0.0/16"
       ingress_vip:
         type: string
-        format: ipv4
+        pattern: '^(([0-9]{1,3}\.){3}[0-9]{1,3})?$'
```
avishayt, line 961: Do we want a more robust regex?  This will allow 999.999.999.999
Copied from stackoverflow:
/^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/

ori-amizur: Actually basically took it from the CIDR regex in the swagger and made some changes.
We can make your proposed change, but maybe we would like to change the CIDR as well?

avishayt: Sure

PR #235, opened 2020-06-08
log fixes

ronniel1:
- Added filename, line, function to the log output
- Replaced direct usage of the global logrus by using the logger configured in main

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -93,7 +93,7 @@ func main() {
 	}
 
 	eventsHandler := events.New(db, log.WithField("pkg", "events"))
-	hwValidator := hardware.NewValidator(Options.HWValidatorConfig)
+	hwValidator := hardware.NewValidator(log.WithField("pkg", "validators"), Options.HWValidatorConfig)
 	instructionApi := host.NewInstructionManager(log, db, hwValidator, Options.InstructionConfig)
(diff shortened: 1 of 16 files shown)


cmd/main.go
```diff
@@ -50,6 +50,8 @@ var Options struct {
 
 func main() {
 	log := logrus.New()
+	log.SetReportCaller(true)
```
filanov, line 53: WOW

internal/common/machine_network_cidr_test.go
```diff
@@ -79,14 +82,14 @@ var _ = Describe("inventory", func() {
 			cluster := createCluster("1.2.5.6", "",
 				createInventory(createInterface("3.3.3.3/16"), createInterface("8.8.8.8/8", "1.2.5.7/23")),
 				createInventory(createInterface("127.0.0.1/17")))
-			_, err := GetMachineCIDRHosts(cluster)
+			_, err := GetMachineCIDRHosts(logrus.New(), cluster)
```
filanov, line 85: maybe it should be a test logger? look at how we did it in the subsystem.

ronniel1: I think that for the subsystem, it matters because you want to see the logs.
Here, it's just passed as a dependency.
We don't really care about it

PR #238, opened 2020-06-09
CronJob to periodically list objects in S3

avishayt:
Useful for debugging the s3 object expiration, the job runs every 15
minutes and you can see the list of objects in the pod's log.

### DIFF
--- a/tools/debug/s3-list-cron.yaml
+++ b/tools/debug/s3-list-cron.yaml
@@ -1,0 +1,23 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: s3-list
+  namespace: assisted-installer
+spec:
+  schedule: "*/15 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: s3-list
+            image: amazon/aws-cli
+            env:
+            - name: AWS_ACCESS_KEY_ID
+              value: "accessKey1"
+            - name: AWS_SECRET_ACCESS_KEY
+              value: "verySecretKey1"
+            command: ["aws"]
+            args: ["--endpoint-url", "http://cloudserver-front:8000", "s3api", "list-objects", "--bucket", "test"]
+          restartPolicy: OnFailure
+      backoffLimit: 3


tools/debug/s3-list-cron.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: s3-list
+  namespace: assisted-installer
+spec:
+  schedule: "*/15 * * * *"
```
filanov, line 7: what is that?

tools/debug/s3-list-cron.yaml
```diff
@@ -0,0 +1,23 @@
...
+          containers:
+          - name: s3-list
+            image: amazon/aws-cli
+            env:
+            - name: AWS_ACCESS_KEY_ID
+              value: "accessKey1"
+            - name: AWS_SECRET_ACCESS_KEY
+              value: "verySecretKey1"
+            command: ["aws"]
+            args: ["--endpoint-url", "http://cloudserver-front:8000", "s3api", "list-objects", "--bucket", "test"]
```
filanov, line 21: the bucket probably should be a parameter as well.
We are not going to use test forever.

tools/debug/s3-list-cron.yaml
avishayt: The syntax for "every 15 minutes"

tools/debug/s3-list-cron.yaml
avishayt: How do I make parameters in a yaml?  I figured this would be an example that people can change when needed.  We won't always have the same access/secret also.

filanov: You can probably use envs as well no?

avishayt: Not that I know of... Envs here are hard-coded for use in the yaml.

filanov:
```
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: s3-list
  namespace: assisted-installer
spec:
  schedule: "*/1 * * * *"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: s3-list
            image: amazon/aws-cli
            env:
            - name: AWS_ACCESS_KEY_ID
              value: "accessKey1"
            - name: AWS_SECRET_ACCESS_KEY
              value: "verySecretKey1"
            command: ["echo"]
            args: ["$(AWS_ACCESS_KEY_ID)"]
          restartPolicy: OnFailure
      backoffLimit: 3
```
```
kubectl logs s3-list-1591711800-m7q8p
accessKey1
```

ronniel1:
@avishayt  whouldn't it be better to have a tool to do that?

filanov:
maybe a single shot job?

filanov:
> I guess I could wrap it in a bash script that takes environment variables, writes the yaml to /tmp, and then applies it. Seems like a bit of overkill. Also one-shot job isn't very helpful I think. Basically I wanted to monitor which objects were in S3 over time so that I could see how the GC was working, so getting the list of objects every 15 minutes was what I needed, not just once.

You can wrap with a python script. 
As long as it's for debug and we don't need it as part of a full deployment i think it's fine.

PR #240, opened 2020-06-09
MGMT-1068 node status stay installing although the install command fa…

gamli75:
…iled

@ronniel1 @filanov @eranco74 please review. 
I tested it in a real environment and it's working - the host status changed to error and the cluster status remains installing.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -958,22 +958,27 @@ func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer
 	log.Infof("Received step reply <%s> from cluster <%s> host <%s>  exit-code <%d> stdout <%s> stderr <%s>", params.Reply.StepID, params.ClusterID,
 		params.HostID, params.Reply.ExitCode, params.Reply.Output, params.Reply.Error)
 
-	//check the output exit code
-	if params.Reply.ExitCode != 0 {
-		err = fmt.Errorf("Exit code is %d reply error is %s for %s reply for host %s cluster %s",
-			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
-		log.WithError(err).Errorf("Exit code is <%d> , reply error is <%s> for <%s> reply for host <%s> cluster <%s>",
-			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
-		return installer.NewPostStepReplyBadRequest().
-			WithPayload(common.GenerateError(http.StatusBadRequest, err))
-	}
-
 	var host models.Host
 	if err = b.db.First(&host, "id = ? and cluster_id = ?", params.HostID, params.ClusterID).Error; err != nil {
 		log.WithError(err).Errorf("Failed to find host <%s> cluster <%s> step <%s>",
 			params.HostID, params.ClusterID, params.Reply.StepID)
 		return installer.NewPostStepReplyNotFound().
 			WithPayload(common.GenerateError(http.StatusNotFound, err))
+	}
+
+	//check the output exit code
+	if params.Reply.ExitCode != 0 {
+		err = fmt.Errorf("Exit code is %d reply error is %s for %s reply for host %s cluster %s",
+			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
+		log.WithError(err).Errorf("Exit code is <%d> , reply error is <%s> for <%s> reply for host <%s> cluster <%s>",
+			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
+		//if it's install step - need to move host to error
+		handlingError := handleInstallationError(params, b, ctx, host)
+		if handlingError != nil {
+			log.WithError(err).Errorf("Failed handling installation error for host <%s> cluster <%s>", params.HostID, params.ClusterID)
+		}
+		return installer.NewPostStepReplyBadRequest().
+			WithPayload(common.GenerateError(http.StatusBadRequest, err))
 	}
 
 	var stepReply string
@@ -994,6 +999,14 @@ func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer
 	}
 
 	return installer.NewPostStepReplyNoContent()
+}
+
+func handleInstallationError(params installer.PostStepReplyParams, b *bareMetalInventory, ctx context.Context, host models.Host) error {
+
+	if strings.HasPrefix(params.Reply.StepID, string(models.StepTypeExecute)) {
+		return b.hostApi.HandleInstallationFailure(ctx, &host)
+	}
+	return nil
 }
 
 func handleReplyByType(params installer.PostStepReplyParams, b *bareMetalInventory, ctx context.Context, host models.Host, stepReply string) error {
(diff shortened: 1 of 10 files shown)


internal/bminventory/inventory.go
```diff
@@ -958,24 +958,29 @@ func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer
...
+			params.HostID, params.ClusterID, params.Reply.StepID)
+		return installer.NewPostStepReplyNotFound().
+			WithPayload(common.GenerateError(http.StatusNotFound, err))
+	}
+
 	//check the output exit code
 	if params.Reply.ExitCode != 0 {
 		err = fmt.Errorf("Exit code is %d reply error is %s for %s reply for host %s cluster %s",
 			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
 		log.WithError(err).Errorf("Exit code is <%d> , reply error is <%s> for <%s> reply for host <%s> cluster <%s>",
```
filanov, line 973: not related to current PR but it's strange the same error is logged twice.

internal/bminventory/inventory.go
```diff
@@ -958,24 +958,29 @@ func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer
...
 			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
+		//if it's install step - need to move host to error
+		handlingError := handleInstallationError(params, b, ctx, host)
```
filanov, line 976: maybe you should call this function `handleErrorReply`? then if we will have more errors to handle it will be added to the same function? after all you check the prefix inside.

internal/host/host.go
```diff
@@ -151,6 +152,23 @@ func (m *Manager) RegisterHost(ctx context.Context, h *models.Host) error {
 	})
 }
 
+func (m *Manager) HandleInstallationFailure(ctx context.Context, h *models.Host) error {
+	var host models.Host
+	err := m.db.First(&host, "id = ? and cluster_id = ?", *h.ID, h.ClusterID).Error
```
filanov, line 157: you already have the host why do you get it again?

internal/host/transition.go
```diff
@@ -28,6 +28,10 @@ type TransitionArgsRegisterHost struct {
 	ctx context.Context
 }
 
+type TransitionArgsHostInstallationFailed struct {
```
filanov, line 31: suggesting to move all the code related to this operation under the same place, where you implemented the `PostHostInstallationFailed` 
maybe even add 

```
////////////////////////////////////////////////////////////////////////////
// Installation failure
////////////////////////////////////////////////////////////////////////////
```

internal/host/transition.go
```diff
@@ -65,3 +69,16 @@ func (th *transitionHandler) PostRegisterDuringInstallation(sw stateswitch.State
...
+	sHost, ok := sw.(*stateHost)
+	if !ok {
+		return errors.New("HostInstallationFailed incompatible type of StateSwitch")
+	}
+	params, ok := args.(*TransitionArgsHostInstallationFailed)
+	if !ok {
+		return errors.New("HostInstallationFailed invalid argument")
+	}
+	return updateHostStateWithParams(logutil.FromContext(params.ctx, th.log), sHost.srcState,
+		"installation command failed", sHost.host, th.db)
```
filanov, line 83: @avishayt do we want to put command output here?

avishayt: Yes

gamli75: the status info is limited to 2048 characters and the output is very long - we decided to write only "installation command failed" for now - see my remark in https://issues.redhat.com/browse/MGMT-1068

internal/bminventory/inventory.go
gamli75: moved the info log after validations

internal/bminventory/inventory.go
gamli75: fixed

internal/host/host.go
gamli75: fixed

internal/host/transition.go
gamli75: fixed

gamli75:
@filanov please review again :)

internal/bminventory/inventory.go
filanov: nothing changed. take a look at line 1013 and this line.
there is a `log.WithError(err)` that will write err and then `Errorf` that will print the same information.

internal/bminventory/inventory.go
```diff
@@ -999,26 +999,32 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params installer.
...
+			WithPayload(common.GenerateError(http.StatusNotFound, err))
+	}
 
 	//check the output exit code
 	if params.Reply.ExitCode != 0 {
 		err = fmt.Errorf("Exit code is %d reply error is %s for %s reply for host %s cluster %s",
 			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
 		log.WithError(err).Errorf("Exit code is <%d> , reply error is <%s> for <%s> reply for host <%s> cluster <%s>",
 			params.Reply.ExitCode, params.Reply.Error, params.Reply.StepID, params.HostID, params.ClusterID)
+		//if it's install step - need to move host to error
```
filanov, line 1017: this comment is not relevant here. maybe inside.

internal/bminventory/inventory.go
```diff
@@ -999,26 +999,32 @@ func (b *bareMetalInventory) GetNextSteps(ctx context.Context, params installer.
...
+		handlingError := handleReplyError(params, b, ctx, &host)
+		if handlingError != nil {
+			log.WithError(err).Errorf("Failed handling installation error for host <%s> cluster <%s>", params.HostID, params.ClusterID)
```
filanov, line 1020: wrong error message.

internal/host/transition.go
filanov: As long as @avishayt is ok with it.

filanov:
@gamli75 you can probably add subsystem test for it.

internal/bminventory/inventory.go
gamli75: fixed

internal/bminventory/inventory.go
gamli75: fixed

internal/bminventory/inventory.go
```diff
@@ -1041,6 +1045,15 @@ func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer
 	return installer.NewPostStepReplyNoContent()
 }
 
+func handleReplyError(params installer.PostStepReplyParams, b *bareMetalInventory, ctx context.Context, host *models.Host) error {
+
+	if strings.HasPrefix(params.Reply.StepID, string(models.StepTypeExecute)) {
```
ori-amizur, line 1050: Execute can be anything - not just install.  If I run a debug command with typo on installing host - it will mark the host in error.
You cannot rely that every execute is install and indicate that installation failed.  You need a clear identification to the command that it is install command.

ronniel1: Nice catch @ori-amizur

filanov: @gamli75

PR #249, opened 2020-06-11
MGMT-1128 Every execute command that will fail - will change the host…

gamli75:
… status to error

@ronniel1 @eranco74 @filanov @ori-amizur please review.
I added a subsystem test to verify it.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -1050,11 +1050,11 @@ func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer
 	return installer.NewPostStepReplyNoContent()
 }
 
-func handleReplyError(params installer.PostStepReplyParams, b *bareMetalInventory, ctx context.Context, host *models.Host) error {
+func handleReplyError(params installer.PostStepReplyParams, b *bareMetalInventory, ctx context.Context, h *models.Host) error {
 
-	if strings.HasPrefix(params.Reply.StepID, string(models.StepTypeExecute)) {
+	if strings.HasPrefix(params.Reply.StepID, host.InstallCmdStep) {
 		//if it's install step - need to move host to error
-		return b.hostApi.HandleInstallationFailure(ctx, host)
+		return b.hostApi.HandleInstallationFailure(ctx, h)
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


internal/bminventory/inventory.go
```diff
@@ -1045,11 +1045,11 @@ func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer
 	return installer.NewPostStepReplyNoContent()
 }
 
-func handleReplyError(params installer.PostStepReplyParams, b *bareMetalInventory, ctx context.Context, host *models.Host) error {
+func handleReplyError(params installer.PostStepReplyParams, b *bareMetalInventory, ctx context.Context, h *models.Host) error {
 
-	if strings.HasPrefix(params.Reply.StepID, string(models.StepTypeExecute)) {
+	if strings.HasPrefix(params.Reply.StepID, host.InstallCmdStep) && strings.Contains(params.Reply.StepID, string(models.StepTypeExecute)) {
```
ronniel1, line 1050: We need to know that this is an "install" command.
No need to check the StepTypeExecute string. It's redundant

internal/host/installcmd.go
```diff
@@ -34,6 +35,7 @@ func (i *installCmd) GetStep(ctx context.Context, host *models.Host) (*models.St
 	step := &models.Step{}
 	step.StepType = models.StepTypeExecute
 	step.Command = "bash"
+	step.StepID = fmt.Sprintf("%s-%s-%s", InstallCmdStep, models.StepTypeExecute, uuid.New().String()[:8])
```
ronniel1, line 38: no need to the execute
Install will alway use execute

internal/bminventory/inventory.go
gamli75: fixed

internal/host/installcmd.go
gamli75: @ori-amizur are you using the step type in the step ID somewhere in your code? can I remove the stepType execute from the step ID?

ori-amizur: I just take the step id from get next steps reply and move it the step reply without modification.
So you can remove the execute from step id.
I just rely on the step type.

PR #250, opened 2020-06-11
Reduce lint time by connfiguring go environment on the build containe…

ronniel1:
…r, to point to the host's cache


skipper.yaml
```diff
@@ -20,3 +20,4 @@ env:
     APPLY_NAMESPACE: $APPLY_NAMESPACE
     TARGET: $TARGET
     INGRESS_DOMAIN: $INGRESS_DOMAIN
+    GOCACHE: "/go/pkg/mod"
```
filanov, line 23: why not `GOCACHE: $GOCACHE` ?
in my laptop:
```
ls /go/pkg/mod
ls: cannot access '/go/pkg/mod': No such file or directory
```

ronniel1: because we are pointing it to the cache that is mapped to the container.
see above where the host's cache dir is mapped to

PR #251, opened 2020-06-11
Added dockerignore

ronniel1:


.dockerignore
```diff
@@ -0,0 +1,4 @@
+build/
```
filanov, line 1: i think we should have the build, after deploy we want to keep the targets sometimes.

PR #254, opened 2020-06-15
Avishay/mgmt 1131

RazRegev:


internal/bminventory/inventory.go
```diff
@@ -1453,3 +1453,36 @@ func setPullSecret(cluster *models.Cluster, pullSecret string) {
...
+		log.WithError(err).Errorf("failed to find cluster %s", params.ClusterID)
+		if gorm.IsRecordNotFoundError(err) {
+			return installer.NewCancelInstallationNotFound().WithPayload(common.GenerateError(http.StatusNotFound, err))
+		}
+		return installer.NewCancelInstallationInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+	}
+
+	// cancellation is made by setting the cluster and and hosts states to error.
+	if err := b.clusterApi.CancelInstallation(ctx, &c, "installation was canceled by user"); err != nil {
+		if gorm.Error(err) {
```
filanov, line 1473: You can return the error from `CancelInstallation` i think it's wrong to check specific gorm error type here.

internal/cluster/cluster.go
```diff
@@ -221,3 +223,14 @@ func (m *Manager) VerifyClusterUpdatability(c *models.Cluster) (err error) {
 	}
 	return err
 }
+
+func (m *Manager) CancelInstallation(ctx context.Context, c *models.Cluster, reason string) (err error) {
+	clusterStatus := swag.StringValue(c.Status)
+	if clusterStatus == clusterStatusError {
```
filanov, line 229: take a look at `VerifyClusterUpdatability` you can use the same concept

internal/cluster/cluster.go
```diff
@@ -221,3 +223,14 @@ func (m *Manager) VerifyClusterUpdatability(c *models.Cluster) (err error) {
...
+		return nil
+	} else if clusterStatus != clusterStatusInstalling {
+		return fmt.Errorf("unable to cancel installation of cluster %s in status <%s>", c.ID.String(), clusterStatus)
```
filanov, line 232: please use `github.com/pkg/errors` or "error_utils" to return errors.

internal/host/host.go
```diff
@@ -271,3 +272,14 @@ func (m *Manager) UpdateConnectivityReport(ctx context.Context, h *models.Host,
 	}
 	return nil
 }
+
+func (m *Manager) CancelInstallation(ctx context.Context, h *models.Host, reason string) error {
```
filanov, line 276: i think it should be part of the state machine.

avishayt:
> OK, I went over the code and I can't find any code that actually cancel the installation.
> Seems that it just update the status but the installation is still moving on and at some point the hosts will reboot and we will lose them.
> Also I don't see where we check if the hosts already did the reboot (we should return error in that case since there is no way back once reboot happened)

Setting the status should be enough for the user.  Sure, we're not stopping work that may be happening on the hosts, but we're allowing the user to stop the state machine and start over if they want.  Stopping the hosts would require extra communication between the agent and the service that we don't have today, adding complexity, and it doesn't seem necessary for this use case.

eranco74:
> would require extra communication between the agent and the service that we don't have today, adding complexity, and it doesn't seem necessary for this use case.

Sorry to say, but other than chaining the status it does nothing, if the installation has started we will lose the host once it reboots.
What we should do in case we need to cancel the installation is execute a command that kill the installer container on the host (this is what you mean by extra communication?). it's possible and not that complex.
I perhaps I fail to understand the use case of this API. When is the user going to use it and how will is it going to work if we know that at some point the host is going to reboot and we will lose the agent?

filanov:
> > would require extra communication between the agent and the service that we don't have today, adding complexity, and it doesn't seem necessary for this use case.
> 
> Sorry to say, but other than chaining the status it does nothing, if the installation has started we will lose the host once it reboots.
> What we should do in case we need to cancel the installation is execute a command that kill the installer container on the host (this is what you mean by extra communication?). it's possible and not that complex.
> Perhaps I fail to understand the use case of this API. When is the user going to use it and how is it going to work if we know that at some point the host is going to reboot and we will lose the agent?

If i understand correctly, we don't stop the installation because we don't know if we are able to do so. for example host was shut down so you can't cancel anything, of the agent stopped working and not getting any command, or just disconnected. In those cases we still want to give the user to retry the installation because otherwise he will need to create everything from scratch. 
So for those cases it's the user responsibility to handle the hosts, he will need to shut down, delete or do anything he need to make them stop the installation, so this change will handle only the management side.

avishayt:
> > > would require extra communication between the agent and the service that we don't have today, adding complexity, and it doesn't seem necessary for this use case.
> > 
> > 
> > Sorry to say, but other than chaining the status it does nothing, if the installation has started we will lose the host once it reboots.
> > What we should do in case we need to cancel the installation is execute a command that kill the installer container on the host (this is what you mean by extra communication?). it's possible and not that complex.
> > Perhaps I fail to understand the use case of this API. When is the user going to use it and how is it going to work if we know that at some point the host is going to reboot and we will lose the agent?
> 
> If i understand correctly, we don't stop the installation because we don't know if we are able to do so. for example host was shut down so you can't cancel anything, of the agent stopped working and not getting any command, or just disconnected. In those cases we still want to give the user to retry the installation because otherwise he will need to create everything from scratch.
> So for those cases it's the user responsibility to handle the hosts, he will need to shut down, delete or do anything he need to make them stop the installation, so this change will handle only the management side.
> I agree that it looks hacky, @eranco74 do you see other options to handle those cases?
> @romfreiman @ronniel1 Who required this feature?

Yes @filanov , that is correct.  There were cases where the installation got stuck or failed, so rather than start everything from scratch (create new ISO, download it, boot machines with it, etc.), the request was to cancel from the management's point of view and then to allow resetting it (MGMT-1099).  After it is canceled and reset, the user can simply reboot the hosts and they will register again.

eranco74:
OK, make sense,
I think that in order to make sure we don't keep reference to dead nodes wee should probably delete all nodes from the cluster and wait for them to re-register.
Also it might help the user if we will trigger restart of the cluster nodes so he won't need to do it himself.
Also consider renaming it to `resetState` instead of `cancleInstallation`

Makefile
```diff
@@ -15,8 +15,8 @@ kubectl get service $(1) -n assisted-installer | grep $(1) | awk '{print $$4 ":"
 endef
 endif
 
-SERVICE := $(or ${SERVICE},quay.io/ocpmetal/bm-inventory:stable)
-OBJEXP := $(or ${OBJEXP},quay.io/ocpmetal/s3-object-expirer:stable)
+SERVICE := $(or ${SERVICE},quay.io/rregev/bm-inventory:stable)
```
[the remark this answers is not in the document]
filanov: why overwrite it the first place? just use `export SERVICE=quay.io/rregev/bm-inventory:stable`

internal/bminventory/inventory.go
```diff
@@ -1462,3 +1462,30 @@ func setPullSecret(cluster *common.Cluster, pullSecret string) {
...
+	if err := b.db.Preload("Hosts").First(&c, "id = ?", params.ClusterID).Error; err != nil {
+		log.WithError(err).Errorf("failed to find cluster %s", params.ClusterID)
+		if gorm.IsRecordNotFoundError(err) {
+			return installer.NewCancelInstallationNotFound().WithPayload(common.GenerateError(http.StatusNotFound, err))
+		}
+		return installer.NewCancelInstallationInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+	}
+
+	// cancellation is made by setting the cluster and and hosts states to error.
+	if err := b.clusterApi.CancelInstallation(ctx, &c, "installation was canceled by user"); err != nil {
```
filanov, line 1481: Maybe should be done in a transaction?
If you set the cluster but for some reason you failed to set the one of the hosts then you will not be able to retry.

internal/host/host.go
```diff
@@ -274,3 +279,21 @@ func (m *Manager) UpdateConnectivityReport(ctx context.Context, h *models.Host,
 	}
 	return nil
 }
+
+func (m *Manager) CancelInstallation(ctx context.Context, h *models.Host, reason string) *common.ApiErrorResponse {
+	hostStatus := swag.StringValue(h.Status)
+	allowedStatuses := []string{HostStatusInstalling, HostStatusInstallingInProgress}
```
filanov, line 285: why do you need this part? it should be part of the state machine.

filanov, line 285: what will happen if one of the hosts is in error? or done installing?

RazRegev: as far as I know it's okay for one of the hosts to be in error, in this case we just skip it.
the goal is to set the hosts to error, does it matter if it already in that state?

RazRegev: I thought to do that so we could not involve the state machine with the error code handling.
I will think of a better way to check this.

filanov: i think that the state machine should handle it not some external logic.

internal/bminventory/inventory_test.go
```diff
@@ -805,6 +809,15 @@ var _ = Describe("cluster", func() {
 
 			verifyApiError(reply, http.StatusInternalServerError)
 		})
+
+		It("cancel installation success", func() {
```
filanov, line 813: Please test error cases as well

internal/cluster/statecluster.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/filanov/bm-inventory/internal/common"
+	"github.com/filanov/stateswitch"
+	"github.com/go-openapi/swag"
+	"github.com/jinzhu/gorm"
+)
+
+type stateCluster struct {
+	srcState string
+	cluster  *common.Cluster
+	db       *gorm.DB
```
filanov, line 13: i don't think it's belong here. it's part of the parameters  not part of the state cluster.

internal/cluster/statecluster.go
```diff
@@ -0,0 +1,35 @@
...
+		db:       db,
+	}
+}
+
+func (sh *stateCluster) State() stateswitch.State {
+	return stateswitch.State(swag.StringValue(sh.cluster.Status))
+}
+
+func (sh *stateCluster) SetState(state stateswitch.State) error {
+	err := sh.db.Preload("Hosts").First(&sh.cluster, "id = ?", sh.cluster.ID).Error
```
filanov, line 29: Why do you get the cluster from the DB?  you already have the cluster no?

internal/cluster/statemachine.go
```diff
@@ -0,0 +1,24 @@
...
+	TransitionTypeCancelInstallation = "CancelInstallation"
+)
+
+func NewClusterStateMachine(th *transitionHandler) stateswitch.StateMachine {
+	sm := stateswitch.NewStateMachine()
+
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeCancelInstallation,
+		SourceStates: []stateswitch.State{
+			clusterStatusInstalling,
```
filanov, line 17: what about installing in progress?

internal/cluster/transition.go
```diff
@@ -0,0 +1,41 @@
...
+func (th *transitionHandler) PostCancelInstallation(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error {
+	sCluster, ok := sw.(*stateCluster)
+	if !ok {
+		return errors.New("PostCancelInstallation incompatible type of StateSwitch")
+	}
+	params, ok := args.(*TransitionArgsCancelInstallation)
+	if !ok {
+		return errors.New("PostCancelInstallation invalid argument")
+	}
+	if sCluster.srcState == clusterStatusError {
```
filanov, line 36: why do it here? it's not even in the transition definition? how do you get to this state?

internal/cluster/transition.go
```diff
@@ -0,0 +1,41 @@
...
+		return nil
+	}
+	return updateClusterStateWithParams(logutil.FromContext(params.ctx, th.log), sCluster.srcState,
+		params.reason, sCluster.cluster, sCluster.db)
```
filanov, line 40: the DB should be part of the args.

internal/cluster/transition_test.go
```diff
@@ -0,0 +1,46 @@
...
+		clusterId strfmt.UUID
+	)
+
+	BeforeEach(func() {
+		db = prepareDB()
+		capi = NewManager(getTestLog(), db, nil)
+		clusterId = strfmt.UUID(uuid.New().String())
+	})
+
+	It("cancel_installation", func() {
```
filanov, line 34: testing just the good flow is like not testing at all.

internal/host/transition.go
```diff
@@ -45,19 +43,19 @@ func (th *transitionHandler) PostRegisterHost(sw stateswitch.StateSwitch, args s
 	log := logutil.FromContext(params.ctx, th.log)
 
 	// If host already exists
-	if err := th.db.First(&host, "id = ? and cluster_id = ?", sHost.host.ID, sHost.host.ClusterID).Error; err == nil {
+	if err := sHost.db.First(&host, "id = ? and cluster_id = ?", sHost.host.ID, sHost.host.ClusterID).Error; err == nil {
```
filanov, line 46: same here. db is not part of state host.

internal/host/transition.go
```diff
@@ -91,5 +90,30 @@ func (th *transitionHandler) PostHostInstallationFailed(sw stateswitch.StateSwit
...
+func (th *transitionHandler) PostCancelInstallation(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error {
+	sHost, ok := sw.(*stateHost)
+	if !ok {
+		return errors.New("PostCancelInstallation incompatible type of StateSwitch")
+	}
+	params, ok := args.(*TransitionArgsCancelInstallation)
+	if !ok {
+		return errors.New("PostCancelInstallation invalid argument")
+	}
+	if sHost.srcState == HostStatusError {
```
filanov, line 114: Do you still need to update the reason for error state as well?

internal/host/transition.go
RazRegev: So how could I get the db tr

internal/host/transition.go
RazRegev: no, that's why I'm not updating the reason if the source state is already error.

internal/cluster/statemachine.go
RazRegev: there is no such state 'in-progress' for the cluster

internal/cluster/transition.go
RazRegev: We don't want to update the state info of a cluster that is already in error state.

internal/host/transition.go
filanov: It was a question.

internal/cluster/statemachine.go
```diff
@@ -0,0 +1,24 @@
...
+const (
+	TransitionTypeCancelInstallation = "CancelInstallation"
+)
+
+func NewClusterStateMachine(th *transitionHandler) stateswitch.StateMachine {
+	sm := stateswitch.NewStateMachine()
+
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeCancelInstallation,
+		SourceStates: []stateswitch.State{
```
filanov, line 16: What about running cancel on a cluster that is already in error state? Do we support it?

RazRegev: Yes, I excellently removed it

PR #255, opened 2020-06-15
VIP: Avishay/mgmt 1099

RazRegev:


internal/bminventory/inventory.go
```diff
@@ -1453,3 +1453,35 @@ func setPullSecret(cluster *models.Cluster, pullSecret string) {
 		cluster.PullSecretSet = false
 	}
 }
+
+func (b *bareMetalInventory) ResetCluster(ctx context.Context, params installer.UpdateHostInstallProgressParams) middleware.Responder {
+	log := logutil.FromContext(ctx, b.log)
+	log.Infof("reset installation for cluster %s", params.ClusterID)
```
filanov, line 1459: Not only installation right? 
It will change the cluster from every error state

internal/bminventory/inventory.go
```diff
@@ -1453,3 +1453,35 @@ func setPullSecret(cluster *models.Cluster, pullSecret string) {
...
+	for _, h := range c.Hosts {
+		if err := b.hostApi.DeregisterFailedHost(h); err != nil {
+			if gorm.Error(err) {
+				return installer.NewResetClusterInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+			}
+			return installer.NewResetClusterConflict().WithPayload(common.GenerateError(http.StatusConflict, err))
+		}
+	}
+	if err := b.clusterApi.ResetCluster(ctx, &c, "installation was reset by user"); err != nil {
+		if gorm.Error(err) {
```
filanov, line 1480: why gorm error? 
Use `error_utils.go`

internal/host/host.go
```diff
@@ -153,6 +154,14 @@ func (m *Manager) RegisterHost(ctx context.Context, h *models.Host) error {
 	})
 }
 
+func (m *Manager) DeregisterFailedHost(h *models.Host) error {
+	hostStatus := swag.StringValue(h.Status)
+	if hostStatus == HostStatusError {
```
filanov, line 159: Are you sure that all the hosts will be in error if installation fails?

internal/host/host.go
```diff
@@ -153,6 +154,14 @@ func (m *Manager) RegisterHost(ctx context.Context, h *models.Host) error {
...
+		return fmt.Errorf("unable to deregister host %s in status %s", h.ID.String(), hostStatus)
```
filanov, line 160: use `github.com/pkg/errors`

PR #257, opened 2020-06-16
MGMT-1150 Add an option to deplpoy with specific tag

filanov:
All the targets will be taken from ocpmetal quay.io account.
The usage is to add DEPLOY_TAG env to the makefile:
skipper make deploy-all DEPLOY_TAG=stable

For the UI
skipper make deploy-ui DEPLOY_TAG=latest

### DIFF
--- a/tools/deploy_assisted_installer.py
+++ b/tools/deploy_assisted_installer.py
@@ -1,14 +1,24 @@
 import os
 import utils
+import argparse
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--deploy-tag", help='Tag for all deployment images', type=str, default='latest')
+args = parser.parse_args()
 
 SRC_FILE = os.path.join(os.getcwd(), "deploy/bm-inventory.yaml")
 DST_FILE = os.path.join(os.getcwd(), "build/bm-inventory.yaml")
+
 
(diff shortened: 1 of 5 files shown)


tools/deploy_assisted_installer.py
```diff
@@ -1,14 +1,24 @@
 import os
 import utils
+import argparse
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--deploy-tag")
```
eranco74, line 7: Consider adding
```suggestion
parser.add_argument("--deploy-tag", help='Tag for all deployment images', type=str, default='latest')
```
Will save you some if statements in the code.

filanov: Because i'm using it in the Makefile it will be complicated to have functions that will determinate if the flag was set or not, so in the makefile the default value is "" (empty string) in this case the default value that i set here will not be used. for example
```
python bla.py  --deploy-tag ""
Namespace(create_namespace=False, deploy_tag='', domain=None)
```
And currently everyone are using stable and not latest, except the UI. 
But i don't mind adding it if it will help you in the future.

filanov: Done, added.

PR #259, opened 2020-06-16
README improvements

rollandf:


romfreiman:
why arent we merging?

PR #260, opened 2020-06-16
MGMT-1164 - Adapted prometheus endpoint metrics to contain the

ronniel1:
cluster-id where applicable


Makefile
```diff
@@ -26,7 +26,7 @@ lint:
 	golangci-lint run -v
 
 .PHONY: build
-build: create-build-dir lint unit-test
+build: create-build-dir #lint unit-test
```
filanov, line 29: ?

ronniel1: sorry, will fix

internal/metrics/reporter.go
```diff
@@ -0,0 +1,103 @@
...
+			w:       wi,
+			method:  r.Method,
+			urlPath: r.URL.Path,
+			ctx:     r.Context(),
+		}
+
+		mr := rmiddleware.MatchedRouteFrom(r)
+		if mr != nil {
+			reporter.ctx = matchedRouteContext.ToContext(reporter.ctx, mr, r.Method)
+			log.Info("Matched rout found")
```
filanov, line 35: Maybe you should add the route to the log as well, in both cases found and not found.

internal/metrics/recorder.go
```diff
@@ -0,0 +1,160 @@
...
+	ret := metricLabels{
+		clusterID: "",
+		path:      "",
+		method:    "",
+	}
+	mr, method := matchedRouteContext.FromContext(ctx)
+	if mr != nil {
+		ret.path = mr.PathPattern
+
+		if v, _, ok := mr.Params.GetOK("cluster_id"); ok {
```
filanov, line 137: some times it can just id right?

internal/metrics/reporter.go
avishayt: Is this log too much noise if found?
Also s/rout/route/ - also on line 37

PR #262, opened 2020-06-16
Add release to GitHub workflow 

eranco74:


.github/workflows/image-push.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         REGISTRY: 'quay.io/ocpmetal'
         dockerfile: "Dockerfile.bm-inventory"
         buildargs: GIT_REVISION
-        tags: "stable,${{ env.GIT_REVISION }}"
+        tags: "latest,${{ env.GIT_REVISION }}"
```
filanov, line 40: You need to change the makefile use latest as default.

PR #263, opened 2020-06-16
Updated assisted-installer default image tag from stable to latest

eranco74:


internal/host/hwinfocmd.go
```diff
@@ -21,6 +21,6 @@ func (h *hwInfoCmd) GetStep(ctx context.Context, host *models.Host) (*models.Ste
 	step := &models.Step{}
 	step.StepType = models.StepTypeHardwareInfo
 	step.Command = "podman"
-	step.Args = strings.Split("run,--rm,--privileged,--quiet,--net=host,-v,/var/log:/var/log,quay.io/ocpmetal/hardware_info:stable,/usr/bin/hardware_info", ",")
+	step.Args = strings.Split("run,--rm,--privileged,--quiet,--net=host,-v,/var/log:/var/log,quay.io/ocpmetal/hardware_info:latest,/usr/bin/hardware_info", ",")
```
filanov, line 24: @ori-amizur is it part of you pr as well?

filanov:
@ori-amizur @eranco74 you both touch the same area, please sync.

internal/host/hwinfocmd.go
ori-amizur: yes

ori-amizur: yes

ronniel1: @eranco74 PR should be merged first

ronniel1: @eranco74 shouldn't it take the tag from the environment variable?

eranco74: It should get it from env.
I'll update

internal/host/hwinfocmd.go
```diff
@@ -2,25 +2,30 @@ package host
...
 )
 
-type hwInfoCmd baseCmd
+type hwInfoCmd struct {
+	baseCmd
+	instructionConfig InstructionConfig
+}
 
-func NewHwInfoCmd(log logrus.FieldLogger) *hwInfoCmd {
+func NewHwInfoCmd(log logrus.FieldLogger, instructionConfig InstructionConfig) *hwInfoCmd {
```
filanov, line 18: why does HW info need `InstructionConfig`? why not just set the image?

eranco74: It should get the image from env and I wanted to have all the instructions env overrides in one place.
Think I should move it the hwinfocmd.go?

filanov: no, i think you can have a single struct like `InstructionConfig` but specific command should get only the relevant fields no?

eranco74: Agree, updating

internal/host/instructionmanager.go
```diff
@@ -28,15 +28,16 @@ type InstructionManager struct {
 	stateToSteps stateToStepsMap
 }
 type InstructionConfig struct {
```
filanov, line 30: @ori-amizur is `quay.io/ocpmetal/inventory` should be used somewhere?

PR #268, opened 2020-06-17
Avishay/mgmt 948

avishayt:


filanov:
@avishayt what will happen if the value already exists in the DB and the user didn't pass any parameters?

avishayt:
> @avishayt what will happen if the value already exists in the DB and the user didn't pass any parameters?

Any values that aren't passed get deleted, that's the semantics of PUT, and that's how the user can reset properties back to their default values.

filanov:
> > @avishayt what will happen if the value already exists in the DB and the user didn't pass any parameters?
> 
> Any values that aren't passed get deleted, that's the semantics of PUT, and that's how the user can reset properties back to their default values.

it's a big change, does UI and test-infra ready for it? subsystem as well.

avishayt:
> > > @avishayt what will happen if the value already exists in the DB and the user didn't pass any parameters?
> > 
> > 
> > Any values that aren't passed get deleted, that's the semantics of PUT, and that's how the user can reset properties back to their default values.
> 
> it's a big change, does UI and test-infra ready for it? subsystem as well.

avishayt:
Looking into alternatives

filanov:
i'm not saying that we should not do it. just want to be sure that we don't break anything.

PR #269, opened 2020-06-18
MGMT-1203: Set StatusUpdatedAt upon cluster/host create

avishayt:


internal/bminventory/inventory.go
```diff
@@ -274,6 +274,7 @@ func (b *bareMetalInventory) RegisterCluster(ctx context.Context, params install
 		ServiceNetworkCidr:       swag.StringValue(params.NewClusterParams.ServiceNetworkCidr),
 		SSHPublicKey:             params.NewClusterParams.SSHPublicKey,
 		UpdatedAt:                strfmt.DateTime{},
+		StatusUpdatedAt:          strfmt.DateTime(time.Now()),
```
filanov, line 277: shouldn't it be part of the internal state machine logic?

avishayt: It already is - the problem is that it's not set upon creation (before state machine is in action)

filanov: if it's only on registration then you can add it here: https://github.com/filanov/bm-inventory/blob/master/internal/host/transition.go#L53 just before the create call

avishayt:
@filanov Makes sense, is this better?

PR #274, opened 2020-06-21
Ori/mgmt 1220 bm-inventory should invoke the free-ip tool on the agent and validate the VIPs provided by the user

ori-amizur:


internal/host/freeaddressescmd_test.go
```diff
@@ -0,0 +1,46 @@
...
+		fCmd = NewFreeAddressesCmd(getTestLog(), "quay.io/ocpmetal/free_addresses:latest")
+
+		id = strfmt.UUID(uuid.New().String())
+		clusterId = strfmt.UUID(uuid.New().String())
+		host = getTestHost(id, clusterId, HostStatusInsufficient)
+		host.Inventory = defaultInventory()
+		Expect(db.Create(&host).Error).ShouldNot(HaveOccurred())
+	})
+
+	It("get_step", func() {
```
filanov, line 34: what about bad flows? empty up addresses?

ori-amizur: OK.  Will add.

ori-amizur:
> Where is the part of the validation for the VIPs provided by the user?

In inventory.go

internal/host/freeaddressescmd.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (f *freeAddressesCmd) prepareParam(host *models.Host) (string, error) {
+	var inventory models.Inventory
+	err := json.Unmarshal([]byte(host.Inventory), &inventory)
+	if err != nil {
+		f.log.WithError(err).Warn("Inventory parse")
+		return "", err
+	}
+	m := make(map[string]string)
```
ronniel1, line 33: I assume that you need a set here, and not really a map.
Better make it map[string]bool, or even map[string]struct{}

ori-amizur: OK.

PR #276, opened 2020-06-21
Liat/mgmt 953 move insufficient host status handling to host monitor and add connectivity check 

gamli75:
I moved the insufficient/ known host status handling to host monitor and add connectivity check 

@filanov @ronniel1 @ori-amizur @avishayt please review


cmd/main.go
```diff
@@ -52,7 +53,7 @@ var Options struct {
 	InstructionConfig           host.InstructionConfig
 	ClusterStateMonitorInterval time.Duration `envconfig:"CLUSTER_MONITOR_INTERVAL" default:"10s"`
 	S3Config                    s3wrapper.Config
-	HostStateMonitorInterval    time.Duration `envconfig:"HOST_MONITOR_INTERVAL" default:"30s"`
+	HostStateMonitorInterval    time.Duration `envconfig:"HOST_MONITOR_INTERVAL" default:"8s"`
```
filanov, line 56: Why 8 seconds?

internal/common/validators.go
```diff
@@ -0,0 +1,7 @@
+package common
```
filanov, line 1: Why not put it under validators?
It looks like common is our garbage collector.

internal/connectivity/validator.go
```diff
@@ -0,0 +1,64 @@
...
+
+	_, err := v.GetHostValidInterfaces(host)
+	if err != nil {
+		isSufficient = false
+		reason = "Waiting to receive connectivity information"
+	}
+
+	if cluster.MachineNetworkCidr == "" {
+		isSufficient = false
+		reason += ", Could not determine connectivity because API VIP not set"
```
filanov, line 40: if this is the only error then the end reply will be `", Could not determine connectivity because API VIP not set"`?
maybe you can collect multiple reasons with an array of strings and then create a single reply from them?

internal/connectivity/validator.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if !common.IsHostInMachineNetCidr(v.log, cluster, host) {
```
filanov, line 43: don't think that `IsHostInMachineNetCidr` should be under common. it sounds like something really specific.
like network validations or something like this.

internal/connectivity/validator.go
```diff
@@ -0,0 +1,64 @@
...
+	return &validator{
+		log: log,
+	}
+}
+
+type validator struct {
+	log logrus.FieldLogger
+}
+
+func (v *validator) IsSufficient(host *models.Host, cluster *common.Cluster) (*common.IsSufficientReply, error) {
```
filanov, line 28: probably should add a comment about what is validated here.

internal/connectivity/validator_test.go
```diff
@@ -0,0 +1,154 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestValidator(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Connectivity Validator tests Suite")
+}
+
+var _ = Describe("connectivity_validator", func() {
```
filanov, line 24: is there a test for a good flow?

internal/connectivity/validator_test.go
```diff
@@ -0,0 +1,154 @@
...
+		roles := []string{"", "master", "worker"}
+		for _, role := range roles {
+			host.Role = role
+			insufficient(connectivityValidator.IsSufficient(host, cluster))
+		}
+	})
+
+})
+
+func insufficient(reply *common.IsSufficientReply, err error) {
```
filanov, line 109: the name is not clear maybe `validateInsufficiency`?

internal/connectivity/validator_test.go
```diff
@@ -0,0 +1,154 @@
...
+		hw, err := json.Marshal(&inventory)
+		Expect(err).NotTo(HaveOccurred())
+		host.Inventory = string(hw)
+		interfaces, err := connectivityValidator.GetHostValidInterfaces(host)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(len(interfaces)).Should(Equal(1))
+	})
+
+	It("invalid interfaces", func() {
+
```
filanov, line 147: please remove empty line

internal/hardware/validator_test.go
```diff
@@ -191,13 +152,13 @@ var _ = Describe("hardware_validator", func() {
 
 })
 
-func sufficient(reply *IsSufficientReply, err error) {
+func sufficient(reply *common.IsSufficientReply, err error) {
```
filanov, line 155: the names `sufficient` and `insufficient` are not clear

internal/host/common.go
```diff
@@ -119,18 +125,99 @@ func getCluster(clusterID strfmt.UUID, db *gorm.DB) (*common.Cluster, error) {
...
 }
 
-func updateStateFromInventory(log logrus.FieldLogger, hwValidator hardware.Validator, h *models.Host, db *gorm.DB) (*UpdateReply, error) {
+func updateInventory(log logrus.FieldLogger, hwValidator hardware.Validator, h *models.Host, db *gorm.DB) (*UpdateReply, error) {
 	cluster, err := getCluster(h.ClusterID, db)
 	if err != nil {
 		return nil, err
 	}
-	reply, err := hwValidator.IsSufficient(h, cluster)
+	_, err = hwValidator.IsSufficient(h, cluster)
```
filanov, line 133: What is the purpose of this call if you don't care about the output?
If it's to validate that the interfaces list is not empty then it's not clear at all, consider adding a another API for hw info valaidation

internal/host/common.go
```diff
@@ -119,18 +125,99 @@ func getCluster(clusterID strfmt.UUID, db *gorm.DB) (*common.Cluster, error) {
...
 	if err != nil {
 		return nil, err
 	}
-	if !reply.IsSufficient {
-		return updateStateWithParams(log, HostStatusInsufficient, reply.Reason, h, db,
-			"inventory", h.Inventory)
+	status, statusInfo := "", ""
+	if h.Status != nil {
```
filanov, line 138: Consider adding a function that will get the status and status info with empty string as default. 
I see this code repeat itself

internal/host/common.go
```diff
@@ -119,18 +125,99 @@ func getCluster(clusterID strfmt.UUID, db *gorm.DB) (*common.Cluster, error) {
...
+	if err != nil || stateReply.IsChanged {
+		return stateReply, err
+	}
+	var statusInfoDetails = make(map[string]string)
+	//checking inventory isInsufficient
+	cluster, err := getCluster(h.ClusterID, db)
+	if err != nil {
+		return nil, err
+	}
+	inventoryReply, _ := hwValidator.IsSufficient(h, cluster)
```
filanov, line 186: why not check the error?

internal/host/common.go
```diff
@@ -119,18 +125,99 @@ func getCluster(clusterID strfmt.UUID, db *gorm.DB) (*common.Cluster, error) {
...
+	}
+
+	return &common.IsSufficientReply{
+		Type:         "role",
+		IsSufficient: isSufficient,
+		Reason:       reason,
+	}
+}
+
+func isSufficientHost(log logrus.FieldLogger, h *models.Host, db *gorm.DB, hwValidator hardware.Validator, connectivityValidator connectivity.Validator) (*UpdateReply, error) {
```
filanov, line 174: The name is confusing. a functions starting with `is` usually return boolean, this function actually update host status, consider renaming.

subsystem/cluster_test.go
```diff
@@ -146,6 +146,9 @@ func waitForClusterState(ctx context.Context, clusterID strfmt.UUID, state strin
 	Expect(err).NotTo(HaveOccurred())
 	c := rep.GetPayload()
 	Expect(swag.StringValue(c.Status)).Should(Equal(state))
+	if stateInfo != "NR" {
```
filanov, line 149: `NR` should probably a const, and consider changing to `IgnoreStateInfo` or something like this.

gamli75:
> I like the `waitFor...` functions in the subsystem

cmd/main.go
gamli75: The host monitor interval should be up to 10 seconds because I don't think we want to wait 30 seconds if we update inventory/ connectivity and there is a problem.
I didn't want it to be 10 seconds in order to reduce the cases that the cluster & host monitor will run at the same time. 
I think it won't be relevant if we merge the cluster & host monitors to one monitor.

internal/common/validators.go
gamli75: fixed

internal/connectivity/validator.go
gamli75: moved it to network package

internal/connectivity/validator.go
gamli75: added a comment

internal/host/common.go
gamli75: The purpose of this call is that I don't want to update the hw info if there is a parsing error - I think it's better to keep the old data.

filanov: It's not clear when reading the code.
Please add a comment or add another function `ValidateHWFormat` or something like this.

internal/host/common.go
gamli75: because in this case, I need the reason from the reply and the only case that I will get an error is if there is a parsing error

internal/host/common.go
gamli75: I changed it to checkAndUpdateSufficientHost, if you have better name please suggest :)

subsystem/cluster_test.go
gamli75: fixed

internal/host/common.go
filanov: you right today, but if it's going to change it will become a bug.

internal/connectivity/validator.go
gamli75: fixed. I fixed it also in the hardware validator.

internal/host/common.go
```diff
@@ -119,18 +132,88 @@ func getCluster(clusterID strfmt.UUID, db *gorm.DB) (*common.Cluster, error) {
...
+	}
+	inventoryReply, _ := hwValidator.IsSufficient(h, cluster)
+	if inventoryReply != nil {
+		statusInfoDetails[inventoryReply.Type] = inventoryReply.Reason
+	} else {
+		statusInfoDetails["hardware"] = "parsing error"
+	}
+
+	//checking connectivity isSufficient
+	connectivityReply, _ := connectivityValidator.IsSufficient(h, cluster)
```
filanov, line 190: should probably validate the error, to prevent future bugs.

internal/host/host_test.go
```diff
@@ -260,7 +267,48 @@ func getTestHost(hostID, clusterID strfmt.UUID, state string) models.Host {
 		Status:       swag.String(state),
 		HardwareInfo: defaultHwInfo,
 		Inventory:    defaultInventory(),
+		Role:         "worker",
+		CheckedInAt:  strfmt.DateTime(time.Now()),
+	}
+}
+
+func mockConnectivityAndHwValidators(h *models.Host, mockHWValidator *hardware.MockValidator, mockConnectivityValidator *connectivity.MockValidator, hwError, sufficientHw, sufficientConnectivity bool) string {
```
filanov, line 275: Looks very complicated.
consider splitting to multiple functions.

internal/host/common.go
gamli75: fixed

cmd/main.go
```diff
@@ -111,8 +112,9 @@ func main() {
 	versionHandler := versions.NewHandler(Options.Versions)
 	eventsHandler := events.New(db, log.WithField("pkg", "events"))
 	hwValidator := hardware.NewValidator(log.WithField("pkg", "validators"), Options.HWValidatorConfig)
-	instructionApi := host.NewInstructionManager(log, db, hwValidator, Options.InstructionConfig)
-	hostApi := host.NewManager(log.WithField("pkg", "host-state"), db, hwValidator, instructionApi)
+	connectivityValidator := connectivity.NewValidator(log.WithField("pkg", "validators"))
+	instructionApi := host.NewInstructionManager(log, db, hwValidator, Options.InstructionConfig, connectivityValidator)
```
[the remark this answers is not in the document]
gamli75: fixed

internal/host/common.go
gamli75: fixed

internal/host/common.go
gamli75: I added a comment in the code

PR #277, opened 2020-06-21
MGMT-1043 Disallow update role when host is disabled

YuviGold:


filanov:
Just to make sure, did you run subsystem?

PR #278, opened 2020-06-21
MGMT-948: UpdateCluster set in DB only if param not nil

avishayt:


avishayt:
@filanov want to take a look too?

PR #279, opened 2020-06-22
Sepcify go-swagger version to v0.24.0

filanov:
The new version need some adjustments to go-modules. 
Until now we used latest so each one had a different version that he used.


filanov:
Checked with test-infra, no need to update the agent.

PR #281, opened 2020-06-22
Igal/mgmt 1207

tsorya:


swagger.yaml
```diff
@@ -872,9 +872,16 @@ definitions:
         type: string
 
   steps:
-    type: array
-    items:
-      $ref: '#/definitions/step'
+    type: object
+    properties:
+      next_step_in:
```
ronniel1, line 877: @avishayt Can you review the swagger?

avishayt: The API is "/clusters/{cluster_id}/hosts/{host_id}/instructions" so please change "next_step_in" to "next_instruction_seconds", and "steps" to "instructions".  I don't care if "steps" is in the code or structs, just don't want it in the API.  Thanks!

PR #282, opened 2020-06-22
MGMT-1101: introduce Route 53 DNS service support

danielerez:
As part of cluster installation flow, base DNS domain
has to be specified by the user. When using a supported
DNS domain [1], requests for adding relevant record sets
are invoked using a dns-client [2]. For now, supporting
AWS Route 53 service.

The following record sets (of type A) should be registered:
- api.cluster-name.domain-name -> API Virtual IP
- *.apps.cluster-name.domain-name -> Ingress Virtual IP

Affected flows:
- Cluster installation: create DNS record sets.
- Cluster deregister: cleanup (remove the record sets).

Usage requirements:
- Create 'route53-creds' secret from the /.aws/credentials file:
   $ oc -n assisted-installer create secret generic route53-creds \
          --from-file=/root/.aws/credentials
- Add the base domain (name/id/type) to bm-inventory-configmap file.

Not included in this PR:
- UI/API changes.
- DNS server mock for functional tests: might want to consider
  later on.
- Validation for resolvability after creating the record sets:
  not sure we want this, as a fallback is merely trying manually again.

[1] List of supported domains are defined in bm-inventory-configmap
    along with dns provider type.
[2] https://github.com/danielerez/go-dns-client
    - Can be used also as a command line tool.

### DIFF
--- a/Dockerfile.bm-inventory
+++ b/Dockerfile.bm-inventory
@@ -1,4 +1,8 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest as certs
+
 FROM scratch
+COPY --from=certs /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-bundle.crt
+COPY --from=certs /etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt /etc/ssl/certs/ca-bundle.trust.crt
 ARG GIT_REVISION
 LABEL "git_revision"=${GIT_REVISION}
 ADD build/bm-inventory /bm-inventory
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -9,20 +9,20 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"regexp"
 	"strings"
 	"sync"
 	"text/template"
 	"time"
 
-	"github.com/filanov/bm-inventory/internal/network"
-	"k8s.io/apimachinery/pkg/api/resource"
-
+	"github.com/danielerez/go-dns-client/pkg/dnsproviders"
 	"github.com/filanov/bm-inventory/internal/cluster"
 	"github.com/filanov/bm-inventory/internal/cluster/validations"
 	"github.com/filanov/bm-inventory/internal/common"
 	"github.com/filanov/bm-inventory/internal/events"
 	"github.com/filanov/bm-inventory/internal/host"
 	"github.com/filanov/bm-inventory/internal/installcfg"
+	"github.com/filanov/bm-inventory/internal/network"
 	"github.com/filanov/bm-inventory/models"
 	"github.com/filanov/bm-inventory/pkg/filemiddleware"
 	"github.com/filanov/bm-inventory/pkg/job"
@@ -39,6 +39,7 @@ import (
 	"github.com/sirupsen/logrus"
 	batch "k8s.io/api/batch/v1"
 	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 )
@@ -61,22 +62,23 @@ var (
 )
 
 type Config struct {
-	ImageBuilder        string `envconfig:"IMAGE_BUILDER" default:"quay.io/ocpmetal/installer-image-build:latest"`
-	ImageBuilderCmd     string `envconfig:"IMAGE_BUILDER_CMD" default:"echo hello"`
-	AgentDockerImg      string `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/ocpmetal/agent:latest"`
-	KubeconfigGenerator string `envconfig:"KUBECONFIG_GENERATE_IMAGE" default:"quay.io/ocpmetal/ignition-manifests-and-kubeconfig-generate:latest"`
-	InventoryURL        string `envconfig:"INVENTORY_URL" default:"10.35.59.36"`
-	InventoryPort       string `envconfig:"INVENTORY_PORT" default:"30485"`
-	S3EndpointURL       string `envconfig:"S3_ENDPOINT_URL" default:"http://10.35.59.36:30925"`
-	S3Bucket            string `envconfig:"S3_BUCKET" default:"test"`
-	AwsAccessKeyID      string `envconfig:"AWS_ACCESS_KEY_ID" default:"accessKey1"`
-	AwsSecretAccessKey  string `envconfig:"AWS_SECRET_ACCESS_KEY" default:"verySecretKey1"`
-	Namespace           string `envconfig:"NAMESPACE" default:"assisted-installer"`
-	UseK8s              bool   `envconfig:"USE_K8S" default:"true"` // TODO remove when jobs running deprecated
-	JobCPULimit         string `envconfig:"JOB_CPU_LIMIT" default:"500m"`
-	JobMemoryLimit      string `envconfig:"JOB_MEMORY_LIMIT" default:"1000Mi"`
-	JobCPURequests      string `envconfig:"JOB_CPU_REQUESTS" default:"300m"`
-	JobMemoryRequests   string `envconfig:"JOB_MEMORY_REQUESTS" default:"400Mi"`
+	ImageBuilder        string            `envconfig:"IMAGE_BUILDER" default:"quay.io/ocpmetal/installer-image-build:latest"`
+	ImageBuilderCmd     string            `envconfig:"IMAGE_BUILDER_CMD" default:"echo hello"`
+	AgentDockerImg      string            `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/ocpmetal/agent:latest"`
+	KubeconfigGenerator string            `envconfig:"KUBECONFIG_GENERATE_IMAGE" default:"quay.io/ocpmetal/ignition-manifests-and-kubeconfig-generate:stable"` // TODO: update the latest once the repository has git workflow
+	InventoryURL        string            `envconfig:"INVENTORY_URL" default:"10.35.59.36"`
+	InventoryPort       string            `envconfig:"INVENTORY_PORT" default:"30485"`
+	S3EndpointURL       string            `envconfig:"S3_ENDPOINT_URL" default:"http://10.35.59.36:30925"`
+	S3Bucket            string            `envconfig:"S3_BUCKET" default:"test"`
+	AwsAccessKeyID      string            `envconfig:"AWS_ACCESS_KEY_ID" default:"accessKey1"`
+	AwsSecretAccessKey  string            `envconfig:"AWS_SECRET_ACCESS_KEY" default:"verySecretKey1"`
+	Namespace           string            `envconfig:"NAMESPACE" default:"assisted-installer"`
+	UseK8s              bool              `envconfig:"USE_K8S" default:"true"` // TODO remove when jobs running deprecated
+	BaseDNSDomains      map[string]string `envconfig:"BASE_DNS_DOMAINS" default:""`
+	JobCPULimit         string            `envconfig:"JOB_CPU_LIMIT" default:"500m"`
+	JobMemoryLimit      string            `envconfig:"JOB_MEMORY_LIMIT" default:"1000Mi"`
+	JobCPURequests      string            `envconfig:"JOB_CPU_REQUESTS" default:"300m"`
+	JobMemoryRequests   string            `envconfig:"JOB_MEMORY_REQUESTS" default:"400Mi"`
 }
 
 const ignitionConfigFormat = `{
@@ -327,6 +329,10 @@ func (b *bareMetalInventory) DeregisterCluster(ctx context.Context, params insta
 			WithPayload(common.GenerateError(http.StatusNotFound, err))
 	}
 
+	if err := b.deleteDNSRecordSets(ctx, cluster); err != nil {
+		log.Warnf("failed to delete DNS record sets for base domain: %s", cluster.BaseDNSDomain)
+	}
+
 	err := b.clusterApi.DeregisterCluster(ctx, &cluster)
 	if err != nil {
 		log.WithError(err).Errorf("failed to deregister cluster cluster %s", params.ClusterID)
@@ -567,6 +573,11 @@ func (c clusterInstaller) install(tx *gorm.DB) error {
 		return err
 	}
 
+	if err = c.b.createDNSRecordSets(c.ctx, cluster); err != nil {
+		return common.NewApiError(http.StatusInternalServerError,
+			fmt.Errorf("failed to create DNS record sets for base domain: %s", cluster.BaseDNSDomain))
+	}
+
 	if err = c.b.clusterApi.Install(c.ctx, &cluster, tx); err != nil {
 		return common.NewApiError(http.StatusConflict, errors.Wrapf(err, "failed to install cluster %s", cluster.ID.String()))
 	}
@@ -614,6 +625,9 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
 	err = b.db.Transaction(cInstaller.install)
 	if err != nil {
 		log.WithError(err).Warn("Cluster install")
+		if errDelete := b.deleteDNSRecordSets(ctx, cluster); errDelete != nil {
+			log.Warnf("failed to delete DNS record sets for base domain: %s", cluster.BaseDNSDomain)
+		}
 		return common.GenerateErrorResponder(err)
 	}
 	if err = b.db.Preload("Hosts").First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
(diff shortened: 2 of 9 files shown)


internal/bminventory/inventory.go
```diff
@@ -1462,3 +1477,79 @@ func setPullSecret(cluster *common.Cluster, pullSecret string) {
...
+	case "route53":
+		var dnsProvider dnsproviders.Provider = dnsproviders.Route53{
+			RecordSet: dnsproviders.RecordSet{
+				HostedZoneID:  dnsDomainID,
+				RecordSetType: "A",
+				TTL:           60,
+			},
+		}
+
+		dnsRecordSetFunc := dnsProvider.UpdateRecordSet
```
ori-amizur, line 1511: Are we should we want to update? This will override an existing record-set if it already exists

internal/bminventory/inventory.go
```diff
@@ -1462,3 +1477,79 @@ func setPullSecret(cluster *common.Cluster, pullSecret string) {
...
+	return nil
+}
+
+func (b *bareMetalInventory) getDNSDomain(baseDNSDomain string) (string, string, error) {
+	var dnsDomainID string
+	var dnsProviderType string
+
+	// Parse base domains from config
+	if val, ok := b.Config.BaseDNSDomains[baseDNSDomain]; ok {
+		s := strings.Split(val, "/")
```
ori-amizur, line 1543: Maybe use net.ParseCIDR() ?

internal/bminventory/inventory.go
```diff
@@ -1462,3 +1477,79 @@ func setPullSecret(cluster *common.Cluster, pullSecret string) {
...
+			},
+		}
+
+		dnsRecordSetFunc := dnsProvider.UpdateRecordSet
+		if delete {
+			dnsRecordSetFunc = dnsProvider.DeleteRecordSet
+		}
+
+		// Update/Delete A record for API Virtual IP
+		recordSetName := fmt.Sprintf("%s.%s.%s", "api", cluster.Name, cluster.BaseDNSDomain)
```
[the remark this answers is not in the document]
danielerez: So you think we should add proper fqdn name to cluster validation? Or, do we need to use some other unique cluster attribute?

We validate the the base domain in 'getDNSDomain' function. I.e., retrieving the default domains from the config and find the domain according to the specified one (cluster.BaseDNSDomain).

As for the uniqueness, will try to add validation for dns existence first.

internal/bminventory/inventory.go
danielerez: does it work for non valid addresses(as example.com/route53)? will try :)

internal/bminventory/inventory.go
```diff
@@ -1462,3 +1477,79 @@ func setPullSecret(cluster *common.Cluster, pullSecret string) {
...
+		return err
+	}
+	if dnsDomainID == "" || dnsProviderType == "" {
+		// No default base domains
+		return nil
+	}
+
+	switch dnsProviderType {
+	case "route53":
+		var dnsProvider dnsproviders.Provider = dnsproviders.Route53{
```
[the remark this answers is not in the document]
ori-amizur: Just a suggestion, maybe it is not.  @avishayt what do you think

internal/bminventory/inventory.go
```diff
@@ -585,6 +593,13 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
 		return common.GenerateErrorResponder(err)
 	}
 
+	err = b.createDNSRecordSets(ctx, cluster)
```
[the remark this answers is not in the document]
ori-amizur: The vips are verified on line 470,  and the function is called from line 540.
In general we have a database transaction, starts on line 603.  It commits or rollbacks depending if the function "install" succeeded or failed.

internal/bminventory/inventory.go
ori-amizur: Probably not

internal/bminventory/inventory.go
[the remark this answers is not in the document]
ori-amizur: We need to avoid the user making mistakes, that they will not be rejected by route53.
The first 2 issues route53 will reject in case of these errors, so we will have to handle them before.
The 3rd issue is the dns-name uniqueness. We must find a way to verify it (maybe unique index in the database).

internal/bminventory/inventory.go
ronniel1: Maybe can use url.Parse instead

Dockerfile.bm-inventory
```diff
@@ -1,4 +1,4 @@
-FROM scratch
+FROM alpine:3.12
```
ronniel1, line 1: Why is this needed?

danielerez: This is required for aws-sdk-go[*] secure api invocation (using ca-certificate package).
Without it, getting the following error on a request to route53:
"RequestError: send request failed caused by: Post "https://route53.amazonaws.com/***/hostedzone/***/rrset/": x509: certificate signed by unknown authority"

[*] https://github.com/aws/aws-sdk-go

internal/bminventory/inventory.go
danielerez: no luck with url.Parse also, just used regexp instead.

internal/bminventory/inventory.go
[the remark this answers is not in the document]
danielerez: got it, moved to 'install' then.

internal/bminventory/inventory.go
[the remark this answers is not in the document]
danielerez: Seems cluster name validation[1] is even more restrictive than route53[2], so it should be ok, as we validate cluster name first.
[1] https://github.com/mareklibra/facet-lib/blob/d3b53a7bf8fca732b4459ff21981fe6172f162f6/src/components/ui/formik/validationSchemas.ts#L11
[2] https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html#domain-name-format-hosted-zones

For the uniqueness issue, added a proper check to validation phase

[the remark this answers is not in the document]
ori-amizur: What about a validation that the dns-name is sub-domain of the hosted zone name?  Route53 will fail if this is not the case

Dockerfile.bm-inventory
filanov: Not if you set the credentials with environment variables. we already use the sdk to create the bucket in s3.

internal/cluster/validations/validations.go
```diff
@@ -85,3 +87,20 @@ func ValidatePullSecret(secret string) error {
 	*/
 	return nil
 }
+
+func ValidateDNSRecords(names []string, dnsDomainID string) error {
+	for _, name := range names {
+		res, err := dnsproviders.Route53{
+			RecordSet: dnsproviders.RecordSet{
+				HostedZoneID: dnsDomainID,
```
ori-amizur, line 95: Are you using this route53 command https://docs.aws.amazon.com/Route53/latest/APIReference/API_ListResourceRecordSets.html to get the record set?  Are you doing any manipulation on it?

internal/bminventory/inventory.go
```diff
@@ -1517,3 +1540,108 @@ func setPullSecret(cluster *common.Cluster, pullSecret string) {
...
+	if dnsDomainID == "" || dnsProvider == "" {
+		// Specified domain is not defined in config
+		return nil, nil
+	}
+
+	return &dnsDomain{
+		Name:           baseDNSDomainName,
+		ID:             dnsDomainID,
+		Provider:       dnsProvider,
+		APIAddress:     fmt.Sprintf("%s.%s.%s", "api", clusterName, baseDNSDomainName),
```
ori-amizur, line 1635: I think that variable names APIDomainName and IngressDomainName will be more suitable

internal/cluster/validations/validations.go
danielerez: Yes, to dns-client code is implemented separately in another repo:
https://github.com/danielerez/go-dns-client/blob/master/pkg/dnsproviders/aws-route53.go
So I am doing some manipulation, abstraction, etc. Do you prefer including the client code under pkg?

internal/bminventory/inventory.go
danielerez: sure, makes sense.

Dockerfile.bm-inventory
filanov: I prefer to have this image as light as possible.

danielerez: ok, I'll give it a try. But i guess we'll have to use other env vars, as you're already using the default ones for s3. Or is it actually used in different containers?

romfreiman: I think there are default redhat images that should be used. we should ask Michael Hrivnak or his team

djzager: `scratch` is certainly best, next best is `registry.access.redhat.com/ubi8/ubi-minimal:latest`.

I see in the baremetal-operator they are building images using [`FROM registry.svc.ci.openshift.org/openshift/release:golang-1.14 AS builder`](https://github.com/metal3-io/baremetal-operator/blob/master/build/Dockerfile#L1) but I forget what exactly is needed in order to be able to access them.

**tl;dr** `scratch` is best, then language based images, then `ubi8`.

danielerez: Getting the same error using env vars and static credentials :( maybe since it's a remote service.
The advantage of alpine is size, it's the smallest image (<3MB).

djzager: I don't know enough about this to jump in. At a minimum though you should look at using the `registry.access.redhat.com/ubi8/ubi-minimal:latest` instead of Alpine.

djzager: Looking at https://medium.com/on-docker/use-multi-stage-builds-to-inject-ca-certs-ad1e8f01de1b makes me think it's possible to use multi-stage builds to still use `scratch` and include the necessary certificate authority stuffs.

filanov: I'm using the default names but set the explicitly to the api https://github.com/filanov/bm-inventory/blob/master/pkg/s3wrapper/client.go#L52

danielerez: > Looking at https://medium.com/on-docker/use-multi-stage-builds-to-inject-ca-certs-ad1e8f01de1b makes me think it's possible to use multi-stage builds to still use `scratch` and include the necessary certificate authority stuffs.

nice! it indeed works :)

internal/bminventory/inventory.go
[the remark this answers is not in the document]
danielerez: > What about a validation that the dns-name is sub-domain of the hosted zone name? Route53 will fail if this is not the case

I'm actually creating the DNS recordset only if the specified domain is one of predefined default base domains (in 'getDNSDomain' func). I.e., building the full domain name only after validating that the base domain is supported.

Dockerfile.bm-inventory
```diff
@@ -1,4 +1,7 @@
+FROM alpine:3.12 as certs
```
djzager, line 1: I don't think alpine is going to work from the perspective of wanting to include this project in OpenShift build pipelines. However, I think we can tackle that problem later.

djzager: Went and had a look at the ubi8 image and there are certs in there that you could use:

```
FROM registry.redhat.io/ubi8/ubi-minimal:latest as certs
# alternatively
# FROM registry.access.redhat.com/ubi8/ubi-minimal:latest

...
COPY --from=certs /etc/ssl/certs /etc/ssl/certs
```

internal/bminventory/inventory.go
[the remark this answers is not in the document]
ori-amizur: The fact that you get the dns-domain from local configuration does not mean that it is the route53 hosted zone domain.
I think you need to perform get hosted zone operation from route53 and verify that our domain-names are sub-domains of the domain represented by the hosted zone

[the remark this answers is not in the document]
ori-amizur: Concerning the domain-name validation, we usually verify validity in the backend even if there are similar validations done in the UI.  So I think it should be validated in the backend too. @ronniel1 what do you think?

Dockerfile.bm-inventory
danielerez: > Went and had a look at the ubi8 image and there are certs in there that you could use:
> 
> ```
> FROM registry.redhat.io/ubi8/ubi-minimal:latest as certs
> # alternatively
> # FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
> 
> ...
> COPY --from=certs /etc/ssl/certs /etc/ssl/certs
> ```

Gave it a try, but no luck as the cert files are empty in this image :/

internal/bminventory/inventory.go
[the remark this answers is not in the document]
danielerez: isn't it already handled as part of cluster name validation in backend?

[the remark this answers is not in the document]
danielerez: > The fact that you get the dns-domain from local configuration does not mean that it is the route53 hosted zone domain.
> I think you need to perform get hosted zone operation from route53 and verify that our domain-names are sub-domains of the domain represented by the hosted zone

ok, added a validation that the domain name in config is indeed correlated to the hosted zone.

internal/cluster/validations/validations.go
```diff
@@ -85,3 +88,62 @@ func ValidatePullSecret(secret string) error {
...
+	return validateBaseDNS(dnsDomainName, dnsDomainID, dnsProvider)
+}
+
+func validateBaseDNS(dnsDomainName, dnsDomainID string, dnsProvider dnsproviders.Provider) error {
+	dnsNameFromService, err := dnsProvider.GetDomainName()
+	if err != nil {
+		return fmt.Errorf("Can't validate base DNS domain: %v", err)
+	}
+
+	if !strings.HasSuffix(dnsDomainName, dnsNameFromService) {
```
ori-amizur, line 113: 1. I think that route53 may add a dot "." at the end of the domain-name/dns-name.  I am not sure we are verifying that our domain names are ending with ".".  I think we need to transform to a single form and then to compare.
2. strings.HasSuffix is not exact.  For example b.c.  is the suffix of a.bb.c. It is an edge case but if you want to be exact, you will probably need to split by "."  and then compare parts

internal/cluster/validations/validations.go
```diff
@@ -85,3 +88,62 @@ func ValidatePullSecret(secret string) error {
...
+			HostedZoneID: dnsDomainID,
+			SharedCreds:  true,
+		}
+	default:
+		return nil
+	}
+	return validateDNSRecords(names, dnsProvider)
+}
+
+func validateDNSRecords(names []string, dnsProvider dnsproviders.Provider) error {
```
ori-amizur, line 138: The function doesn't validate the records, but verifies that the record-set doesn't exist.  Maybe change name?

internal/cluster/validations/validations.go
ori-amizur: Already saw that the suffix of the dnsDomainName is trimmed.  We just need to verify that the "dnsNameFromService" also does not allow trailing "".".  Currently I didn't find any restriction on it in the backend.

Dockerfile.bm-inventory
djzager:
```bash
$ docker run -it --entrypoint /bin/bash registry.redhat.io/ubi8/ubi-minimal:latest
[root@9afe7d57cc5b /]# ls /etc/ssl/certs/
ca-bundle.crt  ca-bundle.trust.crt
[root@9afe7d57cc5b /]#
```

djzager:
```bash
[root@9afe7d57cc5b /]# wc -l /etc/ssl/certs/**
  3856 /etc/ssl/certs/ca-bundle.crt
  4477 /etc/ssl/certs/ca-bundle.trust.crt
  8333 total
```

internal/cluster/validations/validations.go
danielerez: right, removed the trailing dot in the dns-client. handled the domains comparing issue.

internal/cluster/validations/validations.go
danielerez: sure, renamed - hope it looks better :/

Dockerfile.bm-inventory
danielerez: > ```shell
> [root@9afe7d57cc5b /]# wc -l /etc/ssl/certs/**
>   3856 /etc/ssl/certs/ca-bundle.crt
>   4477 /etc/ssl/certs/ca-bundle.trust.crt
>   8333 total
> ```

but these are symbolic links, created when running the container.
we want to just copy them without actually include the image...

danielerez: > > ```shell
> > [root@9afe7d57cc5b /]# wc -l /etc/ssl/certs/**
> >   3856 /etc/ssl/certs/ca-bundle.crt
> >   4477 /etc/ssl/certs/ca-bundle.trust.crt
> >   8333 total
> > ```
> 
> but these are symbolic links, created when running the container.
> we want to just copy them without actually include the image...

ok, followed symlinks and used certs from /etc/pki/ca-trust/extracted/

PR #283, opened 2020-06-22
MGMT-1016  bm-inventory configuration specifically for subsystem

oshercc:


tools/deploy_assisted_installer.py
```diff
@@ -1,26 +1,34 @@
...
+TEST_HOST_MONITOR_INTERVAL = "1s"
 
 def main():
     with open(SRC_FILE, "r") as src:
-        with open(DST_FILE, "w+") as dst:
-            data = src.read()
+            data = yaml.safe_load(src)
             if args.deploy_tag is not "":
-                data = data.replace("REPLACE_IMAGE", "quay.io/ocpmetal/bm-inventory:{}".format(args.deploy_tag))
+                data["spec"]["template"]["spec"]["containers"][0]["image"] = args.deploy_tag
```
ronniel1, line 21: This is not the same thing as before.
Before we added the default service name with ':TAG' at the end. Now we are just adding the tag

oshercc, line 21: fix the return value to "quay.io/ocpmetal/bm-inventory:{}".format(args.deploy_tag)

tools/wait_for_pod.py
```diff
@@ -0,0 +1,53 @@
...
+            sys.exit("pods app name {} did not enter {} state within {} seconds".format(args.app, args.state, TIMEOUT))
+        time.sleep(1)
+
+    print("pod app {} is in {} status".format(args.app, args.state))
+    return
+
+def is_pod_in_state():
+
+        state_keys = get_pod_state()
+        print(state_keys)
```
ronniel1, line 31: Is this for debug? If so, maybe we should not push it?

tools/wait_for_pod.py
```diff
@@ -0,0 +1,53 @@
...
+
+        if not args.state in state_keys:
+            return False
+
+        # validation check
+        time.sleep  (5)
+        state_keys = get_pod_state()
+        if args.state in state_keys:
+            return True
```
ronniel1, line 40: Why do we do get_pod state twice, instead of relying on the loop in main?

oshercc: a lot of the deployments enter running state and then fails after a few seconds, this waits for the pod to enter running state and make sure it is still running after 5 seconds.

filanov:
@oshercc please sync with @YuviGold, you are touching the same things.

oshercc:
@filanov added build to the new target.

tools/wait_for_pod.py
ronniel1: Please add a comment on why this is done

Makefile
```diff
@@ -91,6 +94,10 @@ deploy-role: deploy-namespace
 deploy-mariadb: deploy-namespace
 	python3 ./tools/deploy_mariadb.py
 
+subsystem-deployment: build create-build-dir deploy-namespace deploy-mariadb deploy-s3 deploy-service-test-mode deploy-expirer
```
filanov, line 97: build just create a new binary it not creating a new docker image

ronniel1: also you should not depend on the order of the pre-requisites here

PR #286, opened 2020-06-23
MGMT-1229 add resources limitation to image generation and kubeconfig generation

filanov:
I set the default values because otherwise the operations will be slower.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -14,23 +14,20 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/filanov/bm-inventory/internal/cluster/validations"
-
-	"github.com/filanov/bm-inventory/restapi"
-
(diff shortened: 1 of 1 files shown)


internal/bminventory/inventory.go
```diff
@@ -177,6 +179,16 @@ func (b *bareMetalInventory) createImageJob(jobName, imgName, ignitionConfig str
 				Spec: core.PodSpec{
 					Containers: []core.Container{
 						{
+							Resources: core.ResourceRequirements{
+								Limits: core.ResourceList{
+									"cpu":    getQuantity("500m"),
+									"memory": getQuantity("1000Mi"),
```
ronniel1, line 185: Shouldn't these values come from environment variables?

filanov: Maybe it should. will fix.

PR #289, opened 2020-06-23
Update drone

oshercc:
Update drone to run when:
        - new PR
	- push to master


.drone.yml
```diff
@@ -21,6 +21,13 @@ steps:
       DB_PORT: 3306
       USE_K8S: false
 
+trigger:
```
ronniel1, line 24: Does this change mean that it will run on any pr against the master and on push to master?

PR #290, opened 2020-06-23
deploy-test deploys bm-invetory image to minikube to ease testing

YuviGold:
Instead of updating quay.io user repository every time
Allowing `make deploy-test` to deploy bm-inventory local code to minikube to be tested


Makefile
```diff
@@ -91,8 +96,22 @@ deploy-role: deploy-namespace
 deploy-mariadb: deploy-namespace
 	python3 ./tools/deploy_mariadb.py
 
+deploy-test:
+	export SERVICE=quay.io/ocpmetal/bm-inventory:test && export LOCAL_IMAGES=--local && \
+	$(MAKE) _build_inventory_in_minikube && \
+	skipper make deploy-all
```
tsorya, line 102: u don't run skipper from skipper

Makefile
```diff
@@ -91,8 +96,22 @@ deploy-role: deploy-namespace
...
+
+########
+# Test #
+########
+
 subsystem-run: test subsystem-clean
 
+_build_inventory_in_minikube:
+	skipper make build
```
tsorya, line 111: same, 
target is ran as skipper make . don't use here skipper inside Make

README.md
```diff
@@ -40,15 +40,35 @@ After every change in the API (`swagger.yaml`) the code should be generated and
...
-Running the tests:
+#### Pre-configuration
+  - Run minikube on your system.
+  - Deploy services `make deploy-test`
 
-`skipper make subsystem-run`
+### Run system tests
+
+```shell script
+make test
```
filanov, line 52: the point of having a build container is to run everything inside it. so all the commands are `skipper make ...`

Makefile
YuviGold: I know.
I don't want to enforce the usage of `skipper` for every `make` run. It should be encapsulated and transparent for him.
As I updated the README accordingly - the usage is `make deploy-test`.

@filanov What do you think?

filanov: Not sure myself. We are using build containers so the user will have minimal requirements to install. 
So even if you need to run python scripts you will not need to install python. 
Mixing skipper with non skipper targets may be confusing. I agree that having skipper inside the makefile looks cleaner.
@ronniel1 what do you think?

ronniel1: right now, in bm-inventory, the way to invoke the makefile is with skipper.
As a user, how will I know when to use skipper and when not to?

YuviGold: > right now, in bm-inventory, the way to invoke the makefile is with skipper.
> As a user, how will I know when to use skipper and when not to?

I think that as a contributor you should not use skipper by default as any other user.
For a newcomer to the project - skipper is a mystery. Just intermediate developers should acknowledge.
Have a look at https://3musketeers.io/ methodology (replace docker-compose with skipper).

ronniel1: That is a valid way to look at it, and something that we should discuss.
However, IMHO the makefile usage should be consistent. Either we use skipper to invoke, or not.
This should not be done this way,
We have a convention, and its possible to change it, but it should be handled as a convention change, after discussion.

Makefile
```diff
@@ -91,8 +96,22 @@ deploy-role: deploy-namespace
 deploy-mariadb: deploy-namespace
 	python3 ./tools/deploy_mariadb.py
 
+deploy-test:
```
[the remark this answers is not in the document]
YuviGold: I talked with him about the PR #283.
We touch the same subsystem tests area but on different aspects.
He made changes to the monitoring of the pods to understand when the pod is ready to be tested (because it crashes multiple times until it succeeds running), while I made a change to run the subsystem test w/o pushing or pulling from a image registry.

After the other PR will be pushed i'll rebase this PR above of it.

Makefile
YuviGold: OK. Changed to skipper default usage.

YuviGold:
Rebased @oshercc and changed the default behavior to `skipper make ...`

PR #293, opened 2020-06-24
mgmt 1143 - enlarge the column for event message

ronniel1:


swagger.yaml
```diff
@@ -760,6 +760,7 @@ definitions:
         x-go-custom-tag: gorm:"index"
       message:
         type: string
+        x-go-custom-tag: gorm:"type:varchar(4096)"
```
filanov, line 763: suggesting making it a text.

ronniel1: I don't think so.
TEXT has large performance overhead
and events that are larger than 4096 might not be such a good idea.
Maybe better shorten the event message

PR #294, opened 2020-06-25
Ori/mgmt 1236 Perform vips validation that they belong to free ip list

ori-amizur:
Also support query for subnet's free ip list


avishayt:
I thought we said that we will do the validation on update+install cluster operations?  Why do we still need the list of free APIs?

internal/bminventory/inventory.go
```diff
@@ -1809,3 +1811,66 @@ func (b *bareMetalInventory) validateDNSRecords(clusterName string, domain *dnsD
...
+	sort.Slice(ret, func(i, j int) bool {
+		return ipAsUint(ret[i].String(), log) < ipAsUint(ret[j].String(), log)
+	})
+
+	ret = applyLimit(ret, params.Limit)
+
+	return installer.NewGetFreeAddressesOK().WithPayload(ret)
+}
+
+func (b *bareMetalInventory) GetFreeAddresses(ctx context.Context, params installer.GetFreeAddressesParams) middleware.Responder {
```
filanov, line 1867: I think all those functions should return an error not `middleware.Responder`
The main functions should take the error and if it's type is `common.ApiErrorResponse` then return it else generate default error.

internal/bminventory/inventory.go
```diff
@@ -1809,3 +1811,66 @@ func (b *bareMetalInventory) validateDNSRecords(clusterName string, domain *dnsD
...
+}
+
+func applyLimit(ret models.FreeAddressesList, limitParam *int64) models.FreeAddressesList {
+	if limitParam != nil && *limitParam >= 0 && *limitParam < int64(len(ret)) {
+		return ret[:*limitParam]
+	}
+	return ret
+}
+
+func (b *bareMetalInventory) getFreeAddresses(params installer.GetFreeAddressesParams, log logrus.FieldLogger) middleware.Responder {
```
filanov, line 1841: It's strange internal functions return `middleware.Responder` i expect internal functions to return `error`
The error can be http error (error_utils) but i think that only api functions should return api reply

filanov:
If we plan to set cluster state based on validations then this validations should be part of refresh cluster right?

swagger.yaml
```diff
@@ -760,6 +760,49 @@ paths:
           schema:
             $ref: '#/definitions/error'
 
+  /clusters/{cluster_id}/free_addresses:
```
ronniel1, line 763: Please add here a remark that this is an admin only operation

internal/bminventory/inventory.go
```diff
@@ -1809,3 +1811,66 @@ func (b *bareMetalInventory) validateDNSRecords(clusterName string, domain *dnsD
...
+	var hosts []*models.Host
+	err := b.db.Select("free_addresses").Find(&hosts, "cluster_id = ? and status in (?)", params.ClusterID.String(), []string{host.HostStatusInsufficient, host.HostStatusKnown}).Error
```
ronniel1, line 1843: If this was a public API then the limit should be applied to the DB query.
No sense in fetching 8k roes from DB if we only need 1.

But sense this will not really be used, only for debugging, I guess it does not really matter

internal/network/machine_network_cidr.go
```diff
@@ -135,3 +144,85 @@ func IsHostInMachineNetCidr(log logrus.FieldLogger, cluster *common.Cluster, hos
...
+	var (
+		availableFreeAddresses []string
+		sets                   = make([]IPSet, 0)
+		resultingSet           = make(IPSet)
+	)
+	for _, h := range hosts {
+		if h.FreeAddresses != "" {
+			availableFreeAddresses = append(availableFreeAddresses, h.FreeAddresses)
+		}
+	}
```
ronniel1, line 194: Why do we need this intermediate array?
Why not add them all to the set, and then check if the set is empty?

internal/network/machine_network_cidr.go
```diff
@@ -64,11 +67,17 @@ func ipInCidr(ipStr, cidrStr string) bool {
...
+func verifyVip(hosts []*models.Host, machineNetworkCidr string, vip string, vipName string, mustExist bool, log logrus.FieldLogger) error {
+	if !mustExist && vip == "" {
 		return nil
 	}
-	return fmt.Errorf("%s <%s> does not belong to machine-network-cidr <%s>", vipName, vip, machineNetworkCidr)
+	if !ipInCidr(vip, machineNetworkCidr) {
+		return fmt.Errorf("%s <%s> does not belong to machine-network-cidr <%s>", vipName, vip, machineNetworkCidr)
+	}
+	if !IpInFreeList(hosts, vip, machineNetworkCidr, log) {
+		return fmt.Errorf("%s <%s> is not in free list for machine network cidr %s", vipName, vip, machineNetworkCidr)
```
ronniel1, line 78: change message to: "%s <%s> is already in use"

internal/bminventory/inventory.go
```diff
@@ -1809,3 +1811,66 @@ func (b *bareMetalInventory) validateDNSRecords(clusterName string, domain *dnsD
...
+	resultingSet := network.MakeFreeAddressesSet(hosts, params.Network, params.Prefix, log)
+
+	ret := models.FreeAddressesList{}
+	for a := range resultingSet {
+		ret = append(ret, a)
+	}
+
+	// Sort addresses
+	sort.Slice(ret, func(i, j int) bool {
+		return ipAsUint(ret[i].String(), log) < ipAsUint(ret[j].String(), log)
```
ronniel1, line 1859: Why not just sort the IP string?

ori-amizur: If you will sort string then 10.0.0.3 will be after 10.0.0.200.  Seemed more logical to me.

internal/network/machine_network_cidr.go
ori-amizur: OK

swagger.yaml
```diff
@@ -1267,6 +1310,12 @@ definitions:
         format: date-time
         x-go-custom-tag: gorm:"type:datetime"
 
+  free-addresses-list:
+    type: array
+    items:
+      type: string
+      format: ipv4
```
eranco74, line 1317: This format exists?
In the rest of the swagger we use this pattern:
`pattern: '^(([0-9]{1,3}\.){3}[0-9]{1,3})?$'`

internal/bminventory/inventory.go
ori-amizur: In the database it is JSON string.  Need to unmarshal each one and create sets, intersect between the sets, and only then to apply the limit.

internal/network/machine_network_cidr.go
ori-amizur: We need to intersect between the sets, not unify them.

swagger.yaml
ori-amizur: OK

ori-amizur: OK

swagger.yaml
ori-amizur: Yes.  The format you mentioned is a string that can also be empty.

internal/bminventory/inventory.go
ori-amizur: OK. Changed

ronniel1:
> I thought we said that we will do the validation on update+install cluster operations? Why do we still need the list of free APIs?

By mistake, this was implemented.
I thought that instead of removing something that might be helpfull for debugging, we should keep it and when we have authorization, we'll make this an admin-only API, instead.
@avishayt  What do you think?

internal/bminventory/inventory.go
ronniel1: https://stackoverflow.com/questions/48389585/golang-sorting-slice-of-ip-addresses-as-strings
First answer might be a simpler way to do this, by it does not really matter

PR #300, opened 2020-06-28
Raz/revert cancel tests subsystem

RazRegev:


filanov:
Please revert the code as well. 
Disabling the tests is a bad habit.

RazRegev:
> Please revert the code as well.
> Disabling the tests is a bad habit.

reverted

PR #305, opened 2020-06-29
Event improvements

avishayt:
NOT READY

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
+	"github.com/filanov/bm-inventory/internal/events"
 	"github.com/filanov/bm-inventory/internal/validators"
 
 	"github.com/filanov/bm-inventory/internal/common"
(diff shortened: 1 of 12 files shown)


internal/host/host.go
```diff
@@ -292,3 +303,11 @@ func (m *Manager) CancelInstallation(ctx context.Context, h *models.Host, reason
 	}
 	return nil
 }
+
+func (m *Manager) GetHostname(host *models.Host) string {
+	var hwInfo models.Inventory
+	if err := json.Unmarshal([]byte(host.Inventory), &hwInfo); err != nil {
+		return "<unknown>"
```
ronniel1, line 310: If we could not get the name of the host, maybe we should add the host-id to the event?

In the first states of the host, this is what will happen

internal/host/host_test.go
```diff
@@ -93,15 +96,19 @@ var _ = Describe("statemachine", func() {
...
+		host       models.Host
+		ctrl       *gomock.Controller
+		mockEvents *events.MockHandler
 	)
 
 	BeforeEach(func() {
 		db = prepareDB()
-		state = NewManager(getTestLog(), db, nil, nil, nil)
+		ctrl = gomock.NewController(GinkgoT())
+		mockEvents = events.NewMockHandler(ctrl)
```
filanov, line 110: no need to add mocks if they are not used.

PR #308, opened 2020-06-29
Raz/mgmt 1205

RazRegev:


filanov:
Impossible to review. please rebase above master

internal/host/resetcmd.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/filanov/bm-inventory/models"
+)
+
+type resetCmd struct {
```
gamli75, line 11: I think that resetAgendCmd is better

internal/host/instructionmanager.go
```diff
@@ -66,6 +67,7 @@ func NewInstructionManager(log logrus.FieldLogger, db *gorm.DB, hwValidator hard
 			HostStatusDiscovering:  {[]CommandGetter{hwCmd, inventoryCmd, connectivityCmd}, defaultNextInstructionInSec},
 			HostStatusInstalling:   {[]CommandGetter{installCmd}, defaultBackedOffInstructionInSec},
 			HostStatusDisabled:     {[]CommandGetter{}, defaultBackedOffInstructionInSec},
+			HostStatusResetting:    {[]CommandGetter{resetCmd}, defaultBackedOffInstructionInSec},
```
gamli75, line 70: @avishayt I saw that if the host is in HostStatusResetting - we will display "resetting" to the user - I think it's very confusing since it's "agent resetting" and not "host resetting"

subsystem/cluster_test.go
```diff
@@ -636,6 +636,8 @@ var _ = Describe("cluster install", func() {
 				Expect(swag.StringValue(c.Status)).Should(Equal(models.ClusterStatusInsufficient))
 				for _, host := range c.Hosts {
 					Expect(swag.StringValue(host.Status)).Should(Equal(models.HostStatusResetting))
+					_, ok := getStepInList(getNextSteps(clusterID, *host.ID), models.StepTypeResetAgent)
+					Expect(ok).Should(Equal(true))
```
ori-amizur, line 640: It will be better if you check that the host moves to discovering after Register

PR #311, opened 2020-06-29
MGMT-1126 make cluster install async

filanov:
Cluster install will move cluster to `preparing-for-installation` after some basic validations, the user will get an ok reply, after that the async part of the API will start and move the cluster and the hosts to installing right after the ignition is generated.

If the service will be rebooted then the cluster will be left in `preparing-for-installation` state, handling this issue will solved with https://issues.redhat.com/browse/MGMT-1349

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -252,7 +252,7 @@ func (b *bareMetalInventory) formatIgnitionFile(cluster *common.Cluster, params 
 		return "", err
 	}
 	buf := &bytes.Buffer{}
-	if err := tmpl.Execute(buf, ignitionParams); err != nil {
+	if err = tmpl.Execute(buf, ignitionParams); err != nil {
 		return "", err
 	}
 	return buf.String(), nil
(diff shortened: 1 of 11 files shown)


swagger.yaml
```diff
@@ -1120,7 +1120,7 @@ definitions:
       machine_network_cidr:
         type: string
         description: A CIDR that all hosts belonging to the cluster should have an interfaces with IP address that belongs to this CIDR. The api_vip belongs to this CIDR.
-        pattern: '^([0-9]{1,3}\.){3}[0-9]{1,3}\/[0-9]|[1-2][0-9]|3[0-2]?$'
+        pattern: '^([0-9]{1,3}\.){3}[0-9]{1,3}\/[0-9]|[1-2][0-9]|3[0-2]?$.'
```
avishayt, line 1123: Why?

filanov: So this was the issue! will fix.

filanov:
The downside of this pull request is that we will put the cluster into error on every operation that will fail in the async part. 
We are doing it because we have no way to let the user know what happened, status_info is not an option because some other operation can overwrite it.

internal/bminventory/inventory.go
```diff
@@ -483,7 +483,7 @@ type clusterInstaller struct {
 	params installer.InstallClusterParams
 }
 
-func (c *clusterInstaller) verifyClusterNetworkConfig(cluster *common.Cluster, tx *gorm.DB) error {
+func (b *bareMetalInventory) verifyClusterNetworkConfig(ctx context.Context, cluster *common.Cluster) error {
```
ori-amizur, line 486: The verifyClusterNetworkConfig assumes that logging is done once at the InstallCluster function once.  When the method was taken outside, then errors will not be logged in this case.
My suggestion is to have a single validate function that calls to all other validate functions.  The logging will be done once in the InstallCluster function

filanov: Will be handled with https://issues.redhat.com/browse/MGMT-1375 
I added this comment to the ticket.

PR #313, opened 2020-06-30
Avishay/events2

avishayt:


internal/host/host_test.go
```diff
@@ -141,15 +148,19 @@ var _ = Describe("update_progress", func() {
...
-		ctx   = context.Background()
-		db    *gorm.DB
-		state API
-		host  models.Host
+		ctx        = context.Background()
+		db         *gorm.DB
+		state      API
+		host       models.Host
+		ctrl       *gomock.Controller
+		mockEvents *events.MockHandler
```
filanov, line 156: You don't need to create a mock if it's not used.

internal/host/host_test.go
```diff
@@ -203,15 +214,19 @@ var _ = Describe("monitor_disconnection", func() {
...
-		ctx   = context.Background()
-		db    *gorm.DB
-		state API
-		h     models.Host
+		ctx        = context.Background()
+		db         *gorm.DB
+		state      API
+		h          models.Host
+		ctrl       *gomock.Controller
+		mockEvents *events.MockHandler
```
filanov, line 222: You don't need to create a mock if it's not used.

PR #315, opened 2020-06-30
Igal/python client

tsorya:
look on first commit
second one is auto generated python client.


Makefile
```diff
@@ -60,6 +60,19 @@ update-expirer: build
 	GIT_REVISION=${GIT_REVISION} docker build --build-arg GIT_REVISION -f Dockerfile.s3-object-expirer . -t $(OBJEXP)
 	docker push $(OBJEXP)
 
+create_python_client:
+	mkdir -p build
+	cp swagger.yaml build/
+	echo '{"packageName" : "bm_inventory_client", "packageVersion": "$(shell git rev-parse HEAD || echo none)"}' > build/code-gen-config.json
+	sed -i '/pattern:/d' $(PWD)/build/swagger.yaml
```
tsorya, line 67: there is a problem with creating python client with patter, so i just remove it and validation will be on servers side (verified on test infra)

Makefile
```diff
@@ -60,6 +60,19 @@ update-expirer: build
 	GIT_REVISION=${GIT_REVISION} docker build --build-arg GIT_REVISION -f Dockerfile.s3-object-expirer . -t $(OBJEXP)
 	docker push $(OBJEXP)
 
+create_python_client:
```
YuviGold, line 63: Keep this Makefile standard and use dashes `create-python-client`

Makefile
```diff
@@ -60,6 +60,19 @@ update-expirer: build
...
+	mkdir -p build
```
YuviGold, line 64: Use a variable for the build directory and use it throughout the code. In case the directory is temporary use `$(mktemp -d)` and then you don't need to remove stuff.

Makefile
```diff
@@ -60,6 +60,19 @@ update-expirer: build
...
+	cp swagger.yaml build/
+	echo '{"packageName" : "bm_inventory_client", "packageVersion": "$(shell git rev-parse HEAD || echo none)"}' > build/code-gen-config.json
+	sed -i '/pattern:/d' $(PWD)/build/swagger.yaml
+	docker run -it --rm -u $(shell id -u $(USER)) -v $(PWD)/build:/swagger-api/out -v $(PWD)/build/swagger.yaml:/swagger.yaml:ro,Z -v $(PWD)/build/code-gen-config.json:/config.json:ro,Z jimschubert/swagger-codegen-cli:2.3.1 generate --lang python --config /config.json --output ./bm-inventory-client/ --input-spec /swagger.yaml
```
YuviGold, line 68: This line is too long. Break this line to multiple lines.

Makefile
```diff
@@ -60,6 +60,19 @@ update-expirer: build
...
+	rm -f $(PWD)/build/swagger.yaml
+
+build/bm-inventory-client-%.tar.gz:
+	rm -rf $@
+	cd build/bm-inventory-client/ && python3 setup.py sdist --dist-dir $(PWD)/build/
```
YuviGold, line 73: Build python wheel as well and copy them both. Don't treat the package as only a tar. (prerequisites of **wheel** python package)
```bash
python3 setup.py sdist bdist_wheel
```

Makefile
tsorya: next time

Makefile
tsorya: next time, not need it now

ronniel1: Why do we need the wheel?

YuviGold: It's just better.
https://pythonwheels.com/
https://www.python.org/dev/peps/pep-0427/

Makefile
YuviGold: @tsorya I think it is important when I see multiple different calls from different targets which may cause a confusion (Is it a dir? a file? is it relative?)... Variable would make it fixed and organized.
 `build` , `build/` , `$(PWD)/build` ..

filanov: why next time? there is a target for it already.

Makefile
filanov: please add a comment.

filanov:
Why do you have some many files added?

Makefile
filanov: agree with @YuviGold

PR #316, opened 2020-06-30
Add annotation for authentication type

avishayt:
Not for merging, just to describe which APIs use which authentication type

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -22,6 +22,7 @@ paths:
     post:
       tags:
         - installer
+      x-auth-type: user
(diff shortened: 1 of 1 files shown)


eranco74:
Note this is what we are going to get in the headers:
https://github.com/openshift-metal3/facet/blob/938d7adf42c74f216c34a13f2f3de209a9db1c15/pkg/integration/pull_secret.go#L67

rollandf:
@avishayt Why did you closed this one?

avishayt:
Re-opened.  This will never merge, we can add the per-API methods once the main one merges.

PR #317, opened 2020-06-30
deploy: handle route53 secret creation

danielerez:
* Added 'deploy_route53.py' to tools:
- for creating route53 secret using 'route53-secret.yaml' template,
with secret value specified on ROUTE53_SECRET env var.

* Added deploy-route53 rule to Makefile - invoked from deploy-all


danielerez:
@ori-amizur @filanov @romfreiman

PR #318, opened 2020-06-30
NOT FOR MERGE: Avishay/swagger auth type2

avishayt:
How does this look for specifying authentication?  Need to see if we can disable it...

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -16,12 +16,23 @@ consumes:
 produces:
   - application/json
 
+securityDefinitions:
+  user:
+    type: apiKey
+    in: header
+    name: X-Auth-Token
+  agent:
+    type: apiKey
+    in: header
+    name: X-Auth-Secret
 
 paths:
   /clusters:
     post:
       tags:
         - installer
+      security:
+        - user: []
       summary: Creates a new OpenShift bare metal cluster definition.
       operationId: RegisterCluster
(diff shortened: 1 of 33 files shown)


swagger.yaml
```diff
@@ -16,12 +16,23 @@ consumes:
 produces:
   - application/json
 
+securityDefinitions:
```
rollandf, line 19: I was looking at https://swagger.io/docs/specification/authentication/bearer-authentication/

There it is configured like that:
```
components:
  securitySchemes:
    bearerAuth:            # arbitrary name for the security scheme
      type: http
      scheme: bearer
      bearerFormat: JWT 
```

rollandf: Seems I was not looking at 2.0 ..

swagger.yaml
```diff
@@ -224,6 +247,8 @@ paths:
     get:
       tags:
         - installer
+      security:
+        - user: []
```
avishayt, line 251: Also agent

swagger.yaml
```diff
@@ -47,6 +58,8 @@ paths:
...
```
filanov, line 62: Need to add agent here as well, according to https://github.com/eranco74/assisted-installer/blob/master/src/inventory_client/inventory_client.go#L76

PR #322, opened 2020-07-01
Igal/python client 2

tsorya:


filanov:
Are all those files need to be part of the repo? if we need it just for the container shouldn't it be in git ignore?

tsorya:
@filanov that's the problem. Right now they should. cause test-infra will install client from bm-inventory repo.
When we will make multistage working we will delete it and move to multistage.

tsorya:
for multistage we will add client as part of bm-inventory image and use it in building test-infra.

filanov:
So why not make it an egg or something like this?

tsorya:
to say the truth dunno how to make it work with egg. As i said we hope it will be temporary till we find a way to work with multistage in prow.
it was agreed with @ronniel1 to make it this way

tsorya:
@filanov but if u know how to make it work with egg will be great to here

filanov:
> @filanov but if u know how to make it work with egg will be great to hear

Sadly i don't

YuviGold:
I am against adding 168 new files to the repo just for a patch.. Why the 1st PR was closed?

tsorya:
@YuviGold 1st pr will not work in this case. So it is not relevant for now

tsorya:
relevant change in test-infra https://github.com/openshift/assisted-test-infra/pull/12

filanov:
@YuviGold do you know how to make it a single file?

PR #323, opened 2020-07-01
api: managed domains

danielerez:
Added listing for managed domains (base dns domains).
Endpoint: /api/assisted-install/v1/domains'

This is an API for getting the list of default base DNS domains defined in
'bm-inventory-config' configmap - 'BASE_DNS_DOMAINS' map.

The response should be in the following format:
[{"domain":"example.com","provider":"route53"}]


danielerez:
@avishayt @romfreiman @filanov @jtomasek

internal/domains/managed_domains.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Handler represents managed domains handler
+type Handler struct {
+	baseDNSDomains map[string]string
+}
+
+func (h *Handler) parseDomainProvider(val string) (string, error) {
+	re := regexp.MustCompile("/")
+	if !re.MatchString(val) {
+		return "", fmt.Errorf("Invalid format: %s", val)
```
filanov, line 30: please use `"github.com/pkg/errors"`

danielerez: done

PR #326, opened 2020-07-02
Fix github action due to change in skipper

ronniel1:


filanov:
will it break `skipper make update`?

PR #329, opened 2020-07-02
MGMT-1355 service metrics: cluster creation & installation started

gamli75:
@ronniel1 @avishayt @filanov please review


internal/metrics/common.go
```diff
@@ -0,0 +1,66 @@
...
+	HandlerIDLabel string
+	// StatusCodeLabel is the name that will be set to the status code label, by default is `code`.
+	StatusCodeLabel string
+	// MethodLabel is the name that will be set to the method label, by default is `method`.
+	MethodLabel string
+	// ServiceLabel is the name that will be set to the service label, by default is `service`.
+	ServiceLabel string
+	// IDLabel is the name that will be set to the ID label. ny default is 'id'.
+	IDLabel string
+}
```
ronniel1, line 33: All the labels are not common.
They are specific for the http metrics recorder, and you will have different labels for your metrics

internal/metrics/metricsManager.go
```diff
@@ -0,0 +1,78 @@
...
+	m.incClusterInstallationStarted(clusterVersion)
+}
+
+func (m *MetricsManager) incClusterCreation(clusterVersion string) {
+	m.serviceLogicClusterCreation.WithLabelValues(clusterVersion).Inc()
+}
+
+func (m *MetricsManager) incClusterInstallationStarted(clusterVersion string) {
+	m.serviceLogicClusterInstallationStarted.WithLabelValues(clusterVersion).Inc()
+}
```
ronniel1, line 78: Why do we need two methods for each API?

cmd/main.go
```diff
@@ -135,7 +135,9 @@ func main() {
 	}
 
 	jobApi := job.New(log.WithField("pkg", "k8s-job-wrapper"), kclient, Options.JobConfig)
```
ronniel1, line 137: Shouldn't the common config be initialized here somewhere and be passed to both the metrics manager and to the http metrics reporter?

internal/metrics/common.go
gamli75: I'm not sure what should be the common labels - Log, Registry, ServiceLabel - what else?

internal/metrics/metricsManager.go
gamli75: fixed

internal/metrics/common.go
ronniel1: I think that nothing else

gamli75: fixed

cmd/main.go
gamli75: fixed

internal/metrics/metricsManager.go
```diff
@@ -0,0 +1,79 @@
...
+	InstallationStarted(clusterVersion string)
+}
+
+type ServiceConfig struct {
+	Log logrus.FieldLogger
+	// Registry is the registry that will be used by the recorder to store the metrics,
+	// if the default registry is not used then it will use the default one.
+	Registry       prometheus.Registerer
+	Namespace      string
+	ClusterVersion string
```
ronniel1, line 31: I think that this will need to come from the cluster itself.
We might have multiple clusters, each with each own version

internal/metrics/metricsManager.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+type MetricsManager struct {
+	config ServiceConfig
```
ronniel1, line 35: Code is not formatted

internal/metrics/metricsManager.go
```diff
@@ -0,0 +1,79 @@
...
+	CounterDescriptionClusterCreation            = "Number of cluster resources created, by version"
+	CounterDescriptionClusterInstallationStarted = "Number of clusters that entered installing state, by version"
+)
+
+type API interface {
+	RegisterCluster(clusterVersion string)
+	InstallationStarted(clusterVersion string)
+}
+
+type ServiceConfig struct {
```
ronniel1, line 25: Seems that this is internal to the package, so maybe it should not start with a Capital letter.
Also Why do we need this struct? all the fields can be placed directly in the MetricsManager struct, no?

Usuall we create a Config structure if it contains optional config values, and then it is passed to the constructor. But here there are no optional fields, and anyway, users do not pass the Config struct to the constructor so this seems redundant

internal/metrics/metricsManager.go
```diff
@@ -0,0 +1,79 @@
...
+	CounterClusterInstallationStarted = "assisted_installer_cluster_installation_started"
+)
+
+const (
+	CounterDescriptionClusterCreation            = "Number of cluster resources created, by version"
+	CounterDescriptionClusterInstallationStarted = "Number of clusters that entered installing state, by version"
+)
+
+type API interface {
+	RegisterCluster(clusterVersion string)
```
ronniel1, line 21: Like "InstallationStarted" is in the past , I think this should be "ClusterRegistered"

internal/metrics/metricsManager.go
```diff
@@ -0,0 +1,79 @@
...
+	InstallationStarted(clusterVersion string)
+}
+
+type ServiceConfig struct {
+	Log logrus.FieldLogger
+	// Registry is the registry that will be used by the recorder to store the metrics,
+	// if the default registry is not used then it will use the default one.
```
ronniel1, line 28: In this code registry is mandatory, the code does not use the default

internal/bminventory/inventory.go
```diff
@@ -318,6 +323,7 @@ func (b *bareMetalInventory) RegisterCluster(ctx context.Context, params install
 			WithPayload(common.GenerateError(http.StatusInternalServerError, err))
 	}
 
+	b.metricApi.RegisterCluster(swag.StringValue(params.NewClusterParams.OpenshiftVersion))
```
ronniel1, line 326: If cluster was already registered, will we get to this point?

internal/bminventory/inventory.go
```diff
@@ -643,6 +649,7 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
 	if err = b.db.Preload("Hosts").First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
 		return common.GenerateErrorResponder(err)
 	}
+	b.metricApi.InstallationStarted(cluster.OpenshiftVersion)
```
ronniel1, line 652: We will not get here if the install failed on validations.
@romfreiman , @avishayt in the "assisted_installer_cluster_installation_started:"metric  do we want to count every installation attempt?

romfreiman:
of course

internal/bminventory/inventory.go
avishayt: What do you mean that the cluster was already registered?  Are you thinking of a host?

internal/bminventory/inventory.go
avishayt: Only if the installation actually started.  If the user failed validations it's interesting but not in this metric - we might see in the metric for the installation API that lots of times it returns 400 and that might hint at a UX issue - but here we want to know how many clusters really started installation vs. how many completed with success/error.

internal/metrics/metricsManager.go
avishayt: Yes each cluster would have its own version

gamli75: my mistake - it's the openshift version label and not cluster version - I changed it

internal/metrics/metricsManager.go
gamli75: fixed

internal/bminventory/inventory.go
ronniel1: You are right

internal/metrics/metricsManager.go
gamli75: fixed

gamli75: fixed

internal/metrics/metricsManager.go
gamli75: fixed

internal/metrics/metricsManager.go
gamli75: fixed

gamli75:
@ronniel1 I fixed your remarks - please review again

PR #330, opened 2020-07-02
MGMT-1214 - allow to change hostnames of nodes prior to install comma…

yevgeny-shnaidman:
…nd. the change will take place after RCHOS reboot

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -691,7 +691,7 @@ func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, c
 	return b.clusterApi.SetGeneratorVersion(&cluster, b.Config.KubeconfigGenerator, tx)
 }
 
-func (b *bareMetalInventory) refreshClusterHosts(ctx context.Context, cluster *common.Cluster, tx *gorm.DB, log logrus.FieldLogger) middleware.Responder {
+func (b *bareMetalInventory) refreshClusterHosts(ctx context.Context, cluster *common.Cluster, tx *gorm.DB, log logrus.FieldLogger) error {
 	for _, h := range cluster.Hosts {
 		var host models.Host
 		var err error
@@ -699,12 +699,11 @@ func (b *bareMetalInventory) refreshClusterHosts(ctx context.Context, cluster *c
 			h.ID.String(), cluster.ID.String()).Error; err != nil {
 			log.WithError(err).Errorf("failed to find host <%s> in cluster <%s>",
 				h.ID.String(), cluster.ID.String())
-			return installer.NewUpdateClusterNotFound().WithPayload(common.GenerateError(http.StatusNotFound, err))
+			return common.NewApiError(http.StatusNotFound, err)
 		}
 		if _, err = b.hostApi.RefreshStatus(ctx, &host, tx); err != nil {
 			log.WithError(err).Errorf("failed to refresh state of host %s cluster %s", *h.ID, cluster.ID.String())
-			return installer.NewInstallClusterInternalServerError().
-				WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+			return common.NewApiError(http.StatusInternalServerError, err)
 		}
 	}
 	return nil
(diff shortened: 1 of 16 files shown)


internal/bminventory/inventory.go
```diff
@@ -815,56 +849,74 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
 	if dbReply.Error != nil {
 		log.WithError(dbReply.Error).Errorf("failed to update cluster: %s", params.ClusterID)
 		return installer.NewUpdateClusterInternalServerError().
 			WithPayload(common.GenerateError(http.StatusInternalServerError, dbReply.Error))
 	}
 
+	return nil
+}
+
+func (b *bareMetalInventory) updateHostsData(ctx context.Context, params installer.UpdateClusterParams, db *gorm.DB, log logrus.FieldLogger) middleware.Responder {
```
filanov, line 862: It's strange the internal function return middleware.Responder
please return an error.
The error can be `common.GenerateErrorResponder(err)` but it should be an error.

internal/bminventory/inventory.go
```diff
@@ -761,6 +761,39 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
 		return updateClusterConflict
 	}
 
+	response := b.updateClusterData(ctx, &cluster, params, tx, log)
```
filanov, line 764: if you pass the context you don't need to pass the log. 
You can create a new log from the context.

internal/bminventory/inventory.go
```diff
@@ -815,56 +849,74 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
+}
+
+func (b *bareMetalInventory) updateHostsData(ctx context.Context, params installer.UpdateClusterParams, db *gorm.DB, log logrus.FieldLogger) middleware.Responder {
 	for i := range params.ClusterUpdateParams.HostsRoles {
 		log.Infof("Update host %s to role: %s", params.ClusterUpdateParams.HostsRoles[i].ID,
 			params.ClusterUpdateParams.HostsRoles[i].Role)
 		var host models.Host
-		if err = tx.First(&host, "id = ? and cluster_id = ?",
-			params.ClusterUpdateParams.HostsRoles[i].ID, params.ClusterID).Error; err != nil {
+		err := db.First(&host, "id = ? and cluster_id = ?",
```
filanov, line 867: you can pass the cluster as a parameter then you will not need to get the hosts from the db again.

yevgeny-shnaidman: i have refactored the code due to the gocycle error ( the function UpdateCluster has to many if/for in it, so i tried to keep the exact flow. since it UpdateCluster the load the hosts again, i have done the same here

internal/bminventory/inventory.go
yevgeny-shnaidman: it does not make sense to generate log each time we pass the context. Since we pass the context, it is the same log, so we can use it as is. This is also the way it is done in other functions

internal/bminventory/inventory.go
yevgeny-shnaidman: fixed

internal/bminventory/inventory.go
```diff
@@ -761,6 +760,39 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
 		return updateClusterConflict
 	}
 
+	err = b.updateClusterData(ctx, &cluster, params, tx, log)
+	if err != nil {
+		return common.GenerateErrorResponder(err)
```
filanov, line 765: You should probably log the errors in the API function and not in the internal functions.
This way you will not have multiple logs with the same error.

internal/bminventory/inventory.go
```diff
@@ -761,6 +760,39 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
+	if err := tx.Commit().Error; err != nil {
+		log.Error(err)
+		return installer.NewUpdateClusterInternalServerError().
+			WithPayload(common.GenerateError(http.StatusInternalServerError, errors.New("DB error, failed to commit")))
+	}
+	txSuccess = true
+
+	if err := b.db.Preload("Hosts").First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
+		log.WithError(err).Errorf("failed to get cluster %s after update", params.ClusterID)
+		return installer.NewUpdateClusterInternalServerError().
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/bminventory/inventory.go
```diff
@@ -761,6 +760,39 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
...
+	}
+
+	err = b.updateHostsAndClusterStatus(ctx, &cluster, tx, log)
+	if err != nil {
+		return common.GenerateErrorResponder(err)
+	}
+
+	if err := tx.Commit().Error; err != nil {
+		log.Error(err)
+		return installer.NewUpdateClusterInternalServerError().
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/bminventory/inventory.go
yevgeny-shnaidman: i prefer to leave it like this for now, since this is just the refactoring, i would like to keep the logs as Ori wrote them. I will open a ticket to fix them up at a latter stage

swagger.yaml
```diff
@@ -1195,6 +1197,19 @@ definitions:
             role:
               type: string
               enum: ['master', 'worker']
+      hosts_names:
```
ronniel1, line 1200: Where is this used?

yevgeny-shnaidman: requested_host by default is empty. In case user sets t ( in cluster update) it gets the required value, otherwise we are looking at the value in the inventory. This is what we talked about with the UI guys. In addition, i don't want want to set requested_hosts to the value of the inventory, because then i will configure ignition to use this value, which may contradict what the user is planning for the host ( maybe he wants to use DHCP after rchios reboot)

PR #333, opened 2020-07-02
Adding python client

tsorya:


Makefile
```diff
@@ -1,5 +1,6 @@
 PWD = $(shell pwd)
 UID = $(shell id -u)
+BUILD_FOLDER = $(PWD)/build
```
ronniel1, line 3: We have many targets that reference the build directory.
If you've added a variable for it, shouldn't also change the other references?

Makefile
```diff
@@ -27,7 +28,7 @@ lint:
 	golangci-lint run -v
 
 .PHONY: build
-build: create-build-dir lint unit-test
+build: create-build-dir lint unit-test create-python-client
```
ronniel1, line 31: I don't think 'build' should be dependent on create-python-client.

'update' target should

filanov: +1

PR #334, opened 2020-07-02
chore: update bm-inventory to multi-stage build

djzager:
- Update Dockerfile.bm-inventory to build the binary first and then copy
  into the final image.
- Remove pre-reqs from the build recipe in Makefile.
- Make linter fail on error.
- Add unit-test job to unit-test github action (instead of calling `make
  build`)

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -14,30 +14,18 @@ jobs:
     steps:
     - name: Check out code
       uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


filanov:
Why did you remove lint and unit tests?

filanov:
Build lint and unit-tests are the first feedback that the developer should get on the his code before it goes to review. 
Maybe instead of removing them they should become part of make all?

PR #337, opened 2020-07-05
MGMT 1316 move install host to use new state machine model

filanov:


internal/host/transition.go
```diff
@@ -143,3 +143,39 @@ func (th *transitionHandler) PostResetHost(sw stateswitch.StateSwitch, args stat
...
+
+func (th *transitionHandler) IsValidRoleForInstallation(sw stateswitch.StateSwitch, _ stateswitch.TransitionArgs) (bool, error) {
+	sHost, ok := sw.(*stateHost)
+	if !ok {
+		return false, errors.New("IsValidRoleForInstallation incompatible type of StateSwitch")
+	}
+	validRoles := []string{models.HostRoleMaster, models.HostRoleWorker}
+	if !funk.ContainsString(validRoles, sHost.host.Role) {
+		return false, common.NewApiError(http.StatusConflict,
+			errors.Errorf("Can't install host %s doe invalid host role %s should be one of %s",
```
eranco74, line 164: typo
```suggestion
			errors.Errorf("Can't install host %s due to invalid host role: %s, should be one of %s",
```

internal/host/transition.go
```diff
@@ -143,3 +143,39 @@ func (th *transitionHandler) PostResetHost(sw stateswitch.StateSwitch, args stat
...
+
+func (th *transitionHandler) IsValidRoleForInstallation(sw stateswitch.StateSwitch, _ stateswitch.TransitionArgs) (bool, error) {
+	sHost, ok := sw.(*stateHost)
+	if !ok {
+		return false, errors.New("IsValidRoleForInstallation incompatible type of StateSwitch")
+	}
+	validRoles := []string{models.HostRoleMaster, models.HostRoleWorker}
+	if !funk.ContainsString(validRoles, sHost.host.Role) {
+		return false, common.NewApiError(http.StatusConflict,
+			errors.Errorf("Can't install host %s doe to invalid host role: %s, should be one of %s",
```
ori-amizur, line 164: Typo doe

internal/host/transition.go
```diff
@@ -143,3 +143,39 @@ func (th *transitionHandler) PostResetHost(sw stateswitch.StateSwitch, args stat
...
+				sHost.host.ID.String(), sHost.host.Role, validRoles))
+	}
+	return true, nil
```
ori-amizur, line 167: This boolean function always returns true or error.  It is more like validation function.  If it wouldn't return error but false, then it an be used also for the negative case that it is not valid (maybe move to error state)

internal/host/statemachine.go
```diff
@@ -26,33 +27,53 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
...
+		SourceStates:     []stateswitch.State{HostStatusKnown},
+		DestinationState: HostStatusInstalling,
+		PostTransition:   th.PostInstallHost,
+	})
+
+	// Install disabled host will not do anything
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType:   TransitionTypeInstallHost,
+		SourceStates:     []stateswitch.State{HostStatusDisabled},
+		DestinationState: HostStatusDisabled,
```
ori-amizur, line 75: Why do we need this transition?

internal/host/transition.go
filanov: will be fixed in other PR

internal/host/transition.go
filanov: It's an easy way to return specific http error

internal/host/statemachine.go
filanov: Disable is a special case, we don't want to ignore hosts in disabled state while trying to install. so we don't want to change the state and don't want to return an error. 
Any other suggestion to do it? (there is a comment above the transition)

ori-amizur: Will there be an error if no transition rule is fired?

internal/host/transition.go
ori-amizur: Do we assume that it will be always be called from HTTP request?

filanov: Yes

internal/host/statemachine.go
filanov: Yes

PR #340, opened 2020-07-05
MGMT-1284 enhance install progress params

YuviGold:
https://docs.google.com/document/d/1q70d8ZLWNBBUYosWQ5ajjSGUypIxYPKX7GYi256OZ50/edit

On the next separate PR:
* Enforce enum order


swagger.yaml
```diff
@@ -969,6 +969,9 @@ definitions:
         format: date-time
         x-go-custom-tag: gorm:"type:datetime"
         description: The last time that the host status has been updated
+      progress: # #/definitions/host-progress
+        x-go-custom-tag: gorm:"type:text"
```
ronniel1, line 973: This means that we are keeping in the DB the entire progress structure (including the static array) as a json, no?
I don't think this is the right way,
The fact that we now need to send the UI the array, does not mean that we need to keep it in the DB for every host

skipper.yaml
```diff
@@ -7,6 +7,7 @@ containers:
     s3-object-expirer: Dockerfile.s3-object-expirer
 volumes:
     - $HOME/.cache/go-build:/go/pkg/mod
+    - $HOME/.cache/golangci-lint:$HOME/.cache/golangci-lint
```
ronniel1, line 10: This should be in the build container and not mapped to the host.
We want to have a consistent version

swagger.yaml
YuviGold: The DB queries are auto-generated according to the swagger.
In order to get fields as part of the `host` I must insert it as a property.

How do you suggest to solve it?

internal/host/host.go
```diff
@@ -238,28 +232,35 @@ func (m *Manager) ValidateCurrentInventory(host *models.Host, cluster *common.Cl
 	return m.hwValidator.IsSufficient(host, cluster)
 }
 
-func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, progress string) error {
+func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, progress *models.HostInstallProgressParams) error {
```
filanov, line 235: Maybe this functions should move the transition model? it will be clearer.

internal/bminventory/inventory.go
```diff
@@ -1519,13 +1519,21 @@ func (b *bareMetalInventory) UpdateHostInstallProgress(ctx context.Context, para
...
-	if err := b.hostApi.UpdateInstallProgress(ctx, &host, string(params.HostInstallProgressParams)); err != nil {
+	if err := b.hostApi.UpdateInstallProgress(ctx, &host, params.HostProgress); err != nil {
 		log.WithError(err).Errorf("failed to update host %s progress", params.HostID)
 		// host have nothing to do with the error so we just log it
 		return installer.NewUpdateHostInstallProgressOK()
 	}
-	log.Infof("Host %s in cluster %s reached installation step %s", host.ID, host.ClusterID, params.HostInstallProgressParams)
-	msg := fmt.Sprintf("Host %s reached installation step %s", b.hostApi.GetHostname(&host), params.HostInstallProgressParams)
+
+	event := fmt.Sprintf("reached installation stage %s", params.HostProgress.CurrentStage)
```
tsorya, line 1528: why there is a need for event var? just use msg

internal/host/common.go
```diff
@@ -187,3 +191,16 @@ func checkAndUpdateSufficientHost(log logrus.FieldLogger, h *models.Host, db *go
...
 }
+
+func getStagesByRole(role models.HostRole) []models.HostStage {
+	switch role {
+	case models.HostRoleBootstrap:
+		return []models.HostStage{models.HostStageStartingInstallation, models.HostStageStartWaitingForControlPlane,
+			models.HostStageInstalling, models.HostStageWritingImageToDisk, models.HostStageFinishWaitingForControlPlane,
+			models.HostStageRebooting, models.HostStageConfiguring, models.HostStageDone}
+	default:
+		return []models.HostStage{models.HostStageStartingInstallation, models.HostStageInstalling,
```
tsorya, line 202: we don't need worker? it will never get Joined stage. 
I will add one more stage only for worker in a while.

internal/host/host_test.go
```diff
@@ -30,8 +30,9 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var defaultHwInfo = "default hw info"       // invalid hw info used only for tests
-var defaultInventoryS = "default inventory" // invalid inventory info used only for tests
+var defaultHwInfo = "default hw info"                                 // invalid hw info used only for tests
+var defaultInventoryS = "default inventory"                           // invalid inventory info used only for tests
```
filanov, line 34: i think that the `S` is a typo in `defaultInventoryS`

swagger.yaml
filanov: The alternative is to fix the output every time we are doing get cluster, iterate all the hosts and add the relevant information. A bit overkill no?

internal/bminventory/inventory.go
YuviGold: I created an event var so the `log.Info` and the `msg` calls would contain the same event message

internal/host/host_test.go
YuviGold: There's already a function name `defaultInventory`. So I guess it's because of that.

internal/host/common.go
YuviGold: Fixed.

internal/host/host.go
YuviGold: That's a big change and I don't want to change the logic except for the API in this PR.

swagger.yaml
YuviGold: @ronniel1 decided it's enough important to avoid DB changes in the future (because the stages don't really need to be there)
I am pushing a commit with the appropriate changes.

PR #341, opened 2020-07-05
Michaelf/enable disable mgmt 1317

filanov:


internal/host/statemachine.go
```diff
@@ -90,5 +91,32 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
...
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeEnableHost,
+		SourceStates: []stateswitch.State{
+			HostStatusDisabled,
+		},
+		DestinationState: HostStatusDiscovering,
+		PostTransition:   th.PostEnableHost,
+	})
+
+	// Enable disconnected host does not affect the state
```
ronniel1, line 104: If it does not affect the state, Maybe it should not be a transission?

filanov: How do we combine a state when sometimes it have a transition and sometimes we just allow it to run?
We can return an error in this state if it makes more sense?
I did the same trick with install https://github.com/filanov/bm-inventory/blob/master/internal/host/statemachine.go#L47

filanov: I'm open to suggestions.

avishayt: Seems a bit to transition from a state to itself when it doesn't show up that way in the state machine diagram.  It make more sense to return 409.

filanov: Will do for this case.
What about install? there was a requirement to ignore disabled hosts. should it be a transition as i mentioned in the link?

avishayt: Yes disabled->disabled makes sense for the install transition

PR #345, opened 2020-07-06
Raz/mgmt 1231 costum namespace

RazRegev:


romfreiman:
did u try to deploy to psi staging?

eranco74:
1. Rebase, the code base has changed.
2. Consider adding the namespace option to tools/deploy_options.py

tools/deploy_grafana.py
```diff
@@ -59,7 +59,6 @@ def deploy_oauth_reqs():
         with open(src_file, "r") as src:
             with open(dst_file, "w+") as dst:
                 data = src.read()
-                data = data.replace("BASE64_CERT", ca_cert)
```
ronniel1, line 61: Why was this line removed?

tools/deploy_ui.py
```diff
@@ -27,8 +27,9 @@ def main():
         with open(src_file, "r") as src:
             with open(dst_file, "w+") as dst:
                 data = src.read()
+                data = data.replace('REPLACE_NAMESPACE', deploy_options.namespace)
                 data = data.replace("REPLACE_HOSTNAME",
-                                    utils.get_service_host("assisted-installer-ui", deploy_options.target, deploy_options.domain))
+                utils.get_service_host("assisted-installer-ui", deploy_options.target, deploy_options.domain, deploy_options.namespace))
```
YuviGold, line 32: Should be indented

jparrill: Do we (will) have some kind of pep8 validation at some point? should start thinking on that way?

RazRegev:
> Everything looks fine.
> Although I prefer explicit set of a value instead of holding a variable inside of all the _yamls_ and replacing it. It disallows to use a yaml without a script, and I think it's messier and might cause problems.
> e.g.
> 
> ```python
>         data["spec"]["template"]["spec"]["containers"][0]["image"] = image_fqdn
> ```
> 
> What do you think?

I strongly agree. I thought of some more elegant ways to do it (I think this is the [best one](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/)). 
But then I saw the scripts already override some yaml values in a very specific way (__replace__), and I wanted to be aligned with what we already have without refactoring these parts. This is way I did it like that.

YuviGold:
> > Everything looks fine.
> > Although I prefer explicit set of a value instead of holding a variable inside of all the _yamls_ and replacing it. It disallows to use a yaml without a script, and I think it's messier and might cause problems.
> > e.g.
> > ```python
> >         data["spec"]["template"]["spec"]["containers"][0]["image"] = image_fqdn
> > ```
> > 
> > 
> > What do you think?
> 
> I strongly agree. I thought of some more elegant ways to do it (I think this is the [best one](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/)).
> But then I saw that the scripts already override some yaml values in a very specific way (__ replace __), and I wanted to be aligned with what we already have without refactoring these parts. This is why I did it like that.

I gave you one example from `tools/deploy_assisted_installer.py`. All the places used to do "replace" but it doesn't mean it's recommended. I suggest that if you have a better way - Don't just align yourself but make it better :)

tools/deploy_grafana.py
```diff
@@ -67,8 +69,14 @@ def deploy_oauth_reqs():
...
     '''Wrapper for oc/kubectl apply -f'''
     src_file = os.path.join(os.getcwd(), src_file)
-    print("Deploying {}: {}".format(topic ,src_file))
-    utils.apply(src_file)
+    dst_file = os.path.join(os.getcwd(), 'build', os.path.basename(src_file))
+    with open(src_file) as fp:
+        data = fp.read()
+    data = data.replace('REPLACE_NAMESPACE', deploy_options.namespace)
+    with open(dst_file, 'w') as fp:
+        fp.write(data)
```
eranco74, line 77: This code block is repetitive.
Consider adding the namespace option to tools/deploy_options.py and extracting the replace namespace logic to a method

PR #346, opened 2020-07-06
Unite host roles from hardcoded to use swagger models

YuviGold:


swagger.yaml
```diff
@@ -1203,8 +1202,7 @@ definitions:
               type: string
               format: uuid
             role:
-              type: string
-              enum: ['master', 'worker']
+              $ref: '#/definitions/host-role'
```
filanov, line 1205: the user can set only worker or master.

filanov:
You probably should delete the consts `RoleWorker` and `RoleMaster`

swagger.yaml
YuviGold: Fixed.

PR #347, opened 2020-07-06
Raz/1405 stop podman on cancel

RazRegev:
http://10.0.9.195:8080/job/bm-inventory-subsystem-test/419/console


internal/host/instructionmanager.go
```diff
@@ -67,6 +68,7 @@ func NewInstructionManager(log logrus.FieldLogger, db *gorm.DB, hwValidator hard
 			HostStatusDiscovering:  {[]CommandGetter{hwCmd, inventoryCmd, connectivityCmd}, defaultNextInstructionInSec},
 			HostStatusInstalling:   {[]CommandGetter{installCmd}, defaultBackedOffInstructionInSec},
 			HostStatusDisabled:     {[]CommandGetter{}, defaultBackedOffInstructionInSec},
+			HostStatusError:        {[]CommandGetter{stopPodmanCmd}, defaultBackedOffInstructionInSec},
```
filanov, line 71: Shouldn't it be part of the reset?

RazRegev: In reset we stop and clear the container.
At error it just stop the running installation so it wont continue in background.

RazRegev: Will be another PR for reset

filanov:
@eranco74 @ori-amizur @gamli75 do we have things that we run not in a container besides the agent?

eranco74:
> @eranco74 @ori-amizur @gamli75 do we have things that we run not in a container besides the agent?
There are somethings running on the bootstrap:
1. bootkube, approve-scr and progress services
2. openshift temp control plane containers running in crio (not podman)

swagger.yaml
```diff
@@ -1026,6 +1026,7 @@ definitions:
       - install
       - free-network-addresses
       - reset-agent
+      - stop-podman
```
eranco74, line 1029: I think this should be just another execute command, no reason to define a new step-type for every command.
In case you decided to keep the new step type consider renaming it to reset/cancle-installation, no reason to have stop-podman in the swagger.

ronniel1: @ori-amizur

ori-amizur: If we are not expecting to check the result for failure (such in install), or to get formatted output as in inventory, then we can use execute and not create a new one.
Seems we are not doing any of the above - so it is not needed

ori-amizur:
> @eranco74 @ori-amizur @gamli75 do we have things that we run not in a container besides the agent?

No, besides what @eranco74 mantioned.

RazRegev:
> > @eranco74 @ori-amizur @gamli75 do we have things that we run not in a container besides the agent?
> > There are somethings running on the bootstrap:
> 
> 1. bootkube, approve-scr and progress services
> 2. openshift temp control plane containers running in crio (not podman)

As we decided, those thing will be handled on reset.

internal/host/stopinstallation_test.go
```diff
@@ -0,0 +1,45 @@
...
+		stopCmd = NewStopInstallationCmd(getTestLog())
+
+		id = strfmt.UUID(uuid.New().String())
+		clusterId = strfmt.UUID(uuid.New().String())
+		host = getTestHost(id, clusterId, HostStatusError)
+		Expect(db.Create(&host).Error).ShouldNot(HaveOccurred())
+	})
+
+	It("get_step", func() {
+		stepReply, stepErr = stopCmd.GetStep(ctx, &host)
```
ori-amizur, line 34: It would be nice to check the parameters - to verify that podman kill --all is ran.  (execute is generic).

PR #348, opened 2020-07-06
MGMT-1312 move host update hw info from state machine model

filanov:
Update HW info don't affect the state so the only validation is the states that allow getting new hw info, including Disconnected, Discovering, Insufficent and Known

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -26,8 +26,6 @@ import (
 //go:generate mockgen -source=host.go -package=host -aux_files=github.com/filanov/bm-inventory/internal/host=instructionmanager.go -destination=mock_host_api.go
 
 type StateAPI interface {
-	// Set a new HW information
-	UpdateHwInfo(ctx context.Context, h *models.Host, hwInfo string) (*UpdateReply, error)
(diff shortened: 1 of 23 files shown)


internal/host/host.go
```diff
@@ -180,12 +180,17 @@ func (m *Manager) HandleInstallationFailure(ctx context.Context, h *models.Host)
 	})
 }
 
-func (m *Manager) UpdateHwInfo(ctx context.Context, h *models.Host, hwInfo string) (*UpdateReply, error) {
-	state, err := m.getCurrentState(swag.StringValue(h.Status))
-	if err != nil {
-		return nil, err
+func (m *Manager) UpdateHwInfo(ctx context.Context, h *models.Host, hwInfo string) error {
+	hostStatus := swag.StringValue(h.Status)
```
ori-amizur, line 184: Shouldn't the allowed transitions should be controlled by the state machine?  Otherwise it is a duplication

filanov: it's not affecting the state.
We can do it with a hack that will not change the state in case we have this API

PR #354, opened 2020-07-06
MGMT-717: Do not generate image if a valid one already exists

avishayt:
Together with: https://github.com/oshercc/coreos_installation_iso/pull/9

1. Modify the S3 expirer script to look at the timestamp in the object's
   tag rather than it's real timestamp. This is because there's no real
   supported way to update an object's timestamp without overwriting it.
2. If the current image exists, has the same parameters, and was created
   by the same version that's currently running - update the timestamp
   in the tag and return.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -423,6 +423,23 @@ func (b *bareMetalInventory) GenerateClusterISO(ctx context.Context, params inst
 		return installer.NewGenerateClusterISOConflict()
 	}
 
+	/* If the request has the same parameters as the previous request and the image is still in S3,
+	just refresh the timestamp.
+	*/
+	var imageExists bool
+	if cluster.ImageInfo.ProxyURL == params.ImageCreateParams.ProxyURL &&
+		cluster.ImageInfo.SSHPublicKey == params.ImageCreateParams.SSHPublicKey &&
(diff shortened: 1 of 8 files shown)


internal/bminventory/inventory.go
```diff
@@ -423,6 +423,23 @@ func (b *bareMetalInventory) GenerateClusterISO(ctx context.Context, params inst
...
+		imgName := getImageName(params.ClusterID)
+		imageExists, err = b.s3Client.UpdateObjectTag(ctx, imgName, b.S3Bucket, "create_sec_since_epoch", strconv.FormatInt(now.Unix(), 10))
+		if err != nil {
+			log.WithError(tx.Error).Errorf("failed to contact storage backend")
+			return installer.NewInstallClusterInternalServerError().
+				WithPayload(common.GenerateError(http.StatusInternalServerError, errors.New("failed to contact storage backend")))
+		}
+	}
+
 	updates := map[string]interface{}{}
```
ronniel1, line 443: If image exist than why do we need to update all these fields?

avishayt: We need to update image_created_at, so I figured might as well update the rest rather than putting another "if".

ronniel1: ok

PR #356, opened 2020-07-07
Move host UpdateInventory out of state machine model

oshercc:


internal/host/host.go
```diff
@@ -187,12 +187,17 @@ func (m *Manager) UpdateHwInfo(ctx context.Context, h *models.Host, hwInfo strin
...
-func (m *Manager) UpdateInventory(ctx context.Context, h *models.Host, inventory string) (*UpdateReply, error) {
-	state, err := m.getCurrentState(swag.StringValue(h.Status))
-	if err != nil {
-		return nil, err
+func (m *Manager) UpdateInventory(ctx context.Context, h *models.Host, inventory string) error {
+	hostStatus := swag.StringValue(h.Status)
+	allowedStatuses := []string{HostStatusDiscovering, HostStatusKnown, HostStatusDisconnected,
+		HostStatusInsufficient}
+	if !funk.ContainsString(allowedStatuses, hostStatus) {
+		return common.NewApiError(http.StatusBadRequest,
```
filanov, line 195: the error should be `http.StatusConflict`

internal/host/host.go
```diff
@@ -187,12 +187,17 @@ func (m *Manager) UpdateHwInfo(ctx context.Context, h *models.Host, hwInfo strin
 	return m.db.Model(h).Update("hardware_info", hwInfo).Error
 }
 
-func (m *Manager) UpdateInventory(ctx context.Context, h *models.Host, inventory string) (*UpdateReply, error) {
-	state, err := m.getCurrentState(swag.StringValue(h.Status))
-	if err != nil {
-		return nil, err
+func (m *Manager) UpdateInventory(ctx context.Context, h *models.Host, inventory string) error {
+	hostStatus := swag.StringValue(h.Status)
+	allowedStatuses := []string{HostStatusDiscovering, HostStatusKnown, HostStatusDisconnected,
```
filanov, line 192: we need to start using `models.HostStatusDisconnected` and the the consts that we defined in this package.

PR #357, opened 2020-07-07
Change Dockerfile image base to alpine

YuviGold:
In order to allow running debug programs above the container. (e.g. sh, ls, pwd, cat, etc.)
The image size stays small with alpine distro.

```
➜  bm-inventory git:(yuval/alpine-image-base) ✗ docker images bm-inventory
REPOSITORY          TAG                 IMAGE ID            CREATED              SIZE
bm-inventory        scratch             dbc833fc4e24        4 seconds ago        57.7MB
bm-inventory        alpine              94a677466058        About a minute ago   63.3MB
```

### DIFF
--- a/Dockerfile.bm-inventory
+++ b/Dockerfile.bm-inventory
@@ -1,6 +1,6 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest as certs
 
-FROM scratch
+FROM alpine
 COPY --from=certs /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/ssl/certs/ca-bundle.crt
 COPY --from=certs /etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt /etc/ssl/certs/ca-bundle.trust.crt
 ARG GIT_REVISION


tsorya:
@YuviGold  alpine images has a problem with dns resolution. Not sure it was fixed already
@romfreiman  ?

Dockerfile.bm-inventory
```diff
@@ -1,6 +1,6 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest as certs
 
-FROM scratch
+FROM alpine
```
danielerez, line 3: If we can go with alpine, then we no longer need to use ubi-minimal, as it should contain the required certificates.

romfreiman:
yes, they usually do

romfreiman:
https://github.com/gliderlabs/docker-alpine/issues/476
seems that it was fixed not so long time ago

romfreiman:
does subsystem pass?

YuviGold:
> does subsystem pass?

Yes.

@danielerez said he prefers to work with alpine, and he can made the necessary changes later

danielerez:
> > does subsystem pass?
> 
> Yes.
> 
> @danielerez said he prefers to work with alpine, and he can made the necessary changes later

I've meant it's simpler as it contains the certificates already. If we don't have other issues with alpine (as discussed in: https://github.com/filanov/bm-inventory/pull/282#discussion_r444301725), I can remove the ubi8 multistage build later from dockerfile and test it.

PR #359, opened 2020-07-07
Add logger to subsystem

filanov:
Use logger in install cluster flows


filanov:
We didn't found an easy way to add logs to bm-inventory when starting test so this is a good solution for now.

PR #360, opened 2020-07-07
Bugfix SetBootstrap field

YuviGold:


filanov:
I was thinking about subsystem test as well. but it's ok for now.

PR #364, opened 2020-07-07
MGMT-1358: validate cluster name on register and update

danielerez:
Added validation for cluster name format in register/update cluster flows.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -310,6 +310,9 @@ func (b *bareMetalInventory) RegisterCluster(ctx context.Context, params install
 		}
 		setPullSecret(&cluster, params.NewClusterParams.PullSecret)
 	}
+	if err := validations.ValidateClusterNameFormat(swag.StringValue(params.NewClusterParams.Name)); err != nil {
+		return common.NewApiError(http.StatusBadRequest, err)
+	}
 
 	err := b.clusterApi.RegisterCluster(ctx, &cluster)
 	if err != nil {
@@ -754,6 +757,11 @@ func (b *bareMetalInventory) UpdateCluster(ctx context.Context, params installer
 			log.WithError(err).Errorf("Pull-secret for cluster %s, has invalid format", params.ClusterID)
 			return installer.NewUpdateClusterBadRequest().
 				WithPayload(common.GenerateError(http.StatusBadRequest, errors.New("Pull-secret has invalid format")))
+		}
+	}
+	if newClusterName := swag.StringValue(params.ClusterUpdateParams.Name); newClusterName != "" {
+		if err = validations.ValidateClusterNameFormat(newClusterName); err != nil {
+			return common.NewApiError(http.StatusBadRequest, err)
 		}
 	}
 
--- a/internal/cluster/validations/validation_test.go
+++ b/internal/cluster/validations/validation_test.go
@@ -132,6 +132,29 @@ var _ = Describe("Base DNS validation", func() {
 	})
 })
 
+var _ = Describe("Cluster name validation", func() {
+	It("success", func() {
+		err := ValidateClusterNameFormat("test-1")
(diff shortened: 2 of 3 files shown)


danielerez:
@filanov @romfreiman

internal/cluster/validations/validations.go
```diff
@@ -153,3 +157,17 @@ func checkDNSRecordsExistence(names []string, dnsProvider dnsproviders.Provider)
...
 }
+
+// ValidateClusterName validates specified name format and cluster's existance in DB
+func ValidateClusterName(name string, db *gorm.DB) error {
+	if matched, err := regexp.MatchString(clusterNameRegex, name); !matched {
+		return fmt.Errorf("Cluster name is not valid: %v", err)
+	}
+
+	var cluster common.Cluster
+	err := db.First(&cluster, "name = ?", name).Error
```
avishayt, line 168: Why do we need cluster name to be unique?  I assume this is for DNS, but if so, I would validate it there.  Why do we care if two clusters are named "test" if they use different DNS?

internal/bminventory/inventory.go
```diff
@@ -310,6 +310,10 @@ func (b *bareMetalInventory) RegisterCluster(ctx context.Context, params install
 		}
 		setPullSecret(&cluster, params.NewClusterParams.PullSecret)
 	}
+	if err := validations.ValidateClusterName(swag.StringValue(params.NewClusterParams.Name), b.db); err != nil {
+		return installer.NewRegisterClusterBadRequest().
```
filanov, line 314: You can use `common.NewApiError(http.StatusBadRequest, errors.New("Cluster name already exists"))`

internal/bminventory/inventory_test.go
```diff
@@ -783,6 +783,17 @@ var _ = Describe("cluster", func() {
 			Expect(reply).To(BeAssignableToTypeOf(installer.NewUpdateClusterNotFound()))
 		})
 
+		It("Invalid cluster name", func() {
```
filanov, line 786: Usually we mock non local packages, so i think that this test belong to subsystem.

internal/cluster/validations/validation_test.go
```diff
@@ -132,6 +134,39 @@ var _ = Describe("Base DNS validation", func() {
...
+		Expect(err).ShouldNot(HaveOccurred())
+	})
+	It("invalid name", func() {
+		err := ValidateClusterName("test!A", db)
+		Expect(err).Should(HaveOccurred())
+	})
+	It("existing name", func() {
+		err := ValidateClusterName("test", db)
+		Expect(err).Should(HaveOccurred())
+	})
```
filanov, line 160: Probably should add after each and close the DB.

internal/cluster/validations/validations.go
filanov: +1

danielerez: > Why do we need cluster name to be unique? I assume this is for DNS, but if so, I would validate it there. Why do we care if two clusters are named "test" if they use different DNS?

Cluster name uniqueness is actually already validated in the UI, the validation was just missing in the backend. But it's also needed for DNS indeed, in the cluster name is part of the prefix. So don't we want to correlate with the UI validation and validate the cluster name?

danielerez: removed uniqueness check for now, left the format name validation.

internal/bminventory/inventory.go
danielerez: done

internal/bminventory/inventory_test.go
danielerez: removed mock for now.

internal/cluster/validations/validation_test.go
danielerez: removed for now.

PR #365, opened 2020-07-07
MGMT-1424 setBootstrapHost: check if already has bootstrap

RazRegev:
If the cluster was restarted it probably already contain bootstrap host.
In this case we use the same host, instead of setting a new one.
Also added a verification of this check to the reset subsystem test.

Jenkins build:
http://10.0.9.195:8080/job/bm-inventory-subsystem-test/476/console

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -722,6 +722,14 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
 func (b *bareMetalInventory) setBootstrapHost(ctx context.Context, cluster common.Cluster, db *gorm.DB) error {
 	log := logutil.FromContext(ctx, b.log)
 
+	// check if cluster already has bootstrap
+	for _, h := range cluster.Hosts {
+		if h.Bootstrap {
+			log.Infof("Bootstrap ID is %s", h.ID)
+			return nil
+		}
+	}
+
 	masterNodesIds, err := b.clusterApi.GetMasterNodesIds(ctx, &cluster, db)
 	if err != nil {
 		log.WithError(err).Errorf("failed to get cluster %s master node id's", cluster.ID)
--- a/subsystem/cluster_test.go
+++ b/subsystem/cluster_test.go
@@ -854,6 +854,85 @@ var _ = Describe("cluster install", func() {
 				}
 
 			})
+			It("[only_k8s]reset cluster and disable bootstrap", func() {
+				var bootstrapID *strfmt.UUID
+
+				By("verify reset success")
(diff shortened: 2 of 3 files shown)


internal/bminventory/inventory.go
```diff
@@ -659,6 +659,11 @@ func (b *bareMetalInventory) InstallCluster(ctx context.Context, params installe
 func (b *bareMetalInventory) setBootstrapHost(ctx context.Context, cluster common.Cluster, db *gorm.DB) error {
 	log := logutil.FromContext(ctx, b.log)
 
+	if bootstrapId := b.getBootstrapHostId(&cluster); bootstrapId != nil {
```
filanov, line 662: After Reset the user may disable this host. so you probably should cover this case as well.

subsystem/cluster_test.go
```diff
@@ -669,18 +669,31 @@ var _ = Describe("cluster install", func() {
...
+				rep, err := bmclient.Installer.GetCluster(ctx, &installer.GetClusterParams{ClusterID: clusterID})
+				Expect(err).NotTo(HaveOccurred())
+				c := rep.GetPayload()
+				for _, host := range c.Hosts {
+					if host.Role == models.HostRoleBootstrap {
+						originBootstrapID = *host.ID
+						break
+					}
+				}
 				_, err = bmclient.Installer.CancelInstallation(ctx, &installer.CancelInstallationParams{ClusterID: clusterID})
```
filanov, line 685: This test is becoming long, please add steps with `By`

internal/bminventory/inventory.go
eranco74: I guess that reset should unset the bootstrap flag

internal/bminventory/inventory.go
```diff
@@ -721,6 +726,15 @@ func (b *bareMetalInventory) setBootstrapHost(ctx context.Context, cluster commo
 	return nil
 }
 
+func (b *bareMetalInventory) getBootstrapHostId(c *common.Cluster) *strfmt.UUID {
```
filanov, line 729: i think that this functions should be as part of host api, something like `IsBootstrap`

internal/bminventory/inventory.go
```diff
@@ -721,6 +726,15 @@ func (b *bareMetalInventory) setBootstrapHost(ctx context.Context, cluster commo
...
+	for _, h := range c.Hosts {
+		if h.Role == models.HostRoleBootstrap {
```
filanov, line 731: I don't think that you need to check the role. because `SetBootstrap` set a different field
https://github.com/filanov/bm-inventory/blob/master/internal/host/host.go#L294

PR #366, opened 2020-07-07
MGMT-1423 - requested name will always contain the current hostname w…

yevgeny-shnaidman:
…hen the host info is returnd by API

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -2009,6 +2009,15 @@ func (b *bareMetalInventory) GetFreeAddresses(ctx context.Context, params instal
 }
 
 func (b *bareMetalInventory) customizeHost(host *models.Host) error {
+	err := b.customizeHostProgress(host)
+	if err != nil {
+		return err
+	}
+	b.customizeHostname(host)
+	return nil
+}
+
+func (b *bareMetalInventory) customizeHostProgress(host *models.Host) error {
 	var hostProgressReport models.HostProgressReport
 	var err error
 
@@ -2028,4 +2037,8 @@ func (b *bareMetalInventory) customizeHost(host *models.Host) error {
 	}
 
 	return nil
+}
+
+func (b *bareMetalInventory) customizeHostname(host *models.Host) {
+	host.RequestedHostname = b.hostApi.GetHostname(host)
 }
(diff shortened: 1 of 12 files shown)


internal/bminventory/inventory.go
```diff
@@ -958,6 +958,12 @@ func calculateHostNetworks(log logrus.FieldLogger, cluster *common.Cluster) []*m
 	return ret
 }
 
+func (b *bareMetalInventory) updateCurrentHostnames(cluster *common.Cluster) {
```
ronniel1, line 961: please rename to `fillCurrentHostnames`
Update ususally updates the DB

internal/bminventory/inventory_test.go
```diff
@@ -701,6 +701,7 @@ var _ = Describe("cluster", func() {
 			})
 
 			It("GetCluster", func() {
+				mockHostApi.EXPECT().GetHostname(gomock.Any()).Return("whatever").Times(3)
```
ronniel1, line 704: What about testing the reply for the other APIs that return hosts?

internal/common/host_utils.go
```diff
@@ -0,0 +1,19 @@
+package common
```
ronniel1, line 1: I don't think we need a host_utils file.
Just put this file in the hosts.go file

ronniel1: Also where is the UT for this?

internal/bminventory/inventory.go
yevgeny-shnaidman: fixed

internal/common/host_utils.go
yevgeny-shnaidman: it is used from different packages, like host and hardware, so it better be in common. if it is in host, we will have circular dependency

internal/bminventory/inventory_test.go
yevgeny-shnaidman: it was already inside the tests, since it was also called for events and logs

internal/common/host_utils.go
yevgeny-shnaidman: regading unit-test:  to me it seems like a small function without to much logic

ronniel1: the host file is the place for host related stuff. it can be used from anywhere

yevgeny-shnaidman: host.go impelements host manager API, and it import hardware package. In case i define this function in it, hardware package will need to import host package, and compilation will fail

ronniel1: ok

PR #367, opened 2020-07-07
Raz/mgmt 1401 clear s3 files

RazRegev:


avishayt:
I suggest making some const that contains all of the files.  Add a validation to DownloadClusterFiles that param.FileName is one of them, and then also use that const list for deletion.  That way we will know that if a new file is added later it will be deleted.

pkg/s3Client/s3Client.go
```diff
@@ -71,6 +72,20 @@ func (s s3Client) DownloadFileFromS3(ctx context.Context, fileName string, s3Buc
...
 }
 
+func (s s3Client) DeleteFileFromS3(ctx context.Context, fileName string, s3Bucket string) (bool, error) {
+	log := logutil.FromContext(ctx, s.log)
+	log.Infof("Deleting file if %s exists in %s", fileName, s3Bucket)
+	err := s.client.RemoveObject(s3Bucket, fileName)
+	if err != nil {
+		errResponse := minio.ToErrorResponse(err)
+		if errResponse.Code == "NoSuchKey" {
+			return false, nil
```
avishayt, line 82: This is a good check, but don't care about the bool return value.  We can treat NoSuchKey and deleted as success (return nil), and for any other error return the error.

pkg/s3Client/s3Client.go
```diff
@@ -71,6 +72,20 @@ func (s s3Client) DownloadFileFromS3(ctx context.Context, fileName string, s3Buc
 	return resp, contentLength, nil
 }
 
+func (s s3Client) DeleteFileFromS3(ctx context.Context, fileName string, s3Bucket string) (bool, error) {
+	log := logutil.FromContext(ctx, s.log)
+	log.Infof("Deleting file if %s exists in %s", fileName, s3Bucket)
+	err := s.client.RemoveObject(s3Bucket, fileName)
+	if err != nil {
+		errResponse := minio.ToErrorResponse(err)
+		if errResponse.Code == "NoSuchKey" {
```
eranco74, line 81: Add a log

pkg/s3Client/s3Client.go
```diff
@@ -71,6 +72,20 @@ func (s s3Client) DownloadFileFromS3(ctx context.Context, fileName string, s3Buc
 	return resp, contentLength, nil
 }
 
+func (s s3Client) DeleteFileFromS3(ctx context.Context, fileName string, s3Bucket string) (bool, error) {
```
eranco74, line 75: No need for the bool in the return

internal/bminventory/inventory.go
```diff
@@ -1740,6 +1740,10 @@ func (b *bareMetalInventory) ResetCluster(ctx context.Context, params installer.
 		}
 	}
 
+	if err := b.deleteS3ClusterFiles(ctx, &c); err != nil {
+		return installer.NewResetClusterInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
```
filanov, line 1744:
```suggestion
return common.NewApiError(http.StatusInternalServerError, err)
```

pkg/s3Client/s3Client.go
filanov: +1

internal/bminventory/inventory.go
```diff
@@ -1819,6 +1840,15 @@ func (b *bareMetalInventory) ResetCluster(ctx context.Context, params installer.
 	return installer.NewResetClusterAccepted().WithPayload(&c.Cluster)
 }
 
+func (b *bareMetalInventory) deleteS3ClusterFiles(ctx context.Context, c *common.Cluster) error {
+	for _, name := range clusterFileNames {
+		if err := b.s3Client.DeleteFileFromS3(ctx, fmt.Sprintf("%s/%s", c.ID, name), b.S3Bucket); err != nil {
```
filanov, line 1845: what will happen if you try to delete file that no longer exists?

PR #373, opened 2020-07-08
MGMT-1441 add empty implementation to cluster refresh on preparing-for-installation state

filanov:
This is a temp fix - we still need to move cluster api to the new state machine model.


filanov:
subsystem passed.

PR #374, opened 2020-07-08
Eran/MGMT-1413	Add basic identity functionality to bm-inventory

eranco74:


swagger.yaml
```diff
@@ -1325,6 +1325,8 @@ definitions:
       name:
         type: string
         description: Name of the OpenShift cluster.
+      owner:
```
ronniel1, line 1328: @avishayt Should this be in the swagger?
Maybe we also need the org?

eranco74:
Subsystem:
SUCCESS! -- 45 Passed | 0 Failed | 0 Pending | 0 Skipped
System:
http://10.46.10.2:8080/job/test-infra/2029/

pkg/auth/auth.go
```diff
@@ -0,0 +1,30 @@
+package auth
+
+import (
+	"context"
+	"net/http"
+)
+
+type contextKey string
+
+const ContextUserIDKey = contextKey("user_id")
```
ronniel1, line 10: Context keys should usually be private to the module they are defined in
Also these values will all be in a struct in a single context key

internal/bminventory/inventory.go
```diff
@@ -1106,7 +1110,8 @@ func calculateHostNetworks(log logrus.FieldLogger, cluster *common.Cluster) []*m
 func (b *bareMetalInventory) ListClusters(ctx context.Context, params installer.ListClustersParams) middleware.Responder {
```
ronniel1, line 1110: Also need to update GetCluster, no?

eranco74: Technically authorization should come from OCM.
In cluster list specifically we need to filter ourselves.
Also note that this is mainly a place holder and I can't really test it until we have auth.

ronniel1: OCM will just tell us if the user has admin permissions or not.
We will need to filter based on that, in all relevant APIs

eranco74: OK, I'll update, yet I'd like to start with just the list API and add more APIs in another PR

internal/identity/identity.go
```diff
@@ -0,0 +1,21 @@
...
+
+func IsAdmin(ctx context.Context) bool {
+	return auth.UserRoleFromContext(ctx) == auth.AdminUserRole
+}
+
+func GetUserIDFilter(ctx context.Context) string {
+	query := ""
+	if !IsAdmin(ctx) {
+		username := auth.UserIDFromContext(ctx)
+		query = fmt.Sprintf("owner = '%s'", username)
```
tsorya, line 18: no need for org? can't we have same user in different orgs?

internal/identity/identity.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func IsAdmin(ctx context.Context) bool {
+	return auth.UserRoleFromContext(ctx) == auth.AdminUserRole
+}
+
+func GetUserIDFilter(ctx context.Context) string {
+	query := ""
+	if !IsAdmin(ctx) {
+		username := auth.UserIDFromContext(ctx)
```
tsorya, line 17: maybe userId ?

internal/identity/identity.go
eranco74: Unsure, I guess we need "org admin" and "super admin"

eranco74: For now admin will get full access

PR #375, opened 2020-07-08
MGMT-1443 Wait for dummy job to finish pre proceeding

oshercc:


internal/bminventory/inventory.go
```diff
@@ -161,6 +161,10 @@ func generateDummyISOImage(jobApi job.API, b *bareMetalInventory, log logrus.Fie
 	if err := jobApi.Create(context.Background(), b.createImageJob(jobName, imgName, "Dummy")); err != nil {
 		log.WithError(err).Errorf("failed to generate dummy ISO image")
 	}
+	if err := b.job.Monitor(context.Background(), jobName, b.Namespace); err != nil {
```
filanov, line 164: Not sure it will help. because K8S will think that the service is operational and the subsystem will fail because the service will not reply to API

filanov: We have no problem that Generate iso API will take a long time. this part was only a hack to help in the deployment and this will help in system environment but not in the CI.

filanov:
suggesting to add a script that will wait for the job to start running or complete 
```
kubectl get pod  | grep dummyimage | grep 'Running\|Completed'
```

PR #377, opened 2020-07-09
MGMT-1211 Add installing-pending-for-action state & Handle wrong boot…

YuviGold:
… order

Adding a new state installing-pending-user-action with extra status_info explaining the edge-case.
The state will be trigger in case of a combination of <action=registerHost, state=installing-in-progress, progress=rebooting>

[Jenkins issue](https://issues.redhat.com/browse/MGMT-1211)

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -35,16 +35,17 @@ type SpecificHardwareParams interface {
 }
 
 const (
-	HostStatusDiscovering          = "discovering"
-	HostStatusKnown                = "known"
-	HostStatusDisconnected         = "disconnected"
-	HostStatusInsufficient         = "insufficient"
-	HostStatusDisabled             = "disabled"
-	HostStatusInstalling           = "installing"
-	HostStatusInstallingInProgress = "installing-in-progress"
-	HostStatusInstalled            = "installed"
-	HostStatusError                = "error"
-	HostStatusResetting            = "resetting"
+	HostStatusDiscovering                 = "discovering"
+	HostStatusKnown                       = "known"
+	HostStatusDisconnected                = "disconnected"
+	HostStatusInsufficient                = "insufficient"
+	HostStatusDisabled                    = "disabled"
+	HostStatusInstalling                  = "installing"
+	HostStatusInstallingInProgress        = "installing-in-progress"
+	HostStatusInstallingPendingUserAction = "installing-pending-user-action"
+	HostStatusInstalled                   = "installed"
+	HostStatusError                       = "error"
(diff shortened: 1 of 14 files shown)


internal/host/host.go
```diff
@@ -35,16 +35,17 @@ type SpecificHardwareParams interface {
...
-	HostStatusError                = "error"
-	HostStatusResetting            = "resetting"
+	HostStatusDiscovering                = "discovering"
+	HostStatusKnown                      = "known"
+	HostStatusDisconnected               = "disconnected"
+	HostStatusInsufficient               = "insufficient"
+	HostStatusDisabled                   = "disabled"
+	HostStatusInstalling                 = "installing"
+	HostStatusInstallingInProgress       = "installing-in-progress"
+	HostStatusInstallingPendingForAction = "installing-pending-for-action"
```
filanov, line 45: suggesting not to use those consts, and start using states that generated from swagger `models.HostStatus...`

subsystem/cluster_test.go
```diff
@@ -369,7 +369,42 @@ var _ = Describe("cluster install", func() {
 			Expect(err).To(BeNil())
 			host := getHost(clusterID, *hostID)
 			Expect(*host.Status).To(Equal("error"))
+		})
+
+		It("[only_k8s]register host after reboot - wrong boot order", func() {
```
filanov, line 374: suggesting to replace the comments into steps with `By`

internal/host/host.go
YuviGold: I agree. But that would cause a bigger PR changes. I'll do another PR just updating all this consts stuff

subsystem/cluster_test.go
YuviGold: Nice! Done.

swagger.yaml
```diff
@@ -974,6 +974,7 @@ definitions:
           - disabled
           - installing
           - installing-in-progress
+          - installing-pending-for-action
```
avishayt, line 977: Please rename to installing-pending-user-action
IMO this makes it clearer that the user should do something

internal/host/host.go
avishayt, line 45: installing-pending-user-action , see comment below

filanov: i don't think we should do it in one go, you can start small with only the new state.

YuviGold: And then you would suggest I should keep everything as the current standard :)

YuviGold: Fixed.

PR #378, opened 2020-07-09
MGMT-1425 Add subsystem test verifying bootstrap on install

YuviGold:
[Jenknis issue](https://issues.redhat.com/browse/MGMT-1425)


YuviGold:
Added in #381 instead.

PR #379, opened 2020-07-09
MGMT-1118 Update ignition spec version for RHCOS

rollandf:


rollandf:
To merge only when the image will be updated to rhcos in Image Builder

internal/bminventory/inventory.go
```diff
@@ -85,7 +85,7 @@ type Config struct {
 }
 
 const ignitionConfigFormat = `{
-"ignition": { "version": "3.0.0" },
+"ignition": { "version": "2.2.0" },
```
eranco74, line 88: :fearful: 
:crossed_fingers:

internal/bminventory/inventory.go
```diff
@@ -103,6 +103,7 @@ const ignitionConfigFormat = `{
 },
 "storage": {
     "files": [{
+	  "filesystem": "root",
```
eranco74, line 106: Fix indentation

rollandf: Done

PR #381, opened 2020-07-09
MGMT-1454 Separate HostProgressReport field & Fix host stages

YuviGold:
[MGMT-1454 Jira Issue](https://issues.redhat.com/browse/MGMT-1454)
[MGMT-1425 Jira Issue](https://issues.redhat.com/browse/MGMT-1425)


internal/host/host.go
```diff
@@ -47,6 +46,22 @@ const (
...
+}
+var MasterStages = [...]models.HostStage{
+	models.HostStageStartingInstallation, models.HostStageInstalling,
+	models.HostStageWritingImageToDisk, models.HostStageRebooting,
+	models.HostStageConfiguring, models.HostStageJoined, models.HostStageDone,
+}
+var WorkerStages = [...]models.HostStage{
+	models.HostStageStartingInstallation, models.HostStageInstalling,
+	models.HostStageWritingImageToDisk, models.HostStageRebooting,
+	models.HostStageConfiguring, models.HostStageDone,
```
eranco74, line 62: We don't have Joined for workers?
@tsorya ?

internal/host/host.go
```diff
@@ -369,20 +373,16 @@ func (m *Manager) GetHostname(host *models.Host) string {
 	return hostName
 }
 
-func (m *Manager) GetStagesByRole(role models.HostRole) []models.HostStage {
+func (m *Manager) GetStagesByRole(role models.HostRole, isbootstrap bool) []models.HostStage {
+	if isbootstrap || role == models.HostRoleBootstrap {
```
eranco74, line 377: What's the isbootstrap param for?
The host role is bootstrap or not.

internal/bminventory/inventory.go
```diff
@@ -2009,34 +2009,13 @@ func (b *bareMetalInventory) GetFreeAddresses(ctx context.Context, params instal
...
-	progressJson, err := json.Marshal(hostProgressReport)
-	host.Progress = string(progressJson)
-
-	if err != nil {
-		return err
-	}
-
-	return nil
+func (b *bareMetalInventory) customizeHostStages(host *models.Host) {
+	host.Stages = b.hostApi.GetStagesByRole(host.Role, host.Bootstrap)
```
eranco74, line 2018: Is it possible to have a host.Role that isn't bootstrap and host.Bootstrap set to True?

subsystem/cluster_test.go
```diff
@@ -411,79 +411,113 @@ var _ = Describe("cluster install", func() {
...
+				updateProgressWithInfo(*hosts[0].ID, clusterID, installProgress, installInfo)
+				hostFromDB := getHost(clusterID, *hosts[0].ID)
+
+				Expect(*hostFromDB.Status).Should(Equal(host.HostStatusError))
+				Expect(*hostFromDB.StatusInfo).Should(Equal(fmt.Sprintf("%s - %s", installProgress, installInfo)))
+				Expect(hostFromDB.Progress.CurrentStage).Should(Equal(models.HostStageWritingImageToDisk)) // Last stage
+				Expect(hostFromDB.Progress.ProgressInfo).Should(BeEmpty())
+			})
+
+			By("cant_report_after_error", func() {
```
eranco74, line 460: Why can't we report after an error?

internal/bminventory/inventory.go
```diff
@@ -2009,34 +2009,13 @@ func (b *bareMetalInventory) GetFreeAddresses(ctx context.Context, params instal
...
-	if host.Progress != "" {
-		if err = json.Unmarshal([]byte(host.Progress), &hostProgressReport); err != nil {
-			return err
-		}
-	}
-
-	hostProgressReport.Stages = b.hostApi.GetStagesByRole(host.Role)
-
-	progressJson, err := json.Marshal(hostProgressReport)
-	host.Progress = string(progressJson)
```
eranco74, line 2033: So host.Progress is host.Stages now?

swagger.yaml
```diff
@@ -985,9 +985,14 @@ definitions:
         format: date-time
         x-go-custom-tag: gorm:"type:datetime"
         description: The last time that the host status has been updated
-      progress: # #/definitions/host-progress-report
-        x-go-custom-tag: gorm:"type:text"
-        type: string
+      progress:
```
eranco74, line 988: This still exists?

internal/host/host.go
YuviGold: Right now the bootstrap role isn't being really saved but used to initialize the assisted-installer.
After the assisted-installer started the bootstrap routine - it changes to "master" role.
The "isbootstrap" boolean field is the way today we know if a host is a bootstrap or not and not by its role.

internal/bminventory/inventory.go
YuviGold: Yes. As explained in the comment before - bootstrap role and bootstrap boolean are defined in different places.
bootstrap boolean on "InstallCluster" - picking one master and setting it as a bootstrap.
Sometime after that - the "InstallCmd" is being called and checking if bootstrap is set. if so -> assisted-installer gets a flag for a "bootstrap" which is being translated into a config.role - which changes to master immediate after the bootstrap routine is being started.

I needs to be changed. But for now I didn't to break the API anymore.

subsystem/cluster_test.go
YuviGold: @avishayt @romfreiman didn't see a reason to support it.

internal/bminventory/inventory.go
YuviGold: Because the stages aren't being saved in the DB. There's no need to keep the API complex.
The naming of the fields and their content is the same except for hierarchy.
To avoid marshaling issues It was changed from
```yaml
host:
...
     status_info:
       type: string
       x-go-custom-tag: gorm:"type:varchar(2048)"
     progress: # #/definitions/host-progress-report
       x-go-custom-tag: gorm:"type:text"
       type: string

 host-progress-report:
   type: object
   properties:
     stages:
       type: array
       items:
         $ref: '#/definitions/host-stage'
       x-go-custom-tag: gorm:"-"
     current_progress:
       $ref: '#/definitions/host-progress'
```
to:
```yaml
 host:
...
     status_info:
       type: string
       x-go-custom-tag: gorm:"type:varchar(2048)"
     progress:
       $ref: '#/definitions/host-progress'
     stages:
       type: array
       items:
         $ref: '#/definitions/host-stage'
       x-go-custom-tag: gorm:"-"
```

swagger.yaml
YuviGold: Yes. explained in the previous comment.

internal/host/host.go
tsorya: relevant only for 2 masters

internal/host/host.go
eranco74: So this ` role == models.HostRoleBootstrap` is always False?
Why have it?

swagger.yaml
```diff
@@ -985,9 +985,14 @@ definitions:
...
+        $ref: '#/definitions/host-progress'
+        x-go-custom-tag: gorm:"embedded;embedded_prefix:progress_"
+      stages:
```
avishayt, line 991:
```suggestion
      progress_stages:
```

subsystem/cluster_test.go
avishayt: Is anyone going to be reporting progress once the installation failed? (it's not that we technically can't, it just doesn't seem to make sense).

internal/host/host.go
```diff
@@ -47,6 +46,22 @@ const (
 	HostStatusResetting            = "resetting"
 )
 
+var BootstrapStages = [...]models.HostStage{
```
filanov, line 49: What is the difference between `var BootstrapStages = [...]models.HostStage{` and `var BootstrapStages = []models.HostStage{` ?

YuviGold: It is more secured, not allowing to change the array dimensions in runtime.

https://stackoverflow.com/questions/13137463/declare-a-constant-array
```
Note the [...] instead of []: it ensures you get a (fixed size) array instead of a slice. So the values aren't fixed but the size is.
```

filanov: Nice

subsystem/cluster_test.go
YuviGold: Decided to keep it strict to avoid potential bugs

PR #383, opened 2020-07-09
configmap: set base_dns_domains from env var

danielerez:
Set BASE_DNS_DOMAINS in configmap when the env var is defined.


danielerez:
@ori-amizur @filanov @romfreiman

filanov:
How will the code handle empty DNS domain?

danielerez:
> How will the code handle empty DNS domain?

It is passed as an empty string from the Makefile in that case, so it would just keep its default value.

PR #386, opened 2020-07-11
Raz/mgmt 1414 reset command

RazRegev:
http://10.0.9.195:8080/job/bm-inventory-subsystem-test/403/console


filanov:
Adding subsystem test will probably be a good idea.

RazRegev:
> Adding subsystem test will probably be a good idea.

We already have, I just changed the name and the args but the command was exists with subsystem tests

PR #390, opened 2020-07-12
MGMT-1444 Move ignition generation outside of install cluster async p…

filanov:
…art db transaction

Change install cluster asyn part to handle errors with defer
Add requeset ID to install cluster async part context

This will reduce the transaction time to minimum.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -29,6 +29,7 @@ import (
 	"github.com/filanov/bm-inventory/pkg/filemiddleware"
 	"github.com/filanov/bm-inventory/pkg/job"
 	logutil "github.com/filanov/bm-inventory/pkg/log"
+	"github.com/filanov/bm-inventory/pkg/requestid"
 	awsS3CLient "github.com/filanov/bm-inventory/pkg/s3Client"
 	"github.com/filanov/bm-inventory/pkg/transaction"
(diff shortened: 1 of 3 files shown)


filanov:
subsystem passed:
```
Ran 39 of 39 Specs in 492.031 seconds
SUCCESS! -- 39 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestSubsystem (492.03s)
PASS
ok  	github.com/filanov/bm-inventory/subsystem	492.171s
```

internal/bminventory/inventory.go
```diff
@@ -740,7 +748,7 @@ func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, c
 		return errors.Wrapf(err, "Generating kubeconfig files %s failed for cluster %s", jobName, cluster.ID)
 	}
 
-	return b.clusterApi.SetGeneratorVersion(&cluster, b.Config.KubeconfigGenerator, tx)
+	return b.clusterApi.SetGeneratorVersion(&cluster, b.Config.KubeconfigGenerator, b.db)
```
ori-amizur, line 751: Q: Is it always safe to use the same db object in multiple go routines?

filanov: Isn't it the same when we handle multiple APIs in parallel?

PR #393, opened 2020-07-13
MGMT-1487 Agent might not restart in case of failure

eranco74:
Updateed the restart policy:
Set StartLimitIntervalSec to 0 to disable any kind of rate limiting
Set RestartSec to 3 seconds delay between executions


eranco74:
http://10.0.9.195:8080/job/test-infra/1698/

PR #395, opened 2020-07-13
MGMT-1446 Transfer Controller image flag to the Assisted-Installer

YuviGold:
Completing eranco74/assisted-installer#75

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -10,11 +10,12 @@ import (
 )
 
 type Versions struct {
-	SelfVersion         string `envconfig:"SELF_VERSION" default:"quay.io/ocpmetal/installer-image-build:stable"`
-	ImageBuilder        string `envconfig:"IMAGE_BUILDER" default:"quay.io/ocpmetal/installer-image-build:stable"`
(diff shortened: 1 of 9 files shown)


internal/versions/versions.go
```diff
@@ -15,6 +15,7 @@ type Versions struct {
 	AgentDockerImg      string `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/ocpmetal/agent:stable"`
 	KubeconfigGenerator string `envconfig:"KUBECONFIG_GENERATE_IMAGE" default:"quay.io/ocpmetal/ignition-manifests-and-kubeconfig-generate:stable"`
 	InstallerImage      string `envconfig:"INSTALLER_IMAGE" default:"quay.io/ocpmetal/assisted-installer:stable"`
+	ControllerImage     string `envconfig:"CONTROLLER_IMAGE" default:"quay.io/ocpmetal/assisted-installer-controller:stable"`
```
filanov, line 18: stable?

eranco74: @filanov why do we have stable as default?

YuviGold: Old versions. changed to latest.

PR #402, opened 2020-07-14
mgmt-1501 Wrong boot order lead to cluster error state

eranco74:


internal/cluster/installing.go
```diff
@@ -64,14 +64,17 @@ func (i *installingState) getClusterInstallationState(ctx context.Context, c *co
 	}
 
 	// Cluster is installing
-	mastersInInstalling := mappedMastersByRole[intenralhost.HostStatusInstalling]
-	mastersInInstallingInProgress := mappedMastersByRole[intenralhost.HostStatusInstallingInProgress]
-	if (len(mastersInInstalling) + len(mastersInInstallingInProgress) + len(mastersInInstalled)) >= minHostsNeededForInstallation {
+	mastersInSomeInstallingStatus := len(mappedMastersByRole[intenralhost.HostStatusInstalling]) +
```
YuviGold, line 67: I think it should check for one of the hosts to be in an "error" state. There will be new statuses and it will get missed once again.

eranco74: I agree,
Yet, if we don't know what should be the status (non of the if statements match) I think we should set it to error so we will notice the bug and fix the code.
I just added a log so we can see what is the status of each host that led to the error state

eranco74:
Ran 41 of 41 Specs in 644.373 seconds
SUCCESS! -- 41 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestSubsystem (644.38s)

PR #403, opened 2020-07-14
MGMT-1479 Add stage_started_at and stage_updated_at & MGMT-1493 fix s…

YuviGold:
…tatus_updated_at

[MGMT-1493](https://issues.redhat.com/browse/MGMT-1493) Fixing `status_updated_at`
[MGMT-1479](https://issues.redhat.com/browse/MGMT-1479) Adding `stage_started_at` and `stage_updated_at` fields
[MGMT-1495](https://issues.redhat.com/browse/MGMT-1495) Fixing `UpdateClusterStatus` and `UpdateHostStatus` functions confusion and duplication

### DIFF
--- a/internal/cluster/common.go
+++ b/internal/cluster/common.go
@@ -32,45 +32,29 @@ const (
 	statusInfoPreparingForInstallationTimeout = "Preparing cluster for installation timeout"
 )
 
-type UpdateReply struct {
-	State     string
-	IsChanged bool
-}
-
 type baseState struct {
 	//TODO remove when res: https://github.com/golangci/golangci-lint/issues/537
 	log logrus.FieldLogger //nolint:structcheck
 	db  *gorm.DB           //nolint:structcheck
 }
 
-func updateClusterStatusUpdateReplay(log logrus.FieldLogger, db *gorm.DB, clusterId strfmt.UUID, srcStatus string,
-	newStatus string, statusInfo string) (*UpdateReply, error) {
-	wouldChange := srcStatus != newStatus
-
-	if err := updateClusterStatus(log, db, clusterId, srcStatus, newStatus, statusInfo); err != nil {
-		return nil, err
-	}
-
-	return &UpdateReply{
-		State:     newStatus,
-		IsChanged: wouldChange,
-	}, nil
-}
-
 func updateClusterStatus(log logrus.FieldLogger, db *gorm.DB, clusterId strfmt.UUID, srcStatus string,
-	newStatus string, statusInfo string, extra ...interface{}) error {
-	extra = append(extra, "status", newStatus, "status_info", statusInfo)
+	newStatus string, statusInfo string, extra ...interface{}) (*common.Cluster, error) {
+	var cluster *common.Cluster
+	var err error
+
+	extra = append(append(make([]interface{}, 0), "status", newStatus, "status_info", statusInfo), extra...)
 
 	if newStatus != srcStatus {
 		extra = append(extra, "status_updated_at", strfmt.DateTime(time.Now()))
 	}
 
-	if _, err := UpdateCluster(log, db, clusterId, srcStatus, extra...); err != nil {
-		return errors.Wrapf(err, "failed to update cluster %s state from %s to %s",
+	if cluster, err = UpdateCluster(log, db, clusterId, srcStatus, extra...); err != nil || *cluster.Status != newStatus {
+		return nil, errors.Wrapf(err, "failed to update cluster %s state from %s to %s",
 			clusterId, srcStatus, newStatus)
 	}
 
-	return nil
+	return cluster, nil
 }
 
 func UpdateCluster(log logrus.FieldLogger, db *gorm.DB, clusterId strfmt.UUID, srcStatus string, extra ...interface{}) (*common.Cluster, error) {
--- a/internal/cluster/common.go
+++ b/internal/cluster/common.go
@@ -43,54 +43,59 @@ type baseState struct {
 	db  *gorm.DB           //nolint:structcheck
 }
 
-func updateState(state string, statusInfo string, c *common.Cluster, db *gorm.DB, log logrus.FieldLogger) (*UpdateReply, error) {
-	updates := map[string]interface{}{"status": state, "status_info": statusInfo, "status_updated_at": strfmt.DateTime(time.Now())}
-	if *c.Status == clusterStatusReady && state == clusterStatusPrepareForInstallation {
-		updates["install_started_at"] = strfmt.DateTime(time.Now())
+func updateClusterStatus(status string, statusInfo string, c *common.Cluster, db *gorm.DB, log logrus.FieldLogger) (*UpdateReply, error) {
+	wouldChange := status != swag.StringValue(c.Status)
+
+	if err := updateClusterStateWithParams(log, status, statusInfo, c, db); err != nil {
+		return nil, err
 	}
-	dbReply := db.Model(&common.Cluster{}).Where("id = ? and status = ?",
-		c.ID.String(), swag.StringValue(c.Status)).Updates(updates)
-	if dbReply.Error != nil {
-		return nil, errors.Wrapf(dbReply.Error, "failed to update cluster %s state from %s to %s",
-			c.ID.String(), swag.StringValue(c.Status), state)
-	}
-	if dbReply.RowsAffected == 0 {
-		return nil, errors.Errorf("failed to update cluster %s state from %s to %s, nothing have changed",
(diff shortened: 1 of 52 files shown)


internal/cluster/common.go
```diff
@@ -41,52 +41,59 @@ type baseState struct {
...
+	// Status is queried as well to avoid races between different components.
+	dbReply := db.Model(&common.Cluster{}).Where("id = ? and status = ?",
+		c.ID.String(), swag.StringValue(c.Status)).Updates(updates)
 	if dbReply.Error != nil {
 		return errors.Wrapf(dbReply.Error, "failed to update cluster %s state from %s to %s",
-			c.ID.String(), srcStatus, swag.StringValue(c.Status))
+			c.ID.String(), swag.StringValue(c.Status), status)
 	}
-	if dbReply.RowsAffected == 0 && swag.StringValue(c.Status) != srcStatus {
+	if dbReply.RowsAffected == 0 {
```
ronniel1, line 86: There is a behaivor change here.
In the new version, if the old-status == new-status there will be an error. in the old version, there will not be an error

internal/cluster/common.go
```diff
@@ -41,52 +41,59 @@ type baseState struct {
...
 		return errors.Errorf("failed to update cluster %s state from %s to %s, nothing have changed",
-			c.ID.String(), srcStatus, swag.StringValue(c.Status))
+			c.ID.String(), swag.StringValue(c.Status), status)
 	}
-	c.StatusInfo = &statusInfo
-	log.Infof("Updated cluster <%s> status from <%s> to <%s> with fields: %s",
-		c.ID.String(), srcStatus, swag.StringValue(c.Status), updates)
+	log.Infof("updated cluster %s from state <%s> to state <%s>", c.ID.String(), swag.StringValue(c.Status), status)
+
+	if err := db.First(c, "id = ?", c.ID.String()).Error; err != nil {
```
ronniel1, line 92: This is another change from before, c passed as an argument is being changed
This has a far-reaching affect, since in most cases this c argument is passed along to many places
Seems kind of dangerous

internal/cluster/common.go
```diff
@@ -41,52 +41,59 @@ type baseState struct {
...
 	return &UpdateReply{
-		State:     state,
-		IsChanged: state != swag.StringValue(c.Status),
+		State:     status,
+		IsChanged: wouldChange,
 	}, nil
 }
 
-func updateClusterStateWithParams(log logrus.FieldLogger, srcStatus, statusInfo string, c *common.Cluster, db *gorm.DB,
+func updateClusterStateWithParams(log logrus.FieldLogger, status string, statusInfo string, c *common.Cluster, db *gorm.DB,
```
filanov, line 57: I'm confused. state machine change the status of the object `common.Cluster`
if `status` is the new status and `common.Cluster` already have the new status then `tatus != swag.StringValue(c.Status)` will be always the same no?

internal/cluster/common.go
```diff
@@ -41,52 +41,59 @@ type baseState struct {
...
 	extra ...interface{}) error {
 
-	updates := map[string]interface{}{"status": swag.StringValue(c.Status), "status_info": statusInfo}
+	updates := map[string]interface{}{"status": status, "status_info": statusInfo}
+
+	if status != swag.StringValue(c.Status) {
+		updates["status_updated_at"] = strfmt.DateTime(time.Now())
+	}
+	if *c.Status == clusterStatusReady && status == clusterStatusPrepareForInstallation {
```
filanov, line 65: please use `models.ClusterStatusPrepareForInstallation`

filanov: The same for all the other states in this function.

internal/cluster/common.go
filanov, line 86: what will happen if the everything stay the same? this function will fail no?

internal/cluster/transition.go
```diff
@@ -63,6 +64,14 @@ func (th *transitionHandler) PostResetCluster(sw stateswitch.StateSwitch, args s
...
-	return updateClusterStateWithParams(logutil.FromContext(params.ctx, th.log), sCluster.srcState,
+	return updateCluster(logutil.FromContext(params.ctx, th.log), sCluster.srcState,
 		params.reason, sCluster.cluster, params.db)
 }
+
+// Updates the status according to the cluster where the status equals the srcStatus
+func updateCluster(log logrus.FieldLogger, srcStatus, statusInfo string, c *common.Cluster, db *gorm.DB,
+	extra ...interface{}) error {
+	newStatus := c.Status
+	c.Status = &srcStatus
```
filanov, line 75: confusing.

internal/cluster/common.go
ronniel1, line 57: Why did you change the semantics here?
Before it was a change from the srcStatus to c.status
Now its a change from c.status to status

YuviGold: The naming were changed in order to make it clearer. Please look at the final changes.

internal/cluster/common.go
YuviGold: Changed to return a value instead of changing an object.

internal/cluster/common.go
YuviGold: Fixed.

internal/cluster/transition.go
YuviGold: Removed.

internal/cluster/common.go
YuviGold: This part was split into two functions - `UpdateCluster` verifies the DB behavior. While `UpdateClusterStatus` verifies that the  status has changed.

internal/cluster/common.go
YuviGold: Fixed.

internal/cluster/common.go
YuviGold: Removed the Cluster struct to avoid confusion.

internal/cluster/common.go
```diff
@@ -32,66 +32,58 @@ const (
...
-	dbReply := db.Model(&models.Cluster{}).Where("id = ? and status = ?", c.ID.String(), srcStatus).Updates(updates)
-	if dbReply.Error != nil {
-		return errors.Wrapf(dbReply.Error, "failed to update cluster %s state from %s to %s",
-			c.ID.String(), srcStatus, swag.StringValue(c.Status))
+
+	// Query by <cluster-id, status>
+	// Status is required as well to avoid races between different components.
+	dbReply := db.Model(&common.Cluster{}).Where("id = ? and status = ?", clusterId, srcStatus).Updates(updates)
+
+	if dbReply.Error != nil || dbReply.RowsAffected == 0 {
```
filanov, line 74: Not sure if we should add `swag.StringValue(c.Status) != srcStatus` validation as well. 
if all the tests passed then maybe it's ok not to add it. 
but i think that we may have a state that we try to update and leave the same state without any changes, and it will be bad if fails.

YuviGold: I don't know about the swagger casting. 
Testing works fine.
Why would you want to allow "Update" without any changes?

filanov: if all tests pass (including test-infa) then it's fine

internal/host/known.go
```diff
@@ -29,6 +29,6 @@ type knownState struct {
 	connectivityValidator connectivity.Validator
 }
 
-func (k *knownState) RefreshStatus(ctx context.Context, h *models.Host, db *gorm.DB) (*UpdateReply, error) {
+func (k *knownState) RefreshStatus(ctx context.Context, h *models.Host, db *gorm.DB) (*models.Host, error) {
```
ori-amizur, line 32: Why is the change in the return value required?

YuviGold: `Monitoring` functions (cluster, host) checks and notifies the user if states were changed.

internal/cluster/transition.go
```diff
@@ -129,6 +137,18 @@ type TransitionArgsHandlePreInstallationError struct {
...
+}
+
+func (th *transitionHandler) updateTransitionCluster(log logrus.FieldLogger, db *gorm.DB, state *stateCluster,
+	statusInfo string, extra ...interface{}) error {
+
+	if cluster, err := updateClusterStatus(log, db, *state.cluster.ID, state.srcState,
+		*state.cluster.Status, statusInfo, extra...); err != nil {
+		return err
+	} else {
+		state.cluster = cluster
```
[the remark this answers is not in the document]
YuviGold: Kept the previous behavior of updating the instance. I'm not sure if it is needed.

gamli75:
@YuviGold @avishayt why stage_started_at and stage_updated_at fields are not part of the models.HostProgress ?

PR #406, opened 2020-07-14
Removed "Finish Waiting for control plane" from host stages in the sw…

eranco74:
…agger


eranco74:
> Why are we removing this?
> What is the jira ticket?

https://issues.redhat.com/browse/MGMT-1482

eranco74:
This status is no longer required (not set by the assisted-installer).
Once we are done "Writing image to disk" we will "Wait for control plane", and once that is done we will move to "Rebooting".

PR #410, opened 2020-07-15
MGMT-1502: remove dns records on cluster reset

danielerez:
On reset cluster flow, DNS records should be cleaned.
Hence, invoke deleteDNSRecordSets on ResetCluster.


danielerez:
@romfreiman @ori-amizur @filanov

PR #414, opened 2020-07-15
Eran/mgmt 1367 validate

eranco74:
The second commit validate the pull secret is set prior before generating the ISO.
I'll push it only after this is reflected in the UI.
@avishayt I think we might want to enforce that the pull secret isn't changed once ISO is downloaded, no?


avishayt:
> The second commit validate the pull secret is set prior before generating the ISO.
> I'll push it only after this is reflected in the UI.
> @avishayt I think we might want to enforce that the pull secret isn't changed once ISO is downloaded, no?

Maybe it should be a property of the image?

PR #416, opened 2020-07-15
MGMT-1486: Image GC support for more than 1K objects

avishayt:


internal/imgexpirer/imgexpirer.go
```diff
@@ -42,14 +42,17 @@ func (m *Manager) ExpirationTask() {
 
 	log.Info("Image expiration monitor woke up, checking for expired images...")
 	// Currently limited to 1000 objects
```
filanov, line 44: is the comment still relevant?

avishayt: No it's not

avishayt:
> How did you test it?

I uploaded 2,500 small files to scality and made sure that it processed all of them.  I would love to add a subsystem test for the GC but don't really know where to begin.

internal/imgexpirer/imgexpirer.go
filanov: So you can probably delete it.

PR #418, opened 2020-07-15
Fix README.md indentation

masayag:
Signed-off-by: Moti Asayag <masayag@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,22 +21,28 @@ To push your build target to a Docker registry you first need to change the defa
 1. Login to quay.io using `docker login quay.io`.
 1. Export the `SERVICE` environment variable to your Docker registry, and pass a tag of your choice, e.g., "test":
 
-```shell script
+```sh
 export SERVICE=quay.io/<username>/bm-inventory:<tag>
 ```
 
 For the first build of the build container run:
-`skipper build bm-inventory-build`
+```shell
+skipper build bm-inventory-build
+```
 
 ## Build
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,22 +21,28 @@ To push your build target to a Docker registry you first need to change the defa
...
 1. Export the `SERVICE` environment variable to your Docker registry, and pass a tag of your choice, e.g., "test":
 
-```shell script
+```sh
 export SERVICE=quay.io/<username>/bm-inventory:<tag>
 ```
 
 For the first build of the build container run:
-`skipper build bm-inventory-build`
+```
```
avishayt, line 29:
```suggestion
```sh
```

avishayt: For all please

masayag: I suggest to add `shell` instead of `sh` since `sh` might highlight builtins that shouldn't be highlighted, e.g. line 154 where part of only `set` will be highlighted in `set-context` which shouldn't be the case.

I'll use ```sh where it adds value for using it.
In other cases, there will not be any effect on the output.

avishayt: No problem, fix and push :)

masayag: A preview with recent changes is available on https://github.com/filanov/bm-inventory/blob/a175c662ed5f9a988c6e7196aeb1f4a15ad83587/README.md

avishayt: Update the current commit and re-push.

PR #419, opened 2020-07-15
Raz Allow to abort installation in wrong boot order …

RazRegev:


subsystem/cluster_test.go
```diff
@@ -849,6 +849,34 @@ var _ = Describe("cluster install", func() {
...
+				host := getHost(clusterID, *hostID)
+				_, err = bmclient.Installer.RegisterHost(ctx, &installer.RegisterHostParams{
+					ClusterID: clusterID,
+					NewHostParams: &models.HostCreateParams{
+						HostID: host.ID,
+					},
+				})
+				waitForHostState(ctx, clusterID, *host.ID, models.HostStatusInstallingPendingUserAction,
+					defaultWaitForHostStateTimeout)
+				_, err = bmclient.Installer.CancelInstallation(ctx, &installer.CancelInstallationParams{ClusterID: clusterID})
```
eranco74, line 870: Add another test that first cancel and than the host re-register (and after validating host status reset)

RazRegev: There is already a test that validate host can register after cancel+reset

eranco74: I think we need a test that first cancel and than the host re-register (prior to the reset)

subsystem/cluster_test.go
```diff
@@ -849,6 +849,35 @@ var _ = Describe("cluster install", func() {
 					Expect(swag.StringValue(host.Status)).Should(Equal(models.HostStatusError))
 				}
 			})
+			It("[only_k8s]cancel installation - wrong boot order", func() {
```
filanov, line 852: it's a long test, please add steps `By(..)`

PR #420, opened 2020-07-15
MGMT-1500 Add preparing-for-installation state to hosts

filanov:
Swagger and auto generated added preparing-for-installation to host status

Main cleanup - remove unused option parameters

Install cluster use transaction to set cluster and hosts status to preparing-for-installation

Add DB parameter to PrepareForInstallation inroder to run it inside a transaction
GetMasterNodesIds return known and preparing-for-installation hosts

Subsystem waitForClusterInstallationToStart use models states instead of strings

Add preparing-for-installation state to hosts and PrepareForInstalltion api that will set the host into this state if current state is known
preparing-for-installation refresh function will validate that cluster is still in preparing for installation state, otherwise host will go into error state

Move HandlePreInstallError to state machine model


filanov:
```
Ran 41 of 41 Specs in 667.062 seconds
SUCCESS! -- 41 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestSubsystem (667.06s)
PASS
```

internal/host/statemachine.go
```diff
@@ -98,8 +103,8 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
 	sm.AddTransition(stateswitch.TransitionRule{
 		TransitionType:   TransitionTypeInstallHost,
 		Condition:        th.IsValidRoleForInstallation,
-		SourceStates:     []stateswitch.State{HostStatusKnown},
-		DestinationState: HostStatusInstalling,
+		SourceStates:     []stateswitch.State{stateswitch.State(models.HostStatusPreparingForInstallation)},
+		DestinationState: stateswitch.State(models.HostStatusInstalling),
```
ori-amizur, line 107: What about the negative condition? What should the state-machine do if th.IsValidRoleForInstallation is false? Seems to be missing.

internal/host/statemachine.go
```diff
@@ -98,8 +103,8 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
 	sm.AddTransition(stateswitch.TransitionRule{
 		TransitionType:   TransitionTypeInstallHost,
 		Condition:        th.IsValidRoleForInstallation,
-		SourceStates:     []stateswitch.State{HostStatusKnown},
-		DestinationState: HostStatusInstalling,
+		SourceStates:     []stateswitch.State{stateswitch.State(models.HostStatusPreparingForInstallation)},
```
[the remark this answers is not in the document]
filanov: Good point. i will move the validation to `TransitionTypePrepareForInstallation`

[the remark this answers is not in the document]
filanov: Then it will resolve both issues.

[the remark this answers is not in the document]
filanov: fixed.

internal/host/statemachine.go
filanov: fixed.

ori-amizur: How is the negative condition handled?

ori-amizur: Looked at the code.  I think it is misleading.  Maybe better add transition that explicitly generates an error in the Transition or PostTransition callback.  Otherwise it is side-effect.  Then IsValidRoleForInstallation can be pure boolean condition.

filanov: The async part will fail and cluster will go into error state.
Then host monitor will move this host into error as well: https://github.com/filanov/bm-inventory/pull/420/files#diff-9023b654421db6c04cd1d373079ab684R31

filanov: I moved the validate `IsValidRoleForInstallation` to the sync part so what will fail?

filanov:
Test infra passed http://10.46.10.2:8080/job/test-infra/1760

internal/host/statemachine.go
ori-amizur: Nothing will fail, I just think the code is misleading that for false evaluation is returning an error.

filanov: Not sure i understand. If the transition fails nothing will change and the api will return an error.
The only failure here is if the DB will not be reachable so in this case setting the host to error or any other state will not work. 
So maybe i missed something, but i don't understand what is misleading.

ori-amizur: Boolean condition should return true or false

filanov: Agree, we can resolve it in a different PR because it's an existing code and not something new.

filanov: This validation will be removed once you push the pending state no? 
Because the host will not be ready/known if it doesn't have a role.

ori-amizur: Yes

PR #422, opened 2020-07-15
MGMT-1453 fix: Ui shows one host in error but actually all are in err…

oshercc:
since cluster error state is a final state we don't really refresh when in error.
so the state-info is not updated after the first transaction to an error state.

there is the option to refresh the cluster state when in error, but I am not sure the exact number of masters in error needs to be in the cluster state info. 
@filanov what do you think?


avishayt:
Maybe more elegant to refresh state once when moving to error?  It may solve some other issues in the future.

oshercc:
@avishayt, In that case, the state info will not update anyway.

PR #423, opened 2020-07-15
Avishay/mgmt 1460

avishayt:


internal/events/event_test.go
```diff
@@ -118,7 +120,7 @@ var _ = Describe("Events library", func() {
 			ctx := context.Background()
 			rid1 := uuid.NewRandom().String()
 			ctx = requestid.ToContext(ctx, rid1)
-			theEvents.AddEvent(ctx, "1", "event1", time.Now(), "2")
+			theEvents.AddEvent(ctx, "1", models.EventSeverityInfo, "event1", time.Now(), "2")
```
ronniel1, line 123: There is no check on what severity was written to the DB

avishayt: Fixed

PR #424, opened 2020-07-16
Eran/mgmt 1544 Add move host API

eranco74:


filanov:
You will need to add empty implementation.

PR #428, opened 2020-07-16
Eran/mgmt 1551 support installing host in an external cluster

eranco74:
Added install action: `POST   /clusters/{cluster_id}/hosts/{host_id}/actions/install`
The API is getting the worker ignition as string and uploading it to s3.
And than set the host status to installing.
This flow need to use this image of assisted-installer: https://github.com/eranco74/assisted-installer/pull/80


swagger.yaml
```diff
@@ -1621,16 +1655,16 @@ definitions:
...
+      - Starting installation
+      - Start Waiting for control plane
+      - Installing
+      - Writing image to disk
+      - Rebooting
+      - Waiting for ignition
+      - Configuring
+      - Joined
+      - Done
+      - Failed
```
avishayt, line 1667: Why?

swagger.yaml
```diff
@@ -1795,10 +1829,10 @@ definitions:
...
         pattern: '^([0-9]{1,3}\.){3}[0-9]{1,3}\/[0-9]|[1-2][0-9]|3[0-2]?$'
       free_addresses:
-          type: array
-          items:
-            type: string
-            format: ipv4
+        type: array
+        items:
+          type: string
+          format: ipv4
```
avishayt, line 1835: Why?

swagger.yaml
eranco74: auto-format

swagger.yaml
eranco74: auto-format

avishayt: Please manual-format it the way it was :)

internal/bminventory/inventory.go
```diff
@@ -1455,6 +1455,37 @@ func (b *bareMetalInventory) EnableHost(ctx context.Context, params installer.En
...
+
+	if err := b.db.First(&host, "id = ? and cluster_id = ?", params.HostID, params.ClusterID).Error; err != nil {
+		if gorm.IsRecordNotFoundError(err) {
+			log.WithError(err).Errorf("host %s not found", params.HostID)
+			return common.NewApiError(http.StatusNotFound, err)
+		}
+		log.WithError(err).Errorf("failed to get host %s", params.HostID)
+		return common.NewApiError(http.StatusInternalServerError, err)
+	}
+	objectKey := fmt.Sprintf("%s/%s-worker.ign", params.ClusterID, params.HostID)
```
masayag, line 1471: Shouldn't the key match the one being search for from https://github.com/eranco74/assisted-installer/pull/80/files#diff-8bea30c4cde4eea8efce766b4de08e43R100 ?
`ignitionFileName := fmt.Sprintf("%s-%s.ign", i.Config.HostID, i.Config.Role)`

internal/bminventory/inventory.go
```diff
@@ -1455,6 +1455,37 @@ func (b *bareMetalInventory) EnableHost(ctx context.Context, params installer.En
...
+	if err := b.s3Client.PushDataToS3(ctx, []byte(*params.InstallHostParams.WorkerIgnition), objectKey, b.S3Bucket); err != nil {
+		return installer.NewUploadClusterIngressCertInternalServerError().
```
masayag, line 1473: Is this the desired error in this case ? Shouldn't it be some other generated error as `installer.NewInstallHostInternalServerError`

swagger.yaml
masayag: This seems like the correct indentation. The array element should have an offset from the 'enum:'.

swagger.yaml
masayag: Although indentation fixes for this file shouldn't be part of this PR, they are correct.
This worth a separate PR to be fixed on 'master' branch and after merge, rebase 'roks' on top of it.

internal/bminventory/inventory.go
eranco74: It matchs:
objectKey = <cluster-id>/<host-id>-worker.ign
assisted-installer - ignitionFileName = <host-id>-worker.ign
ignitionFileName is passed to "DownloadFile" https://github.com/eranco74/assisted-installer/blob/8ea53c1c12db21722048de1f44fc9bb03ed277c1/src/installer/installer.go#L227
and the server add the cluster-id:
https://github.com/filanov/bm-inventory/blob/master/internal/bminventory/inventory.go#L1586

PR #432, opened 2020-07-18
OCPBUGSM-4255 Added reset and cancel to eventsHandler

RazRegev:
http://10.46.10.2:8080/job/bm-inventory-subsystem-test/554/console

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -363,12 +363,20 @@ func (m *Manager) UpdateHostname(ctx context.Context, h *models.Host, hostname s
 }
 
 func (m *Manager) CancelInstallation(ctx context.Context, h *models.Host, reason string, db *gorm.DB) *common.ApiErrorResponse {
+	eventSeverity := models.EventSeverityInfo
+	eventInfo := fmt.Sprintf("Installation canceled for host %s", m.GetHostname(h))
+	defer func() {
+		m.eventsHandler.AddEvent(ctx, h.ID.String(), eventSeverity, eventInfo, time.Now(), h.ClusterID.String())
+	}()
+
 	err := m.sm.Run(TransitionTypeCancelInstallation, newStateHost(h), &TransitionArgsCancelInstallation{
 		ctx:    ctx,
 		reason: reason,
 		db:     db,
 	})
 	if err != nil {
+		eventSeverity = models.EventSeverityError
+		eventInfo = fmt.Sprintf("Failed to cancel installation of host %s. Error: %s", m.GetHostname(h), err.Error())
 		return common.NewApiError(http.StatusConflict, err)
 	}
 	return nil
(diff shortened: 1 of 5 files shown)


internal/host/host.go
```diff
@@ -385,12 +393,20 @@ func (m *Manager) IsRequireUserActionReset(h *models.Host) bool {
 }
 
 func (m *Manager) ResetHost(ctx context.Context, h *models.Host, reason string, db *gorm.DB) *common.ApiErrorResponse {
+	eventSeverity := models.EventSeverityInfo
+	eventInfo := fmt.Sprintf("Updated status of host %s to %s", m.GetHostname(h), models.HostStatusResetting)
```
avishayt, line 397:
```suggestion
	eventInfo := fmt.Sprintf("Reset host installation")
```

internal/host/host.go
```diff
@@ -385,12 +393,20 @@ func (m *Manager) IsRequireUserActionReset(h *models.Host) bool {
...
+	}()
+
 	err := m.sm.Run(TransitionTypeResetHost, newStateHost(h), &TransitionArgsResetHost{
 		ctx:    ctx,
 		reason: reason,
 		db:     db,
 	})
 	if err != nil {
+		eventSeverity = models.EventSeverityError
+		eventInfo = fmt.Sprintf("Failed to reset installation of host %s. Error: %s", m.GetHostname(h), err.Error())
```
avishayt, line 409:
```suggestion
		eventInfo = fmt.Sprintf("Failed to reset host installation. Error: %s", err.Error())
```

internal/host/host.go
```diff
@@ -363,12 +363,20 @@ func (m *Manager) UpdateHostname(ctx context.Context, h *models.Host, hostname s
 }
 
 func (m *Manager) CancelInstallation(ctx context.Context, h *models.Host, reason string, db *gorm.DB) *common.ApiErrorResponse {
+	eventSeverity := models.EventSeverityInfo
+	eventInfo := fmt.Sprintf("Updated status of host %s to %s", m.GetHostname(h), models.HostStatusError)
+	defer func() {
+		m.eventsHandler.AddEvent(ctx, h.ID.String(), eventSeverity, eventInfo, time.Now(), h.ClusterID.String())
```
avishayt, line 369: This should be the host ID

internal/host/host.go
```diff
@@ -385,12 +393,20 @@ func (m *Manager) IsRequireUserActionReset(h *models.Host) bool {
 }
 
 func (m *Manager) ResetHost(ctx context.Context, h *models.Host, reason string, db *gorm.DB) *common.ApiErrorResponse {
+	eventSeverity := models.EventSeverityInfo
+	eventInfo := fmt.Sprintf("Reset host installation")
+	defer func() {
+		m.eventsHandler.AddEvent(ctx, h.ID.String(), eventSeverity, eventInfo, time.Now(), h.ClusterID.String())
```
avishayt, line 399: This should be the host ID

avishayt:
@ronniel1 Do you think there's any reasonable way to add events in middleware for API success and failure?

ronniel1:
> @ronniel1 Do you think there's any reasonable way to add events in middleware for API success and failure?

I guess there is.
You want an event on any API call success/failure?

avishayt:
> > @ronniel1 Do you think there's any reasonable way to add events in middleware for API success and failure?
> 
> I guess there is.
> You want an event on any API call success/failure?

For every non-GET success, and for every failure.  The failure events should include the error, and they should all look like English (probably the hard part).

ronniel1:
> > > @ronniel1 Do you think there's any reasonable way to add events in middleware for API success and failure?
> > 
> > 
> > I guess there is.
> > You want an event on any API call success/failure?
> 
> For every non-GET success, and for every failure. The failure events should include the error, and they should all look like English (probably the hard part).

What about PostNextStepReply?

Do we really want an event for every succesfulll Save&update operation?

avishayt:
> > > > @ronniel1 Do you think there's any reasonable way to add events in middleware for API success and failure?
> > > 
> > > 
> > > I guess there is.
> > > You want an event on any API call success/failure?
> > 
> > 
> > For every non-GET success, and for every failure. The failure events should include the error, and they should all look like English (probably the hard part).
> 
> What about PostNextStepReply?
> 
> Do we really want an event for every succesfulll Save&update operation?

OK, good points on both.  I guess it's easier to just do it manually...

PR #433, opened 2020-07-18
Igal/postgress

tsorya:
Thanks a lot for @ronniel1, mostly it is his job.
Adding postgres, removing mysql support
Unitests:
1. Spawning container for postgress (small one)
2. Every unitest will work on it's own database on "beforeeach" and delete on "aftereach".
3.Adding common functions for test db usage

test-infra passes :
http://10.46.10.2:8080/job/test-infra/1830/

UI looks fine :

Screenshot from 2020-07-18 22-14-06
Screenshot from 2020-07-18 22-13-42

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ else ifdef DEPLOY_MANIFEST_TAG
   DEPLOY_TAG_OPTION = --deploy-manifest-tag "$(DEPLOY_MANIFEST_TAG)"
 endif
 
-deploy-all: create-build-dir deploy-namespace deploy-mariadb deploy-s3 deploy-route53 deploy-service
+deploy-all: create-build-dir deploy-namespace deploy-postgres deploy-s3 deploy-route53 deploy-service
 	echo "Deployment done"
 
 deploy-ui: deploy-namespace
@@ -124,6 +124,9 @@ deploy-role: deploy-namespace
 deploy-mariadb: deploy-namespace
 	python3 ./tools/deploy_mariadb.py
 
+deploy-postgres: deploy-namespace
+	python3 ./tools/deploy_postgres.py
+
 deploy-test:
 	export SERVICE=quay.io/ocpmetal/bm-inventory:test && export TEST_FLAGS=--subsystem-test && \
 	$(MAKE) update-minikube deploy-all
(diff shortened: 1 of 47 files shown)


Makefile
```diff
@@ -152,7 +155,10 @@ deploy-grafana: create-build-dir
 deploy-monitoring: deploy-olm deploy-prometheus deploy-grafana
 
 unit-test:
-	go test -v $(or ${TEST}, ${TEST}, $(shell go list ./... | grep -v subsystem)) -cover
+	docker stop postgres || true
+	docker run -d  --rm --name postgres -e POSTGRES_PASSWORD=admin -e POSTGRES_USER=admin -p 127.0.0.1:5432:5432 postgres:12.3-alpine -c 'max_connections=10000'
```
filanov, line 159: It's unit tests. it's an overkill to run docker for unit tests.

internal/host/host_test.go
```diff
@@ -32,6 +33,34 @@ var defaultHwInfo = "default hw info"                                 // invalid
 var defaultInventoryS = "default inventory"                           // invalid inventory info used only for tests
 var defaultProgressStage = models.HostStage("default progress stage") // invalid progress stage used only for tests
 
+//func common.PrepareTestDB(dbName string) *gorm.DB {
```
filanov, line 36: please delete if not needed.

Makefile
tsorya: Have idea? U are welcome

tsorya: Till now we had database not sure how easy it will to change all to work without any db. Postgres doesn't have in memory option

romfreiman: time docker run  -d --rm --name postgres -e POSTGRES_PASSWORD=admin -e POSTGRES_USER=admin -p 127.0.0.1:5432:5432 postgres:12.3-alpine -c 'max_connections=10000'
a14c964a9d5a3099422be646b70bce1e707ca797f2daa7f81424b5f10517c28a

real	0m0.471s

filanov: it's not only the time, it's that we will not have the option to run test on a file without using the make file 
and it stops being a unit-test when you need to run docker for it

ronniel1: We sill not be able to debug the tests within the IDE

tsorya: I don't understand the problem. We never run tests within idea and without make file

ronniel1: During implementation, I always run the UTs from the IDE

filanov: Same here.

tsorya: I still think we can use this way. I didn't find a way to run postgres without container and in memory. It is not sqlit

tsorya: I think if u want to run from idea u can mock db and not to use postgres

filanov: If the process become complicated then people will put less effort into it
Unit test should be something that simple to write and debug.
If you will have DB issues it will be complicated to debug.

tsorya: Really guys. If you have option to have db that works with postgres , will be glad to hear.
I think it is enough for now

ronniel1: @tsorya  this is a big deal. It means there is no way to debug the test.
See https://github.com/filanov/bm-inventory/pull/437 for a solution

deploy/postgres/postgres-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+      app: postgres
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: postgres
+    spec:
+      containers:
+        - name: postgres
+          image: postgres:12.3-alpine
```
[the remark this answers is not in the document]
filanov: @ronniel1

PR #437, opened 2020-07-19
Ronnie/postgress over igal

ronniel1:


cmd/main.go
```diff
@@ -100,10 +100,10 @@ func main() {
 		kclient = nil
 	}
 
-	db, err := gorm.Open("mysql",
-		fmt.Sprintf("admin:admin@tcp(%s:%s)/installer?charset=utf8&parseTime=True&loc=Local",
+	// Connect to db
+	db, err := gorm.Open("postgres",
+		fmt.Sprintf("host=%s port=%s user=admin dbname=installer password=admin sslmode=disable",
```
filanov, line 105: if we already here maybe we should get the user db name and the password as parameters as well?

internal/common/common_unitest_db.go
```diff
@@ -0,0 +1,85 @@
...
+	err := gDbCtx.pool.Purge(gDbCtx.resource)
+	Expect(err).ShouldNot(HaveOccurred())
+	gDbCtx.pool = nil
+}
+
+func PrepareTestDB(dbName string, extrasSchemas ...interface{}) *gorm.DB {
+	Expect(gDbCtx.pool).ShouldNot(BeNil())
+	Expect(gDbCtx.resource).ShouldNot(BeNil())
+	var dbTemp *gorm.DB
+	err := gDbCtx.pool.Retry(func() error {
```
filanov, line 47: Is there a timeout to this retry?

filanov:
@ronniel1 where can we see the logs of the DB? is there an option to set it to stdout?

internal/common/common_unitest_db.go
ronniel1: I think it is 15 minutes

filanov: anyway default go test timeout is 10min. 
how long does it take to run all the tests this way?

cmd/main.go
ronniel1: done

Makefile
```diff
@@ -152,7 +152,10 @@ deploy-grafana: create-build-dir
 deploy-monitoring: deploy-olm deploy-prometheus deploy-grafana
 
 unit-test:
-	go test -v $(or ${TEST}, ${TEST}, $(shell go list ./... | grep -v subsystem)) -cover
+	docker stop postgres || true
+	docker run -d  --rm --name postgres -e POSTGRES_PASSWORD=admin -e POSTGRES_USER=admin -p 127.0.0.1:5432:5432 postgres:12.3-alpine -c 'max_connections=10000'
+	SKIP_UT_DB=1 go test -v $(or ${TEST}, ${TEST}, $(shell go list ./... | grep -v subsystem)) -cover
```
filanov, line 157: if this part will return an error will the make process continue and kill postgres?

.drone.yml
```diff
@@ -2,12 +2,10 @@ kind: pipeline
 name: default
 
 steps:
-  - name: mariadb
-    image: mariadb
+  - name: postgres
+    image: postgres:12.3-alpine
```
mhrivnak, line 6: Downstream I think we'll want to use this: https://catalog.redhat.com/software/containers/rhel8/postgresql-12/5db133bd5a13461646df330b

mhrivnak: And here is an upstream option: https://hub.docker.com/r/centos/postgresql-12-centos7

ronniel1: done

Makefile
ronniel1: probably not will fix

PR #438, opened 2020-07-19
MGMT-1565 change cluster refresh status using get cluster from db once

filanov:
It will prevent races when we get cluster with state A and then because of a race
we can get a cluster with a different state.


filanov:
```
Ran 45 of 45 Specs in 619.159 seconds
SUCCESS! -- 45 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestSubsystem (619.16s)
PASS
```

PR #442, opened 2020-07-19
MGMT-1241: Clearer validation messages, disk in GB not GiB

avishayt:
Clearer validation error messages with units in GB/GiB and not bytes.
Also physical disk capacity should be in base 10.

Example outputs:
host requires at least 2 CPU cores, got 1; host requires at least 8 GiB RAM, got 3 GiB
master host requires at least 8 GiB RAM, got 8 GiB
host requires a disk of at least 120 GB
host with hostname \"non_unique_hostname\" already exists


avishayt:
> This change is not in line with the RefreshStatus change, so it needs to be updated

Yes I know, waited for it to be merged.  Will do.

PR #445, opened 2020-07-20
MGMT-1339 - BMH CRs should not be created for disabled hosts

yevgeny-shnaidman:


yevgeny-shnaidman:
i knew i should have sent the PR to somebody else :)

internal/installcfg/installcfg.go
```diff
@@ -71,7 +71,7 @@ type InstallerConfigBaremetal struct {
 func countHostsByRole(cluster *common.Cluster, role models.HostRole) int {
 	var count int
 	for _, host := range cluster.Hosts {
-		if host.Role == role {
+		if *host.Status != models.HostStatusDisabled && host.Role == role {
```
filanov, line 74:
```suggestion
		if swag.StringValue(host.Status) != models.HostStatusDisabled && host.Role == role {
```

romfreiman:
@yevgeny-shnaidman is the title correct? Disabled hosts should not be part of the install config at all

yevgeny-shnaidman:
yes, this is the the part of the config that deals with hosts, so the way we can check it is to see that no BMHs for created for those hosts

PR #447, opened 2020-07-20
Ori/mgmt 1380 RefreshStatus, Validations, and pending-for-input

ori-amizur:
Things to look for in this PR:
- validationID support: swagger.yaml, and new file internal/host/validation_id.go
- refresh status pre-processing: new files - internal/host/refresh_status_preprocessor.go and internal/host/validator.go
- refresh-status: Mainly in host/transitions.go, host/statemachine.go and host/host.go. The transitions mostly follow the transitions mentioned in https://docs.google.com/document/d/12xqSFVMgjfOCG798tQV6XHnghV6zkIhoiA028e8YYqU/edit although the document is not completely up to date.  Refresh status changes remove old refresh-status related code to avoid logic duplication.
- Add pending-for-input state: internal/host/statemachine.go, internal/host/host.go


internal/host/statemachine.go
```diff
@@ -156,5 +157,125 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
 		PostTransition:   th.PostPrepareForInstallation,
 	})
 
+	// Refresh host
```
filanov, line 160: you are missing refresh for preparing-for-installation

internal/host/host.go
```diff
@@ -231,17 +169,29 @@ func (m *Manager) UpdateInventory(ctx context.Context, h *models.Host, inventory
 	return m.db.Model(h).Update("inventory", inventory).Error
 }
 
-func (m *Manager) RefreshStatus(ctx context.Context, h *models.Host, db *gorm.DB) (*models.Host, error) {
-	state, err := m.getCurrentState(swag.StringValue(h.Status))
+func (m *Manager) RefreshStatus(ctx context.Context, h *models.Host, db *gorm.DB) error {
+	if db == nil {
+		db = m.db
+	}
+	vc, err := newValidationContext(h, db)
```
filanov, line 176: why do you need to create `newValidationContext` why not just have a function that return the result below?

internal/host/validation_id.go
```diff
@@ -0,0 +1,41 @@
+package host
+
+import (
+	"fmt"
+
+	"github.com/filanov/bm-inventory/models"
+)
+
+type validationID models.HostValidationID
```
YuviGold, line 9: Why are the models definitions duplicated? Can't we use the `models.HostValidationIDXXX` directly?

internal/host/statemachine.go
```diff
@@ -156,5 +157,125 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
...
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType:   TransitionTypeRefresh,
+		SourceStates:     []stateswitch.State{HostStatusDiscovering, HostStatusInsufficient, HostStatusKnown, HostStatusPendingForInput, HostStatusDisconnected},
```
filanov, line 163: need to start working with auto generated host states and the consts that we write, please move to `models.HostStatusDiscovering...`

internal/host/statemachine.go
```diff
@@ -156,5 +157,125 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
...
+		Condition:        stateswitch.And(If(IsConnected), stateswitch.Not(If(HasInventory))),
+		DestinationState: HostStatusDiscovering,
+		PostTransition:   th.PostRefreshHost(statusInfoDiscovering),
+	})
+
+	var hasMinRequiredHardware = stateswitch.And(If(HasMinValidDisks), If(HasMinCPUCores), If(HasMinMemory))
+
+	var requiredInputFieldsExist = stateswitch.And(If(IsMachineCidrDefined), If(IsRoleDefined))
+
+	var isSufficientForInstall = stateswitch.And(If(HasMemoryForRole), If(HasCPUCoresForRole), If(BelongsToMachineCidr), If(IsHostnameUnique), If(IsHostnameValid))
```
filanov, line 181: please break long lines.

internal/host/transition_test.go
```diff
@@ -611,3 +632,1119 @@ var _ = Describe("Enable", func() {
...
+					BelongsToMachineCidr: {status: ValidationPending, messagePattern: "Missing inventory or machine network CIDR"},
+				}),
+				errorExpected: false,
+			},
+			{
+				name:              "insufficient to disconnected",
+				srcState:          HostStatusInsufficient,
+				dstState:          HostStatusDisconnected,
+				statusInfoChecker: makeValueChecker(statusInfoDisconnected),
+				validationsChecker: makeJsonChecker(map[validationID]validationCheckResult{
```
YuviGold, line 750: Make a constant or a function returns this hash map occurring multiple times in this test

internal/host/validation_id.go
```diff
@@ -0,0 +1,41 @@
...
+	switch v {
+	case IsConnected, IsMachineCidrDefined, BelongsToMachineCidr:
+		return "network", nil
+	case HasInventory, HasMinCPUCores, HasMinValidDisks, HasMinMemory,
+		HasCPUCoresForRole, HasMemoryForRole, IsHostnameUnique, IsHostnameValid:
+		return "hardware", nil
+	case IsRoleDefined:
+		return "role", nil
+	}
+	return "", fmt.Errorf("Unexpected validation id %s", string(v))
```
filanov, line 36: please switch errors to `github.com/pkg/errors` or `common.NewAPIError`

subsystem/cluster_test.go
```diff
@@ -1350,11 +1350,14 @@ var _ = Describe("cluster install", func() {
 			ClusterID: clusterID,
 		})
 		Expect(err).To(Not(HaveOccurred()))
-		waitForHostState(ctx, clusterID, *h1.ID, "insufficient", 60*time.Second)
+		waitForHostState(ctx, clusterID, *h1.ID, "pending-for-input", 60*time.Second)
```
filanov, line 1353: please use `models.HostStatusPendingForInput`

internal/host/common.go
```diff
@@ -78,6 +61,13 @@ func updateHostStatus(log logrus.FieldLogger, db *gorm.DB, clusterId strfmt.UUID
 	return host, nil
 }
 
+func exists(db *gorm.DB, hostId, clusterId strfmt.UUID, where map[string]interface{}) bool {
```
YuviGold, line 64: Name isn't clear.. Change to something like `doesObjectExistsInDb`

internal/host/statemachine.go
YuviGold: Is there a way to add static-asserts to verify every status has a `refreshStatus` transition?
Maybe they should be added in some kind of a loop?

internal/host/statemachine.go
```diff
@@ -156,5 +157,125 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
...
+		Condition: stateswitch.And(If(IsConnected), If(HasInventory),
+			hasMinRequiredHardware,
+			requiredInputFieldsExist,
+			isSufficientForInstall),
+		DestinationState: HostStatusKnown,
+		PostTransition:   th.PostRefreshHost(""),
+	})
+
+	// Noop transitions
+	sm.AddTransition(stateswitch.TransitionRule{
```
YuviGold, line 233: Is there a way to add static-asserts to verify every status has a `TransitionTypeRefresh` transition?
Maybe they should be added in some kind of a loop?

internal/host/transition_test.go
ori-amizur: It is not the same map

internal/host/host.go
```diff
@@ -45,6 +31,7 @@ const (
 	HostStatusInstalled                   = "installed"
 	HostStatusError                       = "error"
 	HostStatusResetting                   = "resetting"
+	HostStatusPendingForInput             = "pending-for-input"
```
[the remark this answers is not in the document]
ori-amizur: OK, but not part of this PR

internal/host/validation_id.go
ori-amizur: 2 reasons:
1. For simplicity.  These long names are hard to read.  Easier to see and understand IsHostnameUnique than models.HostValidationIDsHostnameUnique, especially when involved in transitions with multiple conditions
2. validationID has functionality (category)

internal/host/common.go
ori-amizur: It reminded me "SQL" exists.  But it will be changed to hostExistsInDB

internal/host/host.go
ori-amizur: The function is m.rp.preprocess.  It needs data to run on.  This is the context.  Same as statewitch needs the statefull object and the arguments.

internal/host/statemachine.go
ori-amizur: OK.

internal/host/statemachine.go
ori-amizur: OK

internal/host/statemachine.go
ori-amizur: Yes, for the Noop transitions

subsystem/cluster_test.go
ori-amizur: OK

internal/host/validation_id.go
ori-amizur: OK

internal/host/statemachine.go
ronniel1: Maybe we should add a function:
```
AddNoopTransition(source, destination) 
```
Which will return the equivalent of this noop AddTransition() call?

ori-amizur: Added a loop for these transition.  For @ronniel1 maybe it should be part of stateswitch?

internal/host/statemachine.go
ori-amizur: This will make the rules very long and harder to read.  I still think that we should use shorter names, but they should be initialized from models.
models.HostStatusDiscovering is typed string constant. So to use it here, each one will need a cast to stateswtch.State.

filanov: I understand but if swagger will change then we may have bugs :)

ori-amizur: @filanov Maybe we can do something similar to what was done in validationID.  The local validationID type is initialized from the swagger constant.  Then, if the swagger will change, either the change will be adapted automatically, otherwise it will break.

internal/bminventory/inventory.go
```diff
@@ -616,19 +616,15 @@ func (c *clusterInstaller) installHosts(cluster *common.Cluster, tx *gorm.DB) er
...
 
-func (b *bareMetalInventory) validateHostsInventory(cluster *common.Cluster) error {
+func (b *bareMetalInventory) refreshAllHosts(ctx context.Context, cluster *common.Cluster) error {
 	for _, chost := range cluster.Hosts {
-		sufficient, err := b.hostApi.ValidateCurrentInventory(chost, cluster)
-		if err != nil {
-			msg := fmt.Sprintf("failed to validate host <%s> in cluster: %s", chost.ID.String(), cluster.ID.String())
-			b.log.WithError(err).Warn(msg)
-			return common.NewApiError(http.StatusInternalServerError, errors.Wrapf(err, msg))
+		if chost.Status == nil || *chost.Status != host.HostStatusKnown {
```
gamli75, line 621: What will happen if we have more than one host with status != known? I think we should return one message that includes all the hosts that their status != known 
We already have a method that checks all hosts status - validateAllHostsCanBeInstalled - maybe you can reuse it

swagger.yaml
```diff
@@ -1024,6 +1025,9 @@ definitions:
       status_info:
         type: string
         x-go-custom-tag: gorm:"type:varchar(2048)"
+      validations_info:
```
gamli75, line 1028: please add a description for this field

internal/bminventory/inventory.go
ori-amizur: refreshAllHosts perform refresh to sync the hosts status.  Before it was "IsSufficient" check which is only part of the validations that are needed for host to be installable.

swagger.yaml
ori-amizur: OK

internal/host/statemachine.go
filanov: Go for it

internal/bminventory/inventory.go
filanov: Please use swag.StringValue and don't access it directly

ori-amizur: OK

internal/host/statemachine.go
ori-amizur: Not as part of this PR

PR #450, opened 2020-07-21
S3 Expirer unit tests

avishayt:


filanov:
to increase the coverage (if it's needed) you can go to the directory and run
```
go test -v -coverprofile=coverage.out && go tool cover -html=coverage.out && rm coverage.out
```

avishayt:
@filanov The actual file is almost all covered - it shows a low percentage because of the mock in the same package

PR #451, opened 2020-07-21
MGMT-1490 Enforce stages enum ordering

YuviGold:
- Disallowing lower stages (except for "Failed" stage)
- **Behavior change** - `/progress` returns an error in case of an error. The assisted-installer should decide what to do with an error and not to hide it.

### DIFF
--- a/client/installer/update_host_install_progress_responses.go
+++ b/client/installer/update_host_install_progress_responses.go
@@ -7,9 +7,12 @@ package installer
 
 import (
 	"fmt"
+	"io"
 
 	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/strfmt"
+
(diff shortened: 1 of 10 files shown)


filanov:
@YuviGold how does assisted-installer react to errors?

YuviGold:
> @YuviGold how does assisted-installer react to errors?

Write a log and continues to run.

eranco74:
Didn't see this in time, 
Did you check that this play nice with cancel, reset, wrong boot order?

YuviGold:
> Didn't see this in time,
> Did you check that this play nice with cancel, reset, wrong boot order?

I ran the subsystem tests and test-infra.
I hope your scenarios are defined in the 1st flow.

PR #452, opened 2020-07-21
Fixing PSE deployment and added some adjustments on Prom svc

jparrill:
**Aditions:**

- Adding some Pep8 rules
- Fixing Prometheus and Grafana deployment in a non cluster-admin situation
- Added docstrings


filanov:
@jparrill @RazRegev you are working in the same area please sync.

jparrill:
Humm that error I gues is not related with the PR

PR #453, opened 2020-07-21
MGMT-1566: Add missing API events

avishayt:


filanov:
It's a bit strange to see that you define a parameter that is used only for the next line. you can break the event line and add  the message there. 
For example
```
		b.eventsHandler.AddEvent(ctx, params.ClusterID.String(), models.EventSeverityError, 
				"Failed to download image: error fetching from storage backend", time.Now())
```
But it's fine like this as well.

PR #461, opened 2020-07-22
Raz/mgmt 1583

RazRegev:


internal/host/monitor.go
```diff
@@ -15,8 +15,18 @@ func (m *Manager) HostMonitoring() {
 		log       = requestid.RequestIDLogger(m.log, requestID)
 	)
 
-	monitorStates := []string{HostStatusDiscovering, HostStatusKnown, HostStatusDisconnected, HostStatusInsufficient,
-		HostStatusPendingForInput, models.HostStatusPreparingForInstallation}
+	monitorStates := []string{
```
filanov, line 18: what are the states that we don't monitor?

internal/host/validator.go
```diff
@@ -416,3 +418,24 @@ func (v *validator) printHostnameValid(c *validationContext, status validationSt
...
+	switch status {
+	case ValidationSuccess:
+		return fmt.Sprintf("Cluster %s has no errors", c.cluster.ID.String())
+	case ValidationFailure:
+		return fmt.Sprintf("Cluster %s has errors. Aborting installation of host %s",
+			c.cluster.ID.String(), getRealHostname(c.host, c.inventory))
+	case ValidationPending:
+		return "Missing inventory"
+	default:
+		return fmt.Sprintf("Unexpected status %s", status)
```
filanov, line 439: Not sure will the user do with this string, do we need it?

RazRegev:
> No unit tests?

why not? added tests to transition_test

internal/host/monitor.go
RazRegev: preparing, installing-pending-user-action, resetting-pending-user-action, pending-user-input

internal/host/validator.go
```diff
@@ -420,3 +422,23 @@ func (v *validator) printHostnameValid(c *validationContext, status validationSt
 		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
+
+func (v *validator) noClusterErrors(c *validationContext) validationStatus {
+	if c.inventory == nil {
```
ori-amizur, line 427: You don't need the inventory for this validation - so this check is not needed

internal/host/validator.go
```diff
@@ -420,3 +422,23 @@ func (v *validator) printHostnameValid(c *validationContext, status validationSt
...
+	return boolValue(swag.StringValue(c.cluster.Status) != models.ClusterStatusError)
+}
+
+func (v *validator) printNoClusterErrors(c *validationContext, status validationStatus) string {
+	switch status {
+	case ValidationSuccess:
+		return fmt.Sprintf("Cluster %s has no errors", c.cluster.ID.String())
+	case ValidationFailure:
+		return fmt.Sprintf("Cluster %s has errors", c.cluster.ID.String())
+	case ValidationPending:
```
ori-amizur, line 439: Same as above

internal/host/statemachine.go
```diff
@@ -167,6 +167,19 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
...
 
+	// Abort host if cluster has errors
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeRefresh,
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusInstalling),
+			stateswitch.State(models.HostStatusInstallingInProgress),
+			stateswitch.State(models.HostStatusInstalled),
+		},
+		Condition:        stateswitch.Not(If(NoClusterErrors)),
```
ori-amizur, line 178: 1. Currently there is a Noop transitions for all of these states.  They have to be removed.
2. When there is such transition, you need to do opposite rule (Condition: If(NoClusterErrors)) even if it is without a PostTransition (Noop)

internal/host/common.go
```diff
@@ -23,6 +23,7 @@ const (
 	statusInfoResettingPendingUserAction = "Reboot the host into the installation image to complete resetting the installation"
 	statusInfoPreparingForInstallation   = "Preparing host for installation"
 	statusInfoPreparingTimedOut          = "Cluster is no longer preparing for installation"
+	statusInfoAbortingDueClusterErrors   = "Cluster has been aborted due errors"
```
ori-amizur, line 26: Wrong wording - Suggest "Cluster installation has been aborted due to errors"

internal/host/statemachine.go
```diff
@@ -167,6 +167,33 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
...
+		PostTransition:   th.PostRefreshHost(statusInfoAbortingDueClusterErrors),
+	})
+
+	// Noop transitions for cluster error
+	for _, state := range []stateswitch.State{
+		stateswitch.State(models.HostStatusInstalling),
+		stateswitch.State(models.HostStatusInstallingInProgress),
+		stateswitch.State(models.HostStatusInstalled),
+	} {
+		sm.AddTransition(stateswitch.TransitionRule{
```
filanov, line 189: Maybe we can add a `NoStateChange` to state object?
Something like 
```
func (sh *stateHost) SetState(state stateswitch.State) error {
	if NoStateChange == state {
		return nil
	}
	sh.host.Status = swag.String(string(state))
	return nil
}
```
@ori-amizur What do you think?

ori-amizur: Possible, but i think that if statewitch does not find any rule to fire it should not be an error. So, we will not need the noop transitions

filanov: But it is an error, you expected to do some operation and nothing happened. 
Another option is to ignore specific error https://github.com/filanov/stateswitch/blob/master/statemachine.go#L7

PR #464, opened 2020-07-22
MGMT-1603 Prevent dummy iso generation uploading to s3

filanov:
We only need to download the docker image to the cluster so there is no reason
to let the dummy job upload anything to s3.
Overwrite the default command into 'echo pass'

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -181,8 +181,9 @@ func (b *bareMetalInventory) generateDummyISOImage() {
 		requestID = requestid.NewID()
 		log       = requestid.RequestIDLogger(b.log, requestID)
 	)
+	// create dummy job without uploading to s3, we just need to pull the image
 	if err := b.job.Create(requestid.ToContext(context.Background(), requestID),
-		b.createImageJob(jobName, imgName, "Dummy")); err != nil {
+		b.createImageJob(jobName, imgName, "Dummy", false)); err != nil {
 		log.WithError(err).Errorf("failed to generate dummy ISO image")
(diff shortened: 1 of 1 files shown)


filanov:
```
kubectl logs dummyimage-00000000-0000-0000-0000-000000000000-2020072216pt8zw
pass
```

internal/bminventory/inventory.go
```diff
@@ -193,7 +194,12 @@ func getQuantity(s string) resource.Quantity {
 }
 
 // create discovery image generation job, return job name and error
-func (b *bareMetalInventory) createImageJob(jobName, imgName, ignitionConfig string) *batch.Job {
+// cmd is optional to overwrite default command
+func (b *bareMetalInventory) createImageJob(jobName, imgName, ignitionConfig string, cmd *string) *batch.Job {
+	var command []string
+	if cmd != nil {
+		command = strings.Split(swag.StringValue(cmd), " ")
+	}
```
avishayt, line 202:
```suggestion
func (b *bareMetalInventory) createImageJob(jobName, imgName, ignitionConfig string, upload boolean) *batch.Job {
	var command []string
	if !upload {
		command = strings.Split(swag.StringValue("echo pass"), " ")
	}
```

avishayt: s/upload/performUpload/

filanov:
> Would something like that be more readable?

Yes good option, thanks

internal/bminventory/inventory.go
filanov: fixed.

PR #467, opened 2020-07-23
MGMT-1596 Update hostname should be allowed from pending-for-input state

ori-amizur:


internal/host/host.go
```diff
@@ -312,7 +312,8 @@ func (m *Manager) UpdateRole(ctx context.Context, h *models.Host, role models.Ho
 
 func (m *Manager) UpdateHostname(ctx context.Context, h *models.Host, hostname string, db *gorm.DB) error {
 	hostStatus := swag.StringValue(h.Status)
-	allowedStatuses := []string{HostStatusDiscovering, HostStatusKnown, HostStatusDisconnected, HostStatusInsufficient}
+	allowedStatuses := []string{HostStatusDiscovering, HostStatusKnown, HostStatusDisconnected, HostStatusInsufficient,
```
filanov, line 315: please switch to `models.HostStatus...`

ori-amizur: This should be global change and not for a single PR

filanov: We should start somewhere. every time i give this comment i get the same reply, so why not start at least with new code or when someone change existing code, you already changing it so why not change it right now?

PR #470, opened 2020-07-23
MGMT-1231 Reset Installation: allow reset from various states

RazRegev:
We don't really need to set the cluster and host status to error before
cancelling installation. Resetting installing node is just like canceling and
then resetting (cancel cmd just stop podman, but also reset cmd).
Installed nodes will be set to ResettingPendingUserAction.
Update subsystem test to accept the new source states when resetting cluster.
Added unit tests.

### DIFF
--- a/internal/cluster/transition_test.go
+++ b/internal/cluster/transition_test.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/filanov/bm-inventory/internal/metrics"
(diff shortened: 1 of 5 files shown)


internal/cluster/transition_test.go
```diff
@@ -142,3 +142,62 @@ var _ = Describe("Transition tests", func() {
...
+		{state: models.ClusterStatusInstalling, success: true},
+		{state: models.ClusterStatusError, success: true},
+		{state: models.ClusterStatusInsufficient, success: false, statusCode: http.StatusConflict},
+		{state: models.ClusterStatusReady, success: false, statusCode: http.StatusConflict},
+		{state: models.ClusterStatusPreparingForInstallation, success: false, statusCode: http.StatusConflict},
+		{state: models.ClusterStatusFinalizing, success: false, statusCode: http.StatusConflict},
+		{state: models.ClusterStatusInstalled, success: false, statusCode: http.StatusConflict},
+	}
+
+	It("reset_cluster_cases", func() {
```
filanov, line 181: this test will run as a single test if something breaks it will be hard to debug. consider giving names and run in separate tests

internal/host/transition_test.go
```diff
@@ -293,6 +294,71 @@ var _ = Describe("HostInstallationFailed", func() {
...
+		{state: models.HostStatusInstallingPendingUserAction, success: false, statusCode: http.StatusConflict},
+		{state: models.HostStatusResettingPendingUserAction, success: false, statusCode: http.StatusConflict},
+		{state: models.HostStatusDisconnected, success: false, statusCode: http.StatusConflict},
+	}
+
+	acceptResetEvent := func(times int) {
+		mockEventsHandler.EXPECT().AddEvent(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(times)
+	}
+
+	It("reset_host_cases", func() {
```
filanov, line 339: this test will run as a single test if something breaks it will be hard to debug. consider giving names and run in separate tests

PR #473, opened 2020-07-23
Eran/mgmt 1618

eranco74:


internal/bminventory/inventory.go
```diff
@@ -468,6 +468,13 @@ func (b *bareMetalInventory) GenerateClusterISO(ctx context.Context, params inst
 			errors.New("Another request to generate an image has been recently submitted. Please wait a few seconds and try again.")))
 	}
 
+	if !cluster.PullSecretSet {
+		err := "Can't generate cluster ISO without pull secret"
```
filanov, line 472: it's confusing because err is a string and not an error. `errMsg` or `msg` will be clearer.

PR #474, opened 2020-07-23
MGMT-1577 Host role auto selection

filanov:
Automatically set host role after it send hw info.
First trying to fill the masters role, after 3 masters all roles will be set to workers
Using channel as a queue to prevent races between different hosts.
If user already set the role it will not be changed.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -1405,6 +1405,7 @@ func handleReplyByType(params installer.PostStepReplyParams, b *bareMetalInvento
 	switch params.Reply.StepType {
 	case models.StepTypeInventory:
 		err = b.hostApi.UpdateInventory(ctx, &host, stepReply)
+		b.hostApi.AutoRoleSelection(&host)
 	case models.StepTypeConnectivityCheck:
 		err = b.hostApi.UpdateConnectivityReport(ctx, &host, stepReply)
 	case models.StepTypeFreeNetworkAddresses:
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -12,8 +12,8 @@ import (
 	"github.com/filanov/bm-inventory/internal/metrics"
 	"github.com/filanov/bm-inventory/models"
 	logutil "github.com/filanov/bm-inventory/pkg/log"
+	"github.com/filanov/bm-inventory/pkg/requestid"
 	"github.com/filanov/stateswitch"
-
 	"github.com/go-openapi/swag"
 	"github.com/jinzhu/gorm"
 	"github.com/pkg/errors"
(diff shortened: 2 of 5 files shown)


filanov:
test-infra passed http://10.46.10.2:8080/job/test-infra/1956/

internal/host/host.go
```diff
@@ -420,3 +425,67 @@ func (m *Manager) reportInstallationMetrics(ctx context.Context, h *models.Host,
...
+		autoSelectedRole = models.HostRoleWorker
+		log              = logutil.FromContext(ctx, m.log)
+	)
+
+	if err := m.db.Take(&host, "id = ? and cluster_id = ?", h.ID.String(), h.ClusterID.String()).Error; err != nil {
+		log.WithError(err).Errorf("failed to get host %s for auto role selection", h.ID.String())
+		return
+	}
+	// check if role already selected
+	if funk.ContainsString([]string{string(models.HostRoleMaster), string(models.HostRoleWorker)}, string(host.Role)) {
```
avishayt, line 454: Why not role != ""

internal/host/host.go
```diff
@@ -420,3 +425,67 @@ func (m *Manager) reportInstallationMetrics(ctx context.Context, h *models.Host,
...
+	// count already existing masters
+	mastersCount := 0
+	if err := m.db.Model(&models.Host{}).Where("cluster_id = ? and status != ? and role = ?",
+		host.ClusterID, models.HostStatusDisabled, models.HostRoleMaster).Count(&mastersCount).Error; err != nil {
+		log.WithError(err).Errorf("failed to count masters in cluster %s", host.ClusterID.String())
+		return
+	}
+
+	if mastersCount < 3 {
+		host.Role = models.HostRoleMaster
```
avishayt, line 476: What's this for?

internal/host/host.go
filanov: we have a definition of bootstrap and unassigned so i was finking that it will be clearer that way.

internal/host/host.go
filanov: It's for the `newValidationContext` it will run the validations based on master role. but it will not store anything in the DB.
this validation will run only when there is a role defined:
```
conditions[HasCPUCoresForRole] && conditions[HasMemoryForRole]
```

internal/host/host.go
```diff
@@ -420,3 +425,67 @@ func (m *Manager) reportInstallationMetrics(ctx context.Context, h *models.Host,
...
+	var h *models.Host
+	ctx := requestid.ToContext(context.Background(), requestid.NewID())
+	for {
+		h = <-m.autoRoleSelectChan
+		m.autoRoleSelection(ctx, h)
+	}
+}
+
+func (m *Manager) autoRoleSelection(ctx context.Context, h *models.Host) {
+	var (
```
ori-amizur, line 443: You should check if disabled and quit if yes.

internal/host/host.go
```diff
@@ -420,3 +425,67 @@ func (m *Manager) reportInstallationMetrics(ctx context.Context, h *models.Host,
 		m.metricApi.ReportHostInstallationMetrics(log, cluster.OpenshiftVersion, h, previousProgress, CurrentStage)
 	}
 }
+func (m *Manager) AutoRoleSelection(h *models.Host) {
+	// using channel as a queue to prevent races between different hosts
+	go func() { m.autoRoleSelectChan <- h }()
```
ori-amizur, line 430: Why go routine?  Does it take so much time?  Or is it synchronization issue?

internal/host/host.go
```diff
@@ -420,3 +425,67 @@ func (m *Manager) reportInstallationMetrics(ctx context.Context, h *models.Host,
...
+		log.WithError(err).Errorf("failed to get host %s for auto role selection", h.ID.String())
+		return
+	}
+	// check if role already selected
+	if funk.ContainsString([]string{string(models.HostRoleMaster), string(models.HostRoleWorker)}, string(host.Role)) {
+		return
+	}
+
+	// in case of any error host will be set a worker
+	defer func() {
```
ori-amizur, line 459: Why deferred  update? Wouldn't it be better not to attempt to do anything in case of failure on the way?

internal/host/host.go
filanov: if the queue is full i don't want to block the api.

internal/host/host.go
filanov: The feature is auto role selection, it will be strange to select roles only to part of the hosts.
So it's better to select worker then not selecting anything, no? 
@avishayt what do you think?

avishayt: Agree, might as well choose worker

internal/bminventory/inventory.go
```diff
@@ -1405,6 +1405,7 @@ func handleReplyByType(params installer.PostStepReplyParams, b *bareMetalInvento
 	switch params.Reply.StepType {
 	case models.StepTypeInventory:
 		err = b.hostApi.UpdateInventory(ctx, &host, stepReply)
+		b.hostApi.AutoRoleSelection(&host)
```
ronniel1, line 1408: Doing this here means that we do not take into consideration the network connectivity.
We'll select master for hosts that are not connected.

PR #475, opened 2020-07-26
Allow to register an installed host

masayag:
Registering a host which is already installed can be done to the same
cluster or to a different one.

Without this PR, it is already possible to register a host to a different
cluster.

This PR enables the re-registering of the host to its cluster, in
which case the host status will transit from 'Installed' to
'Discovering' and from 'Discovering' to 'Known'.

Signed-off-by: Moti Asayag <masayag@redhat.com>

### DIFF
--- a/internal/host/statemachine.go
+++ b/internal/host/statemachine.go
@@ -27,6 +27,7 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
 			HostStatusDiscovering,
 			HostStatusKnown,
 			HostStatusDisconnected,
+			HostStatusInstalled,
 			HostStatusInsufficient,
 			HostStatusResetting,
 			stateswitch.State(models.HostStatusResettingPendingUserAction),
--- a/internal/host/transition_test.go
+++ b/internal/host/transition_test.go
@@ -149,10 +149,6 @@ var _ = Describe("RegisterHost", func() {
 				name:     "error",
 				srcState: HostStatusError,
 			},
-			{
-				name:     "installed",
-				srcState: HostStatusInstalled,
-			},
 		}
 
 		for i := range tests {


masayag:
@eranco74 How should registering of installed host to a different cluster be addressed ? It looks like a current bug that allows registering the same host to a different cluster, ending up with two entities representing the same host in the DB, however only one record describes its actual state and assignment.

I think registering the host to a different cluster should be supported, but in that case, either we delete the association to the previous cluster or update the state to indicate an error.

avishayt:
> @eranco74 How should registering of installed host to a different cluster be addressed ? It looks like a current bug that allows registering the same host to a different cluster, ending up with two entities representing the same host in the DB, however only one record describes its actual state and assignment.
> 
> I think registering the host to a different cluster should be supported, but in that case, either we delete the association to the previous cluster or update the state to indicate an error.

In case a host boots with an agent that points to a second cluster, it will no longer ping the first cluster and therefore go into "disconnected" state there

masayag:
> > @eranco74 How should registering of installed host to a different cluster be addressed ? It looks like a current bug that allows registering the same host to a different cluster, ending up with two entities representing the same host in the DB, however only one record describes its actual state and assignment.
> > I think registering the host to a different cluster should be supported, but in that case, either we delete the association to the previous cluster or update the state to indicate an error.
> 
> In case a host boots with an agent that points to a second cluster, it will no longer ping the first cluster and therefore go into "disconnected" state there

If this is the expected behavior, there is a bug, since the host remain as 'Installed' on origin cluster:
`test-infra-cluster-master-2` was register to the new cluster and appears in `Insufficient` as expected:
Selection_999(974)

However, on the origin cluster of it it is still in `Installed` status:
Selection_999(970)

eranco74:
> > @eranco74 How should registering of installed host to a different cluster be addressed ? It looks like a current bug that allows registering the same host to a different cluster, ending up with two entities representing the same host in the DB, however only one record describes its actual state and assignment.
> > I think registering the host to a different cluster should be supported, but in that case, either we delete the association to the previous cluster or update the state to indicate an error.
> 
> In case a host boots with an agent that points to a second cluster, it will no longer ping the first cluster and therefore go into "disconnected" state there

This is true as long the host wasn't installed, once the host reach installed status bm-inventory stop monitoring it.
And this is the case we have.

eranco74:
> Registering a host which is already installed can be done to the same
> cluster or to a different one.
> 
> Without this PR, it is already possible to register a host to a different
> cluster. However, in this situation there are two hosts with the same
> machine-uuid in the DB (hosts.connectivity.remote_hosts.host_id).
> 
> This PR enabled the re-registering of the same host to the cluster, in
> which case the host status will transit from 'Installed' to
> 'Discovering' and from 'Discovering' to 'Known'.
> 
> Signed-off-by: Moti Asayag [masayag@redhat.com](mailto:masayag@redhat.com)

I think that a host id is unique only as part of a cluster (host identifier is cluster_id + host_id)
So when a host register to 2 clusters it looks like 2 different hosts in bm-inventory.
To me knowledge hosts.connectivity.remote_hosts.host_id is cluster specific @ori-amizur @ronniel1  please correct me if I'm wrong.

avishayt:
> > Registering a host which is already installed can be done to the same
> > cluster or to a different one.
> > Without this PR, it is already possible to register a host to a different
> > cluster. However, in this situation there are two hosts with the same
> > machine-uuid in the DB (hosts.connectivity.remote_hosts.host_id).
> > This PR enabled the re-registering of the same host to the cluster, in
> > which case the host status will transit from 'Installed' to
> > 'Discovering' and from 'Discovering' to 'Known'.
> > Signed-off-by: Moti Asayag [masayag@redhat.com](mailto:masayag@redhat.com)
> 
> I think that a host id is unique only as part of a cluster (host identifier is cluster_id + host_id)
> So when a host register to 2 clusters it looks like 2 different hosts in bm-inventory.
> To me knowledge hosts.connectivity.remote_hosts.host_id is cluster specific @ori-amizur @ronniel1 please correct me if I'm wrong.

Correct

masayag:
Moved to https://github.com/openshift/assisted-service/pull/34

PR #476, opened 2020-07-26
MGMT-1318 - enhance BMH CRs with relevant inventory data

yevgeny-shnaidman:


eranco74:
Although bm-inentory is starting the kubeconfig-generate job so technically it can just pass the relevant info in the job env.

yevgeny-shnaidman:
i thought about it , but since we want to get all the hosts with all the data and parse, it is better to bring in the client, with defined structs - more clear code

internal/installcfg/installcfg.go
```diff
@@ -184,7 +184,7 @@ func setBMPlatformInstallconfig(log logrus.FieldLogger, cluster *common.Cluster,
 	}
 	cfg.Platform = platform{
 		Baremetal: baremetal{
-			ProvisioningNetworkInterface: "ethh0",
+			ProvisioningNetworkInterface: "ens4",
```
ronniel1, line 187: Will we always have this interface on the host?

yevgeny-shnaidman: for 4.5 it does not matter right now, since we don't expect BMO to run correctly. For 4.6 this is something we need to check:
1) currently if we add a second interface to the node, this is the second interface  that is created by RCHOS ( every time a  node boots with RCHOS)
2) in case we will have more then 2 interfaces on a node, then we need to provide a way for us(user) to define which interface to use.
3) the more troubling issue: how does user (us) knows how interfaces is named before the node is booted at least once with RCHOS

ronniel1: on physical hosts it is very likely that we will have more than 1 physical interfaces

yevgeny-shnaidman: MGMT-1621

PR #477, opened 2020-07-26
MGMT-1041 return forbidden error when host register to cluster that d…

filanov:
…oes not exists

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -1154,8 +1154,10 @@ func (b *bareMetalInventory) RegisterHost(ctx context.Context, params installer.
 
 	if err := b.db.First(&cluster, "id = ?", params.ClusterID.String()).Error; err != nil {
 		log.WithError(err).Errorf("failed to get cluster: %s", params.ClusterID.String())
-		return installer.NewRegisterHostBadRequest().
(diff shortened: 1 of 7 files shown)


avishayt:
Why not 404?

filanov:
> Why not 404?

The ticket (MGMT-1041) require to stop registration on 403 i was thinking that if the cluster don't exist anymore then maybe we should request hosts to stop register as well. 
Do you think i should return 404 and update the agent to stop registration on 404 as well?

filanov:
Agents PR https://github.com/ori-amizur/introspector/pull/31

internal/bminventory/inventory.go
```diff
@@ -1146,8 +1146,7 @@ func (b *bareMetalInventory) RegisterHost(ctx context.Context, params installer.
 
 	if err := b.db.First(&cluster, "id = ?", params.ClusterID.String()).Error; err != nil {
 		log.WithError(err).Errorf("failed to get cluster: %s", params.ClusterID.String())
-		return installer.NewRegisterHostBadRequest().
-			WithPayload(common.GenerateError(http.StatusBadRequest, err))
+		return common.NewApiError(http.StatusNotFound, err)
```
ori-amizur, line 1149: Need to check if this is really not found error, or something else.  Otherwise - it will cause agent to sleep forever

PR #478, opened 2020-07-26
MGMT-1268 Merge cluster and host monitors

gamli75:
@filanov @ronniel1 @ori-amizur please review


filanov:
I'm not sure it will help us
If we monitor the hosts first then one of then may get into error state then we need to set the cluster to error state (for example master failed installation) 
Then only in the next iteration we will fail all the host. 
The flow is not wrong just may be confusing.

PR #480, opened 2020-07-27
Suggestion change host name via state machine

filanov:
It's a suggestion how to use state machine to handle option that don't require state change - we still want to validate the source state, that will enable to run the transition for specific states. 
@ori-amizur you talked with me about how to run transition without changing the state it contain this part as well. 

The downside of this approach is that we need to wrote more code to use the state machine.

What do you think guys?


filanov:
> I guess for role update too? Fine with either way.

It's just a suggestion to how run transitions without changing the state. 
There are many APIs that depend on source states, i changed only one to see what everyone thing about it. 
Maybe someone will have a better idea or we may decide to leave it as is.

internal/host/statemachine.go
```diff
@@ -290,5 +291,18 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
 		})
 	}
 
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeChangeHostName,
```
ori-amizur, line 295: Why isn't it just update database like UpdateInventory? It doesn't do anything else anyway so no need to do it through statewitch

yevgeny-shnaidman:
Currently i don't see any big difference between the options. Although the code becomes quite complicated

yevgeny-shnaidman:
@filanov as discussed, maybe we should consider changing the way we run the requests: add a goroutine per cluster that handler all the requests to that cluster in synchronous way. All the APIs just send the requests to that routine and wait for the answer. Simplifies access for read/write to database, no more race cases,downsize: runtime, but the performance is not really an issue here

internal/host/statemachine.go
filanov: You are the one that suggested to do everything via the state machine. 
And It's not just update the db, it update the DB in case the source state is part of the list.

ori-amizur: Maybe I was wrong.  (We are just discussing)
I am not sure about this one because no state change is involved here, and even no condition.
In UpdateInventory and UpdateRole there are allowed states validation.  Actually they do the same thing.
Do you think we should take the same approach for those APIs?

filanov: I'm not sure. because we need to add a lot of code.
This is why i opened it to discussion.

ori-amizur: Whatever approach we follow, it has to be consistent with other similar operations

filanov:
Looks like we did not manage to decide - closing this PR.

internal/host/statemachine.go
ronniel1: Maybe all we should do is create, in this is something along these lines:
```
HostNameChangeValidStates := []string{HostStatusDiscovering, HostStatusKnown, HostStatusDisconnected, HostStatusInsufficient, HostStatusPendingForInput}	
func IsValidStateFor(validStates []string) bool {
   if func,Contains(
......
}
```

This way we keep all the states lists in this file

filanov: Or have a list of validation functions like:
```
isValidStateForHostNameChange := func(state sting) bool {
return funk.ContainsString([]string{...}, state)
}
```
Then as @ronniel1  suggested if we have all the lists in the same file it will be easier to see all the affected apis when we add a new state or change something.

PR #481, opened 2020-07-27
Updated use of Postgres/S3 credentials from Secrets.

ybettan:
This change is done in order to be as close as possible to the
integration environment which deploy Postgres and S3 and corresponding
secrets which contain the credentials to connect to them.

Now the credentials for connecting to DB comes from
`assisted-installer/assisted-installer-rds` secret.

The credentials for connecting to S3 comes from
`assisted-installer/assisted-installer-s3` secret.

Also grouped all DB config fields into a separate struct.

Signed-off-by: Yoni Bettan <ybettan@redhat.com>)


filanov:
This feature need to support our test infra as well.
As you can see none of the subsystem tests passed. 
Maybe need to add an option to deploy secrets for test infra.

ybettan:
[WIP]

ybettan:
There are other changes required, among them we need to create the bucket outside `bm-inventory`'s code because in integration env S3 bucket already exist and re-creating it will generate in an "Access denied" error from S3.
@filanov

filanov:
> There are other changes required, among them we need to create the bucket outside `bm-inventory`'s code because in integration env S3 bucket already exist and re-creating it will generate in an "Access denied" error from S3.
> @filanov

Are you sure? 
I think that if the bucket already exists then the api will return 200.
Otherwise every reboot of the service will fail.

PR #482, opened 2020-07-27
MGMT-1129 Machine CIDR calculation and VIP verification should exclude disabled hosts

ori-amizur:


ori-amizur:
> No tests changes?

Added

PR #483, opened 2020-07-27
MGMT-1632 Subsystem: fixed "cancel_failed_cluster" test

RazRegev:
"cancel_failed_cluster" test is starting installation, then failing it, then
trying to abort the installation. 
Since host monitor is now setting all hosts to error in case cluster is in
error, on some cases it get race condition when subsystem and monitor try to
set the host state at the same time. I fixed it by canceling the installation
only after hosts monitor set the hosts states to error.

### DIFF
--- a/subsystem/cluster_test.go
+++ b/subsystem/cluster_test.go
@@ -796,8 +796,6 @@ var _ = Describe("cluster install", func() {
 				stateInfo := fmt.Sprintf(clusterErrorInfo, clusterID.String())
 				waitForClusterState(ctx, clusterID, models.ClusterStatusError, defaultWaitForClusterStateTimeout,
 					stateInfo)
-				_, err := bmclient.Installer.CancelInstallation(ctx, &installer.CancelInstallationParams{ClusterID: clusterID})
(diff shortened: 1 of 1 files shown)


subsystem/cluster_test.go
```diff
@@ -807,8 +805,11 @@ var _ = Describe("cluster install", func() {
 				c = rep.GetPayload()
 				Expect(swag.StringValue(c.Status)).Should(Equal(models.ClusterStatusError))
 				for _, host := range c.Hosts {
-					Expect(swag.StringValue(host.Status)).Should(Equal(models.HostStatusError))
+					waitForHostState(ctx, clusterID, *host.ID, models.HostStatusError,
+						defaultWaitForHostStateTimeout)
 				}
+				_, err = bmclient.Installer.CancelInstallation(ctx, &installer.CancelInstallationParams{ClusterID: clusterID})
+				Expect(err).ShouldNot(HaveOccurred())
```
ori-amizur, line 812: Why isn't there verification that cluster and hosts are in expected state after cancel?

PR #487, opened 2020-07-28
MGMT-1616 - move OCP release image configuration to configmap

yevgeny-shnaidman:


internal/bminventory/inventory.go
```diff
@@ -1589,7 +1588,7 @@ func (b *bareMetalInventory) createKubeconfigJob(cluster *common.Cluster, jobNam
 								},
 								{
 									Name:  "OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE",
-									Value: overrideImageName, //TODO: change this to match the cluster openshift version
+									Value: b.ReleaseImage, //TODO: change this to match the cluster openshift version
```
romfreiman, line 1591: what about current hardcoded place?

yevgeny-shnaidman: what do you mean? i removed it

romfreiman: oops :)

deploy/bm-inventory-configmap.yaml
```diff
@@ -9,4 +9,5 @@ data:
   INVENTORY_URL: REPLACE_URL
   INVENTORY_PORT: REPLACE_PORT
   NAMESPACE: assisted-installer
-  BASE_DNS_DOMAINS: REPLACE_DOMAINS # example: name1:id1/provider1,name2:id2/provider2
\ No newline at end of file
+  BASE_DNS_DOMAINS: REPLACE_DOMAINS # example: name1:id1/provider1,name2:id2/provider2
+  OPENSHIFT_INSTALL_RELEASE_IMAGE: "quay.io/openshift-release-dev/ocp-release@sha256:eab93b4591699a5a4ff50ad3517892653f04fb840127895bb3609b3cc68f98f3"
```
romfreiman, line 13: can u please add a comment here saying it's 4.5.3 GA?

internal/bminventory/inventory.go
```diff
@@ -69,22 +69,24 @@ var (
...
-	UseK8s              bool              `envconfig:"USE_K8S" default:"true"` // TODO remove when jobs running deprecated
-	BaseDNSDomains      map[string]string `envconfig:"BASE_DNS_DOMAINS" default:""`
-	JobCPULimit         string            `envconfig:"JOB_CPU_LIMIT" default:"500m"`
-	JobMemoryLimit      string            `envconfig:"JOB_MEMORY_LIMIT" default:"1000Mi"`
-	JobCPURequests      string            `envconfig:"JOB_CPU_REQUESTS" default:"300m"`
-	JobMemoryRequests   string            `envconfig:"JOB_MEMORY_REQUESTS" default:"400Mi"`
+	ImageBuilder        string `envconfig:"IMAGE_BUILDER" default:"quay.io/ocpmetal/installer-image-build:latest"`
+	AgentDockerImg      string `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/ocpmetal/agent:latest"`
+	KubeconfigGenerator string `envconfig:"KUBECONFIG_GENERATE_IMAGE" default:"quay.io/ocpmetal/ignition-manifests-and-kubeconfig-generate:latest"` // TODO: update the latest once the repository has git workflow
+	//[TODO] -  change the default of Releae image to "", once everyine wll update their environment
```
eranco74, line 75: Why the "TODO" we can keep the default

internal/bminventory/inventory.go
eranco74: https://github.com/filanov/bm-inventory/pull/487/files#diff-97c2fe56dacc4fee9ab38e4dbd281ab8L1540 was the hardcoded place.

internal/bminventory/inventory.go
yevgeny-shnaidman: i added the default to be backwards compatible until people will update their environment. Otherwise, we want default to be empty, so we fail in case image is not defined

yevgeny-shnaidman: i think that what we once wanted ( get the release image based on openshift version) is not really applicable. user may decide which release he wants, and configure it in the configmap. OPENSHIFT_VER will be used for RCHOS image, and probably for RHCOS Live CD image, since we want them to be compatible

PR #492, opened 2020-07-28
Add /vendor to .gitignore

carbonin:
If this isn't meant to be committed it would be nice to not see it
in `git status`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ __pycache__/
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+
+/vendor


ronniel1:
Why do we have vendor directory at all?
We are not using vendoring

carbonin:
When I'm working on a go module I usually pull in the dependencies so I can jump to the definition of functions in dependencies.
This would explicitly prevent folks from committing the vendor directory.

carbonin:
Closing in favor of https://github.com/filanov/bm-inventory/pull/516

PR #495, opened 2020-07-28
MGMT-1609 Container image migration from Ubuntu based OS to Fedora

jparrill:
**Additions**:

- Collateral work comming from MGMT-1609 in order to get the image working on openshift/release platform
- Adding Podman support on make execution using `RT_CMD=podman make unit-test`


Dockerfile.bm-inventory-build
```diff
@@ -1,9 +1,14 @@
...
 
-ENV GO111MODULE=on
+ENV GO111MODULE=on \
+    GOPATH=${HOME}/go \
+    GOBIN=${HOME}/go/bin \
+    GOPROXY=https://proxy.golang.org
+ENV PATH="${GOBIN}:${PATH}" \
+    GOCACHE="${GOPATH}/pkg/mod"
+
+RUN microdnf clean all && microdnf install go docker podman libvirt-client awscli python3-pip postgresql make -y
```
romfreiman, line 10: microdnf clean all shoudnt be at the end?

Makefile
```diff
@@ -17,6 +17,7 @@ endef # get_service
 endif # TARGET
 
 SERVICE := $(or ${SERVICE},quay.io/ocpmetal/bm-inventory:latest)
+RT_CMD := $(or ${RT_CMD},podman --storage-driver=vfs --cgroup-manager=cgroupfs)
```
romfreiman, line 20: podman is the default?

jparrill: yep, it's not mandatory of course but I guess at least thinking that is the best on long term.

jparrill: Ok, setting up docker as default as we discussed.

Dockerfile.bm-inventory-build
jparrill: Both are important I think :+1:

jparrill:
Ready to be reviewed

jparrill:
Don't merge it yet, I still need to modify the Dockerfile in order to meet the openshift/release needs and this involves to use UBI or CentOS as BaseOS on FROM field. For both I need to validate the right repositories

Dockerfile.bm-inventory-build
```diff
@@ -1,15 +1,35 @@
-FROM golang:1.14.3
+FROM centos:7
```
djzager, line 1: My contention would be that ubi (and ubi8 for that matter) would be the best choice to use upstream as it would most closely resemble d/s builds.

```suggestion
FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```

djzager: But aren't you able to override the `FROM` for the purposes of prow builds?

PR #496, opened 2020-07-28
OPCBUGSM-4299: Allow canceling installed hosts

RazRegev:
OPCBUGSM-4299: Allow canceling installed hosts

User may try to cancel an installation where some hosts are already done.
This operation should be permitted.
When it happens, first the hosts states will be changed to error, then after
reset the cluster monitor will set them to resetting-pending-user-action.

### DIFF
--- a/internal/host/statemachine.go
+++ b/internal/host/statemachine.go
@@ -77,8 +77,13 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
 
 	// Cancel installation
 	sm.AddTransition(stateswitch.TransitionRule{
-		TransitionType:   TransitionTypeCancelInstallation,
-		SourceStates:     []stateswitch.State{HostStatusInstalling, HostStatusInstallingInProgress, HostStatusError},
+		TransitionType: TransitionTypeCancelInstallation,
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusInstalling),
+			stateswitch.State(models.HostStatusInstallingInProgress),
+			stateswitch.State(models.HostStatusInstalled),
+			stateswitch.State(models.HostStatusError),
+		},
 		DestinationState: HostStatusError,
 		PostTransition:   th.PostCancelInstallation,
 	})
--- a/internal/host/transition_test.go
+++ b/internal/host/transition_test.go
@@ -308,6 +308,78 @@ var _ = Describe("HostInstallationFailed", func() {
 	})
 })
 
+var _ = Describe("Cancel host installation", func() {
+	var (
+		ctx               = context.Background()
+		dbName            = "cancel_host_installation_test"
+		hapi              API
+		db                *gorm.DB
+		hostId, clusterId strfmt.UUID
+		host              models.Host
+		ctrl              *gomock.Controller
+		mockEventsHandler *events.MockHandler
+	)
+
+	BeforeEach(func() {
+		db = common.PrepareTestDB(dbName, &events.Event{})
+		ctrl = gomock.NewController(GinkgoT())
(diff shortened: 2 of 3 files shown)


internal/host/transition_test.go
```diff
@@ -346,19 +416,23 @@ var _ = Describe("Reset host", func() {
 		{state: models.HostStatusDisconnected, success: false, statusCode: http.StatusConflict},
 	}
 
-	acceptResetEvent := func(times int) {
+	acceptNewEvents := func(times int) {
 		mockEventsHandler.EXPECT().AddEvent(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(2 * times)
```
filanov, line 420: why twice times?

internal/host/transition_test.go
```diff
@@ -308,6 +308,76 @@ var _ = Describe("HostInstallationFailed", func() {
...
+		{state: models.HostStatusDisconnected, success: false, statusCode: http.StatusConflict},
+	}
+
+	acceptNewEvents := func(times int) {
+		mockEventsHandler.EXPECT().AddEvent(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(times)
+	}
+
+	It("reset_host_cases", func() {
+		for _, t := range tests {
+			By(fmt.Sprintf("cancel from state %s", t.state))
```
filanov, line 355: Why not do the for on `It`? 
then you can put the name inside It

internal/host/transition_test.go
avishayt: One for the event about cancelling, one for host status change

internal/host/statemachine.go
```diff
@@ -77,8 +77,13 @@ func NewHostStateMachine(th *transitionHandler) stateswitch.StateMachine {
 
 	// Cancel installation
 	sm.AddTransition(stateswitch.TransitionRule{
-		TransitionType:   TransitionTypeCancelInstallation,
```
eranco74, line 80: Consider adding new host state "Canceled" to be the destination state of this operation.
Moving hosts to error make it seem that something failed.

RazRegev:
> > User may try to cancel an installation where some hosts are already done.
> > This operation should be permitted.
> > When it happens, first the hosts states will be changed to error, then the
> > cluster monitor will set them to error, then on reset to
> > resetting-pending-user-action.
> 
> If the hosts are already in error why would the cluster monitor move them to error again?

ops.. it happens only once, I wrote it twice accidentally. will fix that

internal/host/statemachine.go
RazRegev: @avishayt ^^

RazRegev: except the status info that says the operation was cancelled.

internal/host/transition_test.go
filanov: it's not clear form the code. maybe you should add a comment about it

internal/host/transition_test.go
```diff
@@ -308,6 +308,75 @@ var _ = Describe("HostInstallationFailed", func() {
...
+			host.Status = swag.String(t.state)
+			Expect(db.Create(&host).Error).ShouldNot(HaveOccurred())
+			eventsNum := 1
+			if t.success {
+				eventsNum++
+			}
+			acceptNewEvents(eventsNum)
+			err := hapi.CancelInstallation(ctx, &host, "reason", db)
+			if t.success {
+				Expect(err).ShouldNot(HaveOccurred())
```
filanov, line 367: how about validating target state? 
and that in case of failure nothing will change.

RazRegev: done

PR #497, opened 2020-07-28
MGMT-1582 Cancel/Reset from preparing-for-installation

RazRegev:
Allow canceling or resetting installation from preparing-for-installation
state. Added subsystem-test.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -856,7 +856,10 @@ func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, c
 		log.WithError(err).Errorf("failed to get install config for cluster %s", cluster.ID)
 		return errors.Wrapf(err, "failed to get install config for cluster %s", cluster.ID)
 	}
-	jobName := fmt.Sprintf("%s-%s-%s", kubeconfigPrefix, cluster.ID.String(), uuid.New().String())[:63]
+
+	ctime := time.Time(cluster.CreatedAt)
+	cTimestamp := strconv.FormatInt(ctime.Unix(), 10)
+	jobName := fmt.Sprintf("%s-%s-%s", kubeconfigPrefix, cluster.ID.String(), cTimestamp)[:63]
 	if err := b.job.Create(ctx, b.createKubeconfigJob(&cluster, jobName, cfg)); err != nil {
 		log.WithError(err).Errorf("Failed to create kubeconfig generation job %s for cluster %s", jobName, cluster.ID)
 		return errors.Wrapf(err, "Failed to create kubeconfig generation job %s for cluster %s", jobName, cluster.ID)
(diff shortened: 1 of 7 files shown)


internal/bminventory/inventory.go
```diff
@@ -855,7 +855,10 @@ func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, c
 		log.WithError(err).Errorf("failed to get install config for cluster %s", cluster.ID)
 		return errors.Wrapf(err, "failed to get install config for cluster %s", cluster.ID)
 	}
-	jobName := fmt.Sprintf("%s-%s-%s", kubeconfigPrefix, cluster.ID.String(), uuid.New().String())[:63]
+
+	ctime := time.Time(cluster.CreatedAt)
```
filanov, line 859: Why do you need this?

internal/cluster/transition_test.go
```diff
@@ -144,6 +144,73 @@ var _ = Describe("Transition tests", func() {
...
+		{state: models.ClusterStatusFinalizing, success: false, statusCode: http.StatusConflict},
+		{state: models.ClusterStatusInstalled, success: false, statusCode: http.StatusConflict},
+	}
+
+	It("cancel_cluster_installation_cases", func() {
+		acceptNewEvents(len(tests))
+		acceptClusterInstallationFinished(len(tests))
+
+		for _, t := range tests {
+			By(fmt.Sprintf("cancel from state %s", t.state))
```
filanov, line 193: i already commented in this one, but i think that each test case in the table should be a different test and not a case inside the test.
The way it is right now you will get only the first fail. then you will need to run it again to see other fails

internal/bminventory/inventory.go
RazRegev: So I could delete this job by its name in ResetCluster (in case hosts are preparing-for-installation).
The intention was to generate the unique name using const cluster params instead of random uuid.

internal/cluster/transition_test.go
RazRegev: will fix

internal/bminventory/inventory.go
```diff
@@ -1986,6 +1989,15 @@ func (b *bareMetalInventory) ResetCluster(ctx context.Context, params installer.
 	if err := b.clusterApi.ResetCluster(ctx, &c, "cluster was reset by user", tx); err != nil {
 		return common.GenerateErrorResponder(err)
 	}
+
+	// abort installation files generation job if running.
+	ctime := time.Time(c.CreatedAt)
+	cTimestamp := strconv.FormatInt(ctime.Unix(), 10)
+	jobName := fmt.Sprintf("%s-%s-%s", kubeconfigPrefix, c.ID.String(), cTimestamp)[:63]
+	if err := b.job.Delete(ctx, jobName, b.Namespace); err != nil {
```
filanov, line 1997: did you check what will happen if this job don't exist?

RazRegev: yes, nothing will happen, it wont cause errors

internal/cluster/transition_test.go
```diff
@@ -144,6 +144,75 @@ var _ = Describe("Transition tests", func() {
...
+			if t.success {
+				Expect(err).ShouldNot(HaveOccurred())
+			} else {
+				Expect(err).Should(HaveOccurred())
+				Expect(err.StatusCode()).Should(Equal(t.statusCode))
+			}
+		})
+	}
+
+	AfterEach(func() {
```
filanov, line 211: You don't validate the mocks if you don't call for `ctrl.Finish()`

internal/cluster/transition_test.go
```diff
@@ -170,34 +239,36 @@ var _ = Describe("Reset cluster", func() {
...
 			} else {
 				Expect(err).Should(HaveOccurred())
 				Expect(err.StatusCode()).Should(Equal(t.statusCode))
 			}
-		}
-	})
+		})
+	}
 
 	AfterEach(func() {
```
filanov, line 273: You don't validate the mocks if you don't call for `ctrl.Finish()`

PR #501, opened 2020-07-29
OCPBUGSM-10683 Deploy UI: pull image

RazRegev:
Try to pull ui image locally before using it to create deployment configuration.

### DIFF
--- a/tools/deploy_ui.py
+++ b/tools/deploy_ui.py
@@ -14,7 +14,9 @@ def main():
 
     dst_file = os.path.join(os.getcwd(), "build/deploy_ui.yaml")
     image_fqdn = deployment_options.get_image_override(deploy_options, "ocp-metal-ui", "UI_IMAGE")
-    cmd = '{command} run {image} /deploy/deploy_config.sh -i {image} -n {namespace}'.format(command=utils.get_runtime_command(), image=image_fqdn, namespace=deploy_options.namespace)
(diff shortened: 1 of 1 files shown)


tools/deploy_ui.py
```diff
@@ -14,7 +14,9 @@ def main():
 
     dst_file = os.path.join(os.getcwd(), "build/deploy_ui.yaml")
     image_fqdn = deployment_options.get_image_override(deploy_options, "ocp-metal-ui", "UI_IMAGE")
-    cmd = '{command} run {image} /deploy/deploy_config.sh -i {image} -n {namespace}'.format(command=utils.get_runtime_command(), image=image_fqdn, namespace=deploy_options.namespace)
+    runtime_cmd = utils.get_runtime_command()
+    utils.check_output(f'{runtime_cmd} pull {image_fqdn}')
```
filanov, line 18: Does podman pull the image if it's already exist?

RazRegev: when the image is updated yes

filanov: I think that we had an issue with it and podman, @tsorya do you remember something like this?

tsorya: nor podman nor docker pulls image if it already exists. 
U can set flag the will do it

tsorya: not relevant comment : this is pull command, it will always try to pull

PR #502, opened 2020-07-29
Updated S3 bucket creation in cmd/main.

ybettan:
Bucket creation will be defaulted to "false" and created only on demand.
In integration env we don't have permissions to create buckets so we
can't always create it.

Signed-off-by: Yoni Bettan <ybettan@redhat.com>

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -63,6 +63,7 @@ var Options struct {
 	HostStateMonitorInterval    time.Duration `envconfig:"HOST_MONITOR_INTERVAL" default:"8s"`
 	Versions                    versions.Versions
 	UseK8s                      bool          `envconfig:"USE_K8S" default:"true"` // TODO remove when jobs running deprecated
+	CreateS3Bucket              bool          `envconfig:"CREATE_S3_BUCKET" default:"false"`
 	ImageExpirationInterval     time.Duration `envconfig:"IMAGE_EXPIRATION_INTERVAL" default:"30m"`
 	ImageExpirationTime         time.Duration `envconfig:"IMAGE_EXPIRATION_TIME" default:"60m"`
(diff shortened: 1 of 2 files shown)


cmd/main.go
```diff
@@ -63,6 +63,7 @@ var Options struct {
 	HostStateMonitorInterval    time.Duration `envconfig:"HOST_MONITOR_INTERVAL" default:"8s"`
 	Versions                    versions.Versions
 	UseK8s                      bool          `envconfig:"USE_K8S" default:"true"` // TODO remove when jobs running deprecated
+	CreateS3Bucket              bool          `envconfig:"CREATE_S3_BUCKET" default:"false"`
```
filanov, line 66: Maybe the other way around? it should be true and there will be no change in the configmap?
@avishayt what do you think?

ybettan: There are 2 reasons I chose false by default:
1. Logically I assume that in dev env, S3 is given to us and the bucket is created by someone else therefore I don't think creating a bucket by default is the correct way.
2. If we set it to "true" by default then we will need to update https://github.com/openshift/assisted-service/blob/3e9c14b89725d1d3ce441dd530cecbd4b5f90b97/openshift/template.yaml#L46 so we don't actually gain something and one of the deployments will need to be updated anyway.

We have more control on dev env rather than integration/production env so I think it better to put the configurable parameter in dev env.

What do you think?  @filanov

filanov: Our dev environment is minikube at least when we write code and test before we open a pull request.
And there we need to create the bucket.
In production we don't control the bucket. So to support it you will need to add an option to set the parameter in the config map (can be done in the differnet PR) 
Anyway, we can start like this and change it later if needed.

PR #503, opened 2020-07-29
MGMT-1428 Add TLS option for OC deployment

empovit:


deploy/assisted-installer-ingress-tls.yaml
```diff
@@ -0,0 +1,21 @@
+kind: Ingress
```
filanov, line 1: We need a different ingress for TLS? @romfreiman

empovit: No, it's a different template. Only one can be applied. The trade-off is between choosing from two templates and adding a block of YAML to the template in a Python script.

tools/deploy_inventory_service.py
```diff
@@ -1,13 +1,16 @@
...
+import deploy_tls_secret
 import deployment_options
+import utils
 
 
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument("--target")
     parser.add_argument("--domain")
+    parser.add_argument('--enable-tls', type=lambda s: s.lower() in ['true', 'yes'], default=False)
```
RazRegev, line 13:
```suggestion
    parser.add_argument('--enable-tls', action='store_true', default=False)
```

tools/deploy_inventory_service.py
```diff
@@ -21,19 +24,35 @@ def main():
...
-        utils.apply(dst_file)
+        hostname = utils.get_service_host("assisted-installer", deploy_options.target, deploy_options.domain,
+                                          deploy_options.namespace)
+
+        if deploy_options.enable_tls:
+            print("WARNING: To change TLS redirection behavior update "
+                  "spec/tls/insecureEdgeTerminationPolicy (None|Allow|Redirect) "
+                  "in the corresponding OpenShift route")
+            deploy_tls_secret.generate_secret(output_dir=os.path.join(os.getcwd(), "build"),
+                                              service="bm-inventory", san=hostname, namespace=deploy_options.namespace)
```
romfreiman, line 37: what will be the value of san in case of OCP?

tools/deploy_inventory_service.py
empovit: I considered this, but will be trickier to implement in scripting (makefile). What do you think?

tools/deploy_inventory_service.py
empovit: The hostname :) Including the domain, etc. I must use the same value that appears in "host" of the ingress, see REPLACE_HOSTNAME.

romfreiman: but in PSI the certificate should be on the ingress path ("http://assisted.apps.ocp.prod.psi.redhat.com").

tools/deploy_inventory_service.py
RazRegev: Yes, this is my suggestion :)
`ifneq ($(ENABLE_TLS),)
ENABLE_TLS_FLAG := --enable-tls
endif`

tools/deploy_inventory_service.py
empovit: The ingress in PSI responds to http://assisted-installer.<domain>, as well as the route. Staging too. Have a look, curl to https://assisted-installer.staging.apps.ocp.prod.psi.redhat.com/api/assisted-install/v1/component_versions.

empovit: If I understand correctly, it can't be different because ingress is deployed using this script. This logic hasn't changed, I'm just reusing it to populate an additional parameter.

romfreiman: right. Im just asking what hostname will be? Is it gonna be the ingress path?

empovit: Yes, it has to be. Moreover, the controller(?) is smart enough to verify that the host and the SAN match.

tools/deploy_tls_secret.py
```diff
@@ -0,0 +1,75 @@
...
+        os.remove(server_cert_path)
+        os.remove(server_key_path)
+        os.remove(ext_file)
+
+    return secret_name
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--service")
```
filanov, line 64: should it be in in the deploy options as all the other flags?

empovit: I agree, but I see many other options that are outside, such as `--target`, `--domain`. Target seems global, but not included in `deployment_options`. Is there a special reason, or just overlooked?

filanov: @eranco74 ?

empovit:
I'm re-submitting this PR as a single clean commit.

PR #516, opened 2020-07-29
Add /vendor to .gitignore

carbonin:
If this isn't meant to be committed it would be nice to not see it
in `git status`.

Replaces #492 

cc @ronniel1


ronniel1:
Why do you need to `go mod vendor`?

carbonin:
I guess I don't *need* to, but it's a habit and I just noticed that it caused a dirty git status.

PR #493, opened 2020-07-28
Module cleanup

carbonin:
These files were changed after running `go mod vendor` and
`go mod tidy`.

I'm not sure if this is something you want, but feel free to close if not.

PR #517, opened 2020-07-29
Module cleanup

carbonin:
These files were changed after running `go mod vendor` and
`go mod tidy`.

Replaces #493


ronniel1:
Just curios, why do you need to run `go mod vendor`?

PR #518, opened 2020-07-29
OCPBUGSM-11342 Support SSD drives

ori-amizur:


internal/hardware/validator.go
```diff
@@ -59,9 +58,8 @@ func gbToBytes(gb int64) int64 {
 
 func ListValidDisks(inventory *models.Inventory, minSizeRequiredInBytes int64) []*models.Disk {
 	var disks []*models.Disk
-	filter, _ := regexp.Compile(diskNameFilterRegex)
 	for _, disk := range inventory.Disks {
-		if disk.SizeBytes >= minSizeRequiredInBytes && disk.DriveType == "HDD" && !filter.MatchString(disk.Name) {
+		if disk.SizeBytes >= minSizeRequiredInBytes && funk.ContainsString([]string{"HDD", "SSD"}, disk.DriveType) {
```
ronniel1, line 62: @yevgeny-shnaidman will this always work?

tsorya: @ori-amizur we talked yesterday with @yevgeny-shnaidman .
we still need filter for nvme but we want them to be the last in the list to use them only if there is no others.
Don't delete the filter please.

tsorya: do something like this in sort 
`sort.Slice(disks, func(i, j int) bool {
		if filter.MatchString(disks[i].Name) && !filter.MatchString(disks[j].Name) {
			return false
		}
		if !filter.MatchString(disks[i].Name) && filter.MatchString(disks[j].Name) {
			return true
		}
		return disks[i].SizeBytes < disks[j].SizeBytes
	})`

yevgeny-shnaidman: this will allows us to use any HDD, SDD(including NVME) and RAID devices.( which will probably be reported as HDD). The question is: in case we have an nvme and some other disk, would we like to keep nvme for data? And install on NVME only in case all the disks are NVME? IF we do then @tsorya has a slightly different version of this code that can work

tsorya: @ronniel1 @avishayt what do u think? we want them for data  or not?

ori-amizur: Will re-add nvme filter

ori-amizur: The purpose of this function to indicate if disks are valid for our use.  If nvme s valid then it should be listed, otherwise not.
If we would like to prioritize disks, I don't think it is the purpose of this function.  I would recommend a function "selectDiskForInstall" instead.

ori-amizur: Anyway, there is no point for making nvme special.  We need to prioritize HDD over SSD for our case.

tsorya: not really, there are ssds that are not nmve

tsorya: so we need nvme too

ori-amizur: @tsorya We decided that we want SSDs last.  Do we want to distinguish non-nvme SSDs from nvme SSDs? @yevgeny-shnaidman ?

yevgeny-shnaidman: @ori-amizur yes, the decision was that we use NMVE only in case there are no other disks available

ori-amizur: Done

PR #519, opened 2020-07-30
MGMT-1658 enable disabling host in pending-for-input state

filanov:


filanov:
> LTGM except for the resetting -> disabled

You are right, fixed

PR #523, opened 2020-07-30
MGMT-1428 Add TLS option for OC deployment

empovit:


empovit:
Instead of https://github.com/filanov/bm-inventory/pull/503

tools/deploy_tls_secret.py
```diff
@@ -0,0 +1,75 @@
...
+    parser.add_argument("--tls-san")
+    parser.add_argument("--tls-expiration", help="Server certificate expiration (days)", type=int, default=120)
+    deploy_options = deployment_options.load_deployment_options(parser)
+
+    output_dir = os.path.join(os.getcwd(), "build")
+    generate_secret(output_dir=output_dir, service=deploy_options.service, san=deploy_options.tls_san,
+                    namespace=deploy_options.namespace, expiration=deploy_options.tls_expiration, keep_files=False)
+
+
+if __name__ == "__main__":
```
filanov, line 74: where is it used? I see that you are using `generate_secret` in other file but where is `main` used?

deploy/assisted-installer-ingress-tls.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - hosts:
+        - REPLACE_HOSTNAME
+      secretName: bm-inventory-tls
+  rules:
+    - host: REPLACE_HOSTNAME
+      http:
+        paths:
+          - backend:
+              serviceName: bm-inventory
+              servicePort: 8090
```
YuviGold, line 21: I see it is hardcoded in a lot of places. I suggest it should be a variable.

tools/deploy_inventory_service.py
```diff
@@ -21,19 +24,35 @@ def main():
...
-                data = data.replace("REPLACE_HOSTNAME",
-                                    utils.get_service_host("assisted-installer", deploy_options.target, deploy_options.domain, deploy_options.namespace))
-                print("Deploying {}".format(dst_file))
-                dst.write(data)
-        utils.apply(dst_file)
+        hostname = utils.get_service_host("assisted-installer", deploy_options.target, deploy_options.domain,
+                                          deploy_options.namespace)
+
+        if deploy_options.enable_tls:
+            print("WARNING: To change TLS redirection behavior update "
```
ronniel1, line 33: Is this something that needs to be done manually?

tools/deploy_tls_secret.py
```diff
@@ -0,0 +1,75 @@
...
+    with open(ext_file, "w") as f:
+        f.write(f'subjectAltName=DNS:{san}')
+
+    print(utils.check_output(f'openssl x509 -req -days {expiration} '
+                             f'-extfile "{ext_file}" '
+                             f'-CAcreateserial -CA "{ca_csr_path}" -CAkey "{ca_key_path}" '
+                             f'-in "{server_csr_path}" -outform PEM -out "{server_cert_path}"'))
+
+    secret_name = f'{service}-tls'
+    print(utils.check_output('cat <<EOF | kubectl apply -f -\n'
```
ronniel1, line 41: Maybe use utils.py apply() ?

YuviGold, line 41: use `testwrap.dedent` for clarity.
e.g.
```python
x = textwrap.dedent(f"""
apiVersion: v1
kind: Secret
metdata:
    name: {secret_name}
""")

```

tools/deploy_tls_secret.py
```diff
@@ -0,0 +1,75 @@
...
+                             f'    name: {secret_name}\n'
+                             f'    namespace: {namespace}\n'
+                             'type: kubernetes.io/tls\n'
+                             'data:\n'
+                             f'    tls.crt: $(cat {server_cert_path} | base64 -w 0)\n'
+                             f'    tls.key: $(cat {server_key_path} | base64 -w 0)\n'
+                             'EOF'))
+
+    if not keep_files:
+        os.remove(server_csr_path)
```
YuviGold, line 54:
```python
for file_name in [server_csr_path, server_cert_path, server_key_path, ext_file]:
    os.remove(file_name)
```

tools/deploy_tls_secret.py
```diff
@@ -0,0 +1,75 @@
...
+                             'metadata:\n'
+                             f'    name: {secret_name}\n'
+                             f'    namespace: {namespace}\n'
+                             'type: kubernetes.io/tls\n'
+                             'data:\n'
+                             f'    tls.crt: $(cat {server_cert_path} | base64 -w 0)\n'
+                             f'    tls.key: $(cat {server_key_path} | base64 -w 0)\n'
+                             'EOF'))
+
+    if not keep_files:
```
ronniel1, line 53: Why don't we keep them by default in the build directory?
Will be easier to debug.
Also will they be generated differently for each deployment?
Don't wee need to reuse them when deploying another version?

tools/deploy_tls_secret.py
empovit: It can be run manually, kind of self-contained.

filanov: Do you intend to add this option?
It's confusing to dead code.

tools/deploy_inventory_service.py
empovit: Unfortunately yes, unless we want to add route configuration/deployment. On the other hand, this won't be needed when we transition to either HTTPS or HTTP (meaning HTTP won't be exposed alongside HTTPS). OpenShift's routes ignores the redirect annotations of ingress controllers, and doesn't have an annotation of its own. See https://docs.openshift.com/container-platform/3.6/architecture/networking/routes.html#secured-routes

tools/deploy_tls_secret.py
empovit: I don't think it will work with heredoc (cat <<EOF ...), so I will need to create a temporary file and then delete it.

tools/deploy_tls_secret.py
empovit: Do you mean commit to the GitHub repo? I don't want to encourage bad practices with sensitive data. If you ask me, I would store the CA key/certificat somewhere secure and pull them for build/deploy. Will also make it easier to establish trust - the developers will need to import the CA certificate once, and everything signed by it will be trusted.

deploy/assisted-installer-ingress-tls.yaml
empovit: You're right, noted.

tools/deploy_tls_secret.py
empovit: Thanks! Was looking for it :)

tools/deploy_tls_secret.py
empovit: :+1:

PR #531, opened 2020-08-02
    Authorization support with AMS

danielerez:
Introducing authorization support with AMS using OCM client:

* Added OCMAuthorization interface to communticate with AMS
  - AccessReview: Review an account's access to perform an action on a particular resource or resource type.
    We'll use it to ensure that the user has authorization to use the installer service.
  - CapabilityReview: Review account's capabilities. We'll use it to identify a user as a system admin.

* authz_handler in auth package
  - implemented Authorizer method: authorizes a request to ensure
    the authenticated user has permissions to invoke to sepcified API.
  - The flow is as follows:
      -- Invoke an access_review request to AMS for verifying the user can create 
          a cluster resource - an indication that the user has permissions to user the installer.
      -- If the user is a system admin, allow all APIs endpoints.
          The verification is done in auth_handler and stored in the payload (using capability_review request to AMS).

* Filter clusters by owner user_name
  - Clusters table has a user_name column to identify the owner of each cluster.
  - Using the username instead of id, as only the username is guaranteed
     to be available in the token (and it's unique as well).
  - The clusters should be filtered according to the authenticated user.
  - A user with admin capability and list and manipulate all clusters.


danielerez:
@ronniel1 @avishayt @romfreiman @rollandf @nmagnezi

PR #532, opened 2020-08-02
MGMT-1670 Remove redundant log messages

ori-amizur:


internal/host/common.go
```diff
@@ -60,7 +60,7 @@ func updateHostStatus(ctx context.Context, log logrus.FieldLogger, db *gorm.DB,
 		extra = append(extra, "status_updated_at", strfmt.DateTime(time.Now()))
 	}
 
-	if host, err = UpdateHost(log, db, clusterId, hostId, srcStatus, extra...); err != nil ||
+	if host, err = UpdateHost(log, db, clusterId, hostId, srcStatus, newStatus, extra...); err != nil ||
```
ronniel1, line 63: I don't think that the way to solve this is to add the newStatus param to UpdateHost()
I think its better to move the  outputting of the message to this function

@YuviGold What do you think?

ori-amizur: OK

PR #533, opened 2020-08-03
Inserting changes from `assisted-service` (integration with app-sre)

ybettan:


openshift/template.yaml
```diff
@@ -0,0 +1,174 @@
...
+apiVersion: v1
+kind: Template
+metadata:
+  name: assisted-installer
+objects:
+- apiVersion: apps/v1
+  kind: Deployment
+  metadata:
+    name: assisted-service
+    namespace: assisted-installer
```
filanov, line 21: i think that this became a paramter.

ybettan: I think that the namespace is overwritten by app-sre, all deployments in `master` branch are deployed to `assisted-installer` ns and all deployments in `integration-test` branch are deployed to `assisted-installer-test` ns.

@apahim can you approve?

apahim: Yes, it has no effect here. You can safely remove the name space from this file.

PR #376, opened 2020-07-08
MGMT-1173 Run assistant using podman

rwsu:
Adds ISOGenerator and InstallConfigGenerator interfaces to allow k8s and local implementations.

Moves k8s job specific code to job.go.

Introduces deploy-onprem, clean-onprem, and test-onprem make targets.

Dockerfile.bm-inventory-disconnected combines Dockerfile.bm-inventory with additional bits needed to bake install config and, in a follow-on patch, ISO generation scripts into the bm-inventory image.

PR #535, opened 2020-08-03
MGMT-1173 Run assistant using podman

ronniel1:
Option to deploy installer services inside a pod on the localhost


Makefile
```diff
@@ -21,6 +22,7 @@ SERVICE := $(or ${SERVICE},quay.io/ocpmetal/bm-inventory:latest)
 GIT_REVISION := $(shell git rev-parse HEAD)
 APPLY_NAMESPACE := $(or ${APPLY_NAMESPACE},True)
 ROUTE53_SECRET := ${ROUTE53_SECRET}
+CONTAINER_COMMAND = $(shell if [ -x "$(shell command -v docker)" ]; then echo "docker" ; else echo "podman"; fi)
```
pawanpinjarkar, line 25: If one has docker as well as podman installed, this resolves in `docker` and then `${CONTAINER_COMMAND} pod create` below at line 140 results in `docker pod` command which is incorrect

Jenkinsfile
```diff
@@ -20,6 +20,22 @@ pipeline {
...
+
+    stage('clear deployment after subsystem test') {
+      steps {
+        sh 'make clear-deployment'
+      }
+    }
+
+    stage('test subsystem with podman') {
+      steps {
+        sh '''export CONTAINER_COMMAND=podman; export SERVICE=quay.io/ocpmetal/bm-inventory:test-onprem; make build-onprem'''
```
pawanpinjarkar, line 32: `build-onprem` target is missing. How about adding targets as below 
```
all: build
all-onprem: build-onprem

lint:
	golangci-lint run -v

.PHONY: build
build: lint unit-test build-minimal
build-onprem: lint unit-test test-onprem build-minimal
```

Makefile
mhrivnak: Maybe we just need to to be clear that onprem only works with podman, and exit with a clear message if `${CONTAINER_COMMAND}` is `docker`?

pawanpinjarkar:
The final changes are here https://github.com/filanov/bm-inventory/pull/376/files

Jenkinsfile
rwsu: build-onprem is intended to build the container image. I've updated the Makefile in https://github.com/filanov/bm-inventory/pull/376 to include the missing target.

Makefile
rwsu: tsorya suggested that we would need to support both docker and podman, but as you said, onprem is only intended to be deployed using podman. I've removed CONTAINER_COMMAND in the original PR: https://github.com/filanov/bm-inventory/pull/376. Maybe @tsorya can further clarify why docker should be supported.

PR #536, opened 2020-08-03
MGMT-1649 Add host installation disk path property

YuviGold:


ori-amizur (approved):

internal/host/transition.go
```diff
@@ -103,7 +104,8 @@ func (th *transitionHandler) PostRegisterDuringReboot(sw stateswitch.StateSwitch
 	}
 
 	return th.updateTransitionHost(params.ctx, logutil.FromContext(params.ctx, th.log), th.db, sHost,
-		"Expected the host to boot from disk, but it booted the installation image - please reboot and fix boot order to boot from disk")
+		fmt.Sprintf("Expected the host to boot from disk, but it booted the installation image - please reboot and fix boot order to boot from disk %s",
+			sHost.host.InstallationDiskPath))
```
avishayt, line 108: Please use InstallationDiskPath to find the disk in host.inventory, and if the disk's serial isn't empty, add it in parentheses at the end of the string

avishayt (changes requested):
Add serial as described in comment

internal/host/installcmd.go
```diff
@@ -89,8 +89,8 @@ func (i *installCmd) GetStep(ctx context.Context, host *models.Host) (*models.St
 	}
 	step.Args = []string{"-c", buf.String()}
 
-	if err := i.db.Model(&models.Host{}).Where("id = ?", host.ID.String()).
-		Update("installer_version", i.instructionConfig.InstallerImage).Error; err != nil {
+	if _, err := UpdateHost(i.log, i.db, host.ClusterID, *host.ID, *host.Status,
```
filanov, line 92: please use `updateHostStatus` instead, looks like `UpdateHost` function is used only in `updateHostStatus` and it doesn't contain log about the changes.

internal/host/installcmd_test.go
```diff
@@ -59,26 +60,37 @@ var _ = Describe("installcmd", func() {
...
-		stepReply, stepErr = installCmd.GetStep(ctx, &host)
-		postvalidation(true, true, stepReply, stepErr, "")
-	})
-
-	It("get_step_one_master_no_disks", func() {
-		var emptydisks []*models.Disk
-		mockValidator.EXPECT().GetHostValidDisks(gomock.Any()).Return(emptydisks, nil).Times(1)
-		stepReply, stepErr = installCmd.GetStep(ctx, &host)
-		postvalidation(true, true, stepReply, stepErr, "")
+	Describe("negative", func() {
```
filanov, line 63: why not context?

filanov (changes requested):

internal/host/installcmd.go
YuviGold: But I don't change the status

internal/host/installcmd_test.go
YuviGold: What do you mean?

PR #537, opened 2020-08-03
MGMT-1388 add validation data to cluster model

gamli75:
@ori-amizur please review and comment


internal/cluster/cluster.go
```diff
@@ -155,28 +139,65 @@ func (m *Manager) DeregisterCluster(ctx context.Context, c *common.Cluster) erro
...
+		state, err := m.getCurrentState(swag.StringValue(cluster.Status))
+		if err != nil {
+			return nil, err
+		}
+
+		clusterAfterRefresh, err := state.RefreshStatus(ctx, &cluster, db)
+
+		//report installation finished metric if needed
+		reportInstallationCompleteStatuses := []string{models.ClusterStatusInstalled, models.ClusterStatusError}
+		if err == nil && stateBeforeRefresh != "" && stateBeforeRefresh == models.ClusterStatusInstalling &&
```
ori-amizur, line 164: Remove StateBeforeRefresh != "".  StateBeforeRefresh == models.ClusterStatusInstalling is enough

internal/cluster/cluster.go
```diff
@@ -155,28 +139,65 @@ func (m *Manager) DeregisterCluster(ctx context.Context, c *common.Cluster) erro
...
-		return nil, errors.Wrapf(err, "failed to get cluster %s", c.ID.String())
+		if err := db.Preload("Hosts").Take(&cluster, "id = ?", c.ID.String()).Error; err != nil {
+			return nil, errors.Wrapf(err, "failed to get cluster %s", c.ID.String())
+		}
+		state, err := m.getCurrentState(swag.StringValue(cluster.Status))
+		if err != nil {
+			return nil, err
+		}
+
+		clusterAfterRefresh, err := state.RefreshStatus(ctx, &cluster, db)
```
ori-amizur, line 160: Why state.RefreshStatus? Thought it should be removed and replaced by stateswitch

internal/cluster/cluster.go
```diff
@@ -155,28 +139,65 @@ func (m *Manager) DeregisterCluster(ctx context.Context, c *common.Cluster) erro
...
+
+		//report installation finished metric if needed
+		reportInstallationCompleteStatuses := []string{models.ClusterStatusInstalled, models.ClusterStatusError}
+		if err == nil && stateBeforeRefresh != "" && stateBeforeRefresh == models.ClusterStatusInstalling &&
+			funk.ContainsString(reportInstallationCompleteStatuses, swag.StringValue(clusterAfterRefresh.Status)) {
+			m.metricAPI.ClusterInstallationFinished(log, swag.StringValue(cluster.Status), c.OpenshiftVersion, c.InstallStartedAt)
+		}
+		return clusterAfterRefresh, err
```
ori-amizur, line 168: I think this part should be part of post-transition.  Here should be the minimal code to call the state-machine

internal/cluster/installing.go
```diff
@@ -53,11 +53,11 @@ func (i *installingState) RefreshStatus(ctx context.Context, c *common.Cluster,
 func (i *installingState) getClusterInstallationState(ctx context.Context, c *common.Cluster, db *gorm.DB) (string, string, error) {
 	log := logutil.FromContext(ctx, i.log)
 
-	mappedMastersByRole := mapMasterHostsByStatus(c)
+	mappedMastersByRole := MapMasterHostsByStatus(c)
```
ori-amizur, line 56: Shouldn't the RefreshStatus above be removed? It should be handled by stateswitch

internal/cluster/statemachine.go
```diff
@@ -80,5 +81,74 @@ func NewClusterStateMachine(th *transitionHandler) stateswitch.StateMachine {
...
+			stateswitch.State(models.ClusterStatusReady),
+			stateswitch.State(models.ClusterStatusInsufficient),
+		},
+		Condition:        stateswitch.Not(requiredInputFieldsExist),
+		DestinationState: stateswitch.State(models.ClusterStatusPendingForInput),
+		PostTransition:   th.PostRefreshCluster(statusInfoPendingForInput),
+	})
+
+	// In order for this transition to be fired at least one of the validations in sufficientForInstallValidations must fail.
+	// This transition handles the case that one of the required validations that are required in order for the host
```
ori-amizur, line 104: There is overlap between these two transitions.  The overlap is 
stateswitch.And(stateswitch.Not(requiredInputFieldsExist), stateswitch.Not(isSufficientForInstall)).
It means that first rule that matches the conditions will be fired.  If the order between will change, then you will get a different behavior.
Since I assume that the first transition has priority over the second one with regard to the overlap, I suggest that the condition for the second one will be
stateswitch.And(requiredInputFieldsExist, stateswitch.Not(isSufficientForInstall))

internal/cluster/transition.go
```diff
@@ -153,3 +167,124 @@ func (th *transitionHandler) updateTransitionCluster(log logrus.FieldLogger, db
...
+					updatedCluster.OpenshiftVersion, updatedCluster.InstallStartedAt)
+			}
+			return nil
+		}
+		return err
+	}
+	return ret
+}
+
+func setPendingUserResetIfNeeded(ctx context.Context, log logrus.FieldLogger, db *gorm.DB, hostApi host.API, c *common.Cluster) {
```
ori-amizur, line 246: Seems that this whole section regarding the "pending-user-reset" should be part of host refresh status, but not as host validations.  This operation accesses hosts, but does not rely on any information from the cluster.

internal/cluster/validation_id.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/filanov/bm-inventory/models"
+)
+
+type validationID models.ClusterValidationID
+
+const (
+	IsMachineCidrDefined                       = validationID(models.ClusterValidationIDMachineCidrDefined)
+	isMachineCidrEqualsToCalculatedCidr        = validationID(models.ClusterValidationIDMachineCidrEqualsToCalculatedCidr)
+	isApiVipBelongToMachineCidrAndNotInUse     = validationID(models.ClusterValidationIDAPIVipBelongsToMachineCidrAndNotInUse)
```
ori-amizur, line 17: I think that ClusterValidationIDAPIVipValid or ClusterValidationIDAPIVipReady is better

internal/cluster/validator.go
```diff
@@ -0,0 +1,217 @@
...
+		return ValidationFailure
+	}
+}
+
+type clusterValidator struct {
+	log logrus.FieldLogger
+}
+
+func (v *clusterValidator) isMachineCidrDefined(c *clusterValidationContext) validationStatus {
+	if c.cluster.MachineNetworkCidr == "" {
```
ori-amizur, line 90: If machine network cidr is not defined, this validation should fail.  A validation can be pending if it depends on other validation to be successful

internal/cluster/validator.go
```diff
@@ -0,0 +1,217 @@
...
+		return "Machine network CIDR is defined"
+	case ValidationFailure:
+		return fmt.Sprintf("Machine network CIDR has invalid value %s", context.cluster.MachineNetworkCidr)
+	default:
+		return fmt.Sprintf("Unexpected status %s", status)
+	}
+}
+
+func (v *clusterValidator) isMachineCidrEqualsToCalculatedCidr(c *clusterValidationContext) validationStatus {
+	cidr, err := network.CalculateMachineNetworkCIDR(c.cluster.APIVip, c.cluster.IngressVip, c.cluster.Hosts)
```
ori-amizur, line 111: Here this validation should be pending because machine network cidr is not yet defined or none of the vips are defined.

internal/cluster/validator.go
```diff
@@ -0,0 +1,217 @@
...
+	case ValidationSuccess:
+		return "Cluster machine CIDR equals to the calculated CIDR "
+	case ValidationFailure:
+		return fmt.Sprintf("Cluster machine CIDR %s is different than the calculated CIDR %s", context.cluster.MachineNetworkCidr, context.calculateCidr)
+	default:
+		return fmt.Sprintf("Unexpected status %s", status)
+	}
+}
+
+func (v *clusterValidator) isApiVipBelongToMachineCidrAndNotInUse(c *clusterValidationContext) validationStatus {
```
ori-amizur, line 127: Should be pending if vip is not yet defined

internal/cluster/validator.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Sprintf("Cluster has %d known master hosts", MinHostsNeededForInstallation)
+	case ValidationFailure:
+		return fmt.Sprintf("cluster is expected to have exactly %d known master to be installed", MinHostsNeededForInstallation)
+	default:
+		return fmt.Sprintf("Unexpected status %s", status)
+	}
+}
+
+func (v *clusterValidator) noHostInPendingForInput(c *clusterValidationContext) validationStatus {
+	mappedHostsByStatus := MapHostsByStatus(c.cluster)
```
ori-amizur, line 180: Why is this validation needed?

swagger.yaml
```diff
@@ -1804,3 +1809,14 @@ definitions:
...
       - 'belongs-to-machine-cidr'
+
+  cluster-validation-id:
+    type: string
+    enum:
+      - 'machine-cidr-defined'
+      - 'machine-cidr-equals-to-calculated-cidr'
+      - 'api-vip-belongs-to-machine-cidr-and-not-in-use'
+      - 'ingress-vip-belongs-to-machine-cidr-and-not-in-use'
+      - 'no-pending-for-input-host'
```
ori-amizur, line 1820: Not needed. all-hosts-are-ready-to-install should be enough

swagger.yaml
```diff
@@ -1804,3 +1809,14 @@ definitions:
       - 'hostname-unique'
       - 'hostname-valid'
       - 'belongs-to-machine-cidr'
+
+  cluster-validation-id:
+    type: string
+    enum:
+      - 'machine-cidr-defined'
+      - 'machine-cidr-equals-to-calculated-cidr'
+      - 'api-vip-belongs-to-machine-cidr-and-not-in-use'
```
ori-amizur, line 1818: Maybe api-vip-valid is better?  This will enable us to change the vip validation logic without have to change the validation id

ori-amizur (commented):
I think we should consider changing the bmInventory.InstallCluster not to call VerifyVips and verify that all hosts belong to machine network cidr.  Instead, we can call RefreshStatus that does all that and verify that the cluster is ready.
Same was done for hosts.

PR #539, opened 2020-08-04
OCPBUGSM-13662 Allow deploying postgres secret to costum namespace

eranco74:


filanov:
@eranco74 need to fix s3 as well

